PR #33, opened 2018-02-13
Submodules

loganasherjones:
Removing the actual submodules as they have broken out by repos.


hazmat345:
This is super cool. Only question - until your pyrabbit change gets released we will break on rabbit 3.7. Should we be indicating that in the readme?

PR #49, opened 2018-03-23
Yapconf

hazmat345:
I'd like to use this change to streamline a Beergarden introduction tutorial.

I'd like to split out the example plugins into their own repository and reference them as a submodule. That way they can live in both bartender/plugins and beer-garden/example-plugins. I think it's important to have them easily visible for a tutorial as well as available to bartender.


loganasherjones:
LGTM. Just need to release the yapconf stuff so we don't install from that zip file.

Issue #69, opened 2018-05-02
Grafana Dashboards


loganasherjones:
This goes along with #68 

Once we have prometheus metrics, we should be providing some grafana dashboards to visualize the status of beer-garden.

loganasherjones:
Okay, so after working on this, I think I need to tweak my Prometheus metrics. I do not think we need histograms in as many places as I thought, I'm pretty sure just measuring counts and latency for the time being should be enough. We can do this with gauges and counters, and will not need histograms. I do have some pretty sweet graphs in grafana, so I just need to do a couple more hours of tweaking to the prometheus stuff then revamp the metrics I have, and we should be good to go on this. I'll attach a couple examples of the dashboards I have so far

loganasherjones:
* [Beer Garden Performance - 2](https://ibb.co/nqKiNy)
* [Beer Garden Performance - 1](https://ibb.co/kyVwaJ)
* [Plugin Performance](https://ibb.co/cMedpd)

PR #96, opened 2018-06-23
Adding Grafana Dashboards

loganasherjones:
This is for  beer-garden/beer-garden#69

You should probably look at the following before merging:

* https://github.com/beer-garden/brew-view/pull/66
* https://github.com/beer-garden/bartender/pull/29

Changes to the docker-compose to include grafana and prometheus
by default. This enables our metrics to be scraped and the
default installation of beer-garden to be monitorable.

I added the two dashboards I created. You might notice they are
in two spots. There is a reason for this. The dashboards located
at `resources/grafana_dashboards` are the dashboards that get
uploaded to grafana.com.  These can be imported and when they
get imported by people, they get a prompt asking them to point
the dashboards at a prometheus datastore. Unfortunatley, you
cannot do the same for the "provisioned" dashboards that are
in the docker-compose section. In this section, I had to hard
code the datasource, which is okay because I'm also provisioning
the datasource, but still. That's why it is there.

### DIFF
--- a/docker/docker-compose/docker-compose.yml
+++ b/docker/docker-compose/docker-compose.yml
@@ -1,6 +1,9 @@
 version: '3.5'
 
 x-definitions: &env
+    BG_AMQ_ADMIN_HOST: rabbitmq
+    BG_AMQ_ADMIN_USER: beer_garden
+    BG_AMQ_ADMIN_PASSWORD: password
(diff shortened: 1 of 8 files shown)


loganasherjones:
Also I added some stuff to grafana.com

* https://grafana.com/dashboards/6621
* https://grafana.com/dashboards/6624

Issue #205, opened 2018-12-12
UI Update


loganasherjones:
This is a complete overhaul of the web front-end. We are moving from Angular 1.X to React, and Bootstrap to Material UI The main reasons we are moving are:

* Angular 1.X is extremely old and lots of improvements have been made in the front-end framework game
* React is very popular, and I hope it increases people that can contribute to the project
* Testing the front-end is not good right now, we need a better way to test it all
* Angular-schema-form is not well maintained, but the react equivalent looks more maintained.
* To update the look-and-feel of the front end

In addition to this, the front-end update will be the first part of the transition into the `beer-garden` repo. In theory, this can be done without a `3.0.0` migration, and if I'm done fast enough, we may do just that. 

Here are the things that will need to be converted. I'll be checking them off as I do them, and pushing to the `3.0.0-ui` branch as I get them done. Feel free to add to this if you think I'm missing something

[ ] - foo

hazmat345:
Should address #3 as well

hazmat345:
I'm going to move this to react-ui. If it's not useful then we can close.

PR #206, opened 2018-12-12
WIP UI Update

loganasherjones:
See #205 for all the features here. I intend on creating PRs for all the small features into the `3.0.0-ui` branch as I go.


loganasherjones:
Authentication is taking much longer than I expected, however I have a good chunk of it done. Left to do before this is really ready to merge is authenticated routes. This should handle what happens if a user goes to a route they can't see, and should encapsulate the authentication/authorization routing into a single component.

Everything else is mostly done, however there are some things I discovered about how we are doing authentication that I believe will need to change, so there are auth_update branches on several of the other repos.

loganasherjones:
Ok @hazmat345 this is ready for a review. The changes are summarized best like this:

1. Made a basic app layout, dashboard, topbar, sidebar etc
2. Made some container components where we will do the actual fetching (systems, commands, requests, and advanced)
3. Themes
4. Authentication, it works with and without authentication enabled. 

I don't know the best way to review this, so I'm open to ideas. If you want to test this out, the best way I can think of doing so is:

* Clone all the `auth_update` branches and this branch `3.0.0-ui`
* Start bartender/brew-view
* `npm install`
* `npm start`

PR #219, opened 2019-01-28
Issue/215

### LINKED ISSUE #17: Create systemd service for beer-garden
Ideally, we should be creating a systemd script for BEERGARDEN for any RHEL-7 versions of the software. Right now, we just create a single service script that will work with both rhel6 and 7 versions. This is fine, but if we ever get time to create a systemd script as well as a service script, we should do that.

### LINKED ISSUE #215: Rpm install not clearing old package metadata
When upgrading via the rpm it looks like the beer-garden packages aren't getting fully removed.

When upgrading from an older version I can see some extra directories in the `lib/python3.6/site-packages` directory:

- brewtils-2.4.1-py3.6.egg-info
- brewtils-2.4.4.dist-info
- brew_view-2.4.4-py3.6.egg-info
- brew_view-2.4.7.dist-info

It's also happening with non-beergarden packages:
- six-1.11.0.dist-info
- six-1.12.0.dist-info

The old directories are all empty, so this may not actually be a problem.

hazmat345:
This fixes #215 and fixes #17.

It depends on beer-garden/bg-utils#40


hazmat345:
Last thing is to ensure the services stop during upgrade / uninstall.

resources/centos7/after_install.sh
```diff
@@ -34,253 +37,51 @@ case "$1" in
             mkdir -p "$PLUGIN_HOME"
         fi
 
-        if [ ! -d "$PID_HOME" ]; then
-          mkdir -p "$PID_HOME"
```
loganasherjones: This makes me nervous. This came about as the result of a bug where we were not managing PIDs correctly. This might be gone moving to the services, but I feel its necessary to confirm that:

1. Start any beer-garden service
2. Stop the service
3. Ensure that the python process is no longer running

The reason we needed to do this in the first place was because if this directory was not created, then the pid file would not get created and our service scripts could not actually stop the services. So as long as we can do that, I'm good.

hazmat345:
So this does change how things work. Previously we had the one service, `beer-garden`, that was responsible for managing both top-level processes. We could continue to do that, but it doesn't fit into systemd very well. We'd have to continue to do what we did with the sysV script - manually keep track of the processes (and the pids) ourselves.

This change splits the services into one per process. That allows us to let systemd manage the pids completely.

So there are now three services: `beer-garden`, `brew-view`, and `bartender`. Both `brew-view` and `bartender` can be started or stopped independently, and starting or stopping `beer-garden` will start or stop the other two.

hazmat345:
Just pushed up some changes that should also address the issue on Centos 6, but I think we should drop support soon.

hazmat345:
Verified that this works on Centos 6.

loganasherjones:
I wonder how we should test this stuff. It feels like maybe we should use vagrant? Ah, well, maybe another day.

hazmat345:
@loganasherjones just added an update to the rpm scripts to convert .yml configs to .yaml since we're not doing that in the bg-utils utility.

PR #240, opened 2019-03-02
User/Role Management

loganasherjones:
This branch is mostly done with development. Feel free to have a look.


hazmat345:
@west270 @scott-taubman

This branch was @loganasherjones React UI. I don't think there's any expectation that this will be merged, but it may be worth taking a look to see if you like the look of anything.

Afterwards I recommend this PR be closed and the branch deleted.

PR #297, opened 2019-06-04
Changing compose publish host location for better visibility

hazmat345:
Small change to the compose file to make it more obvious what needs to change in order to support remote plugins.

### DIFF
--- a/docker/docker-compose/docker-compose.yml
+++ b/docker/docker-compose/docker-compose.yml
@@ -1,9 +1,10 @@
 version: '3.5'
 
 x-definitions: &env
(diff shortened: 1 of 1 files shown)


hazmat345:
Unfortunately, whatever the default setting is here can either work for local or remote plugins. A setting that works for both will require a change (to the 'real' ip address, for example).

So I'm changing this to the value that will support the simplest possible tutorial. That way the user doesn't need to worry about it at all for the initial run through. Then we can have the user change it as part of the discussion on the differences between local and remote plugins.

loganasherjones:
I like this change!

PR #313, opened 2019-09-03
Config Subsystem

loganasherjones:
# Overview

The old config modules are gone at their respective top levels and
should more-or-less now be replaced with the new module. This
transition is far from done, as all the migration/generation is
not complete, but this is a good start on the work.

Somethings worth noting here:

1. I removed the `thrift` and `backend` parts of the config as I think they should no longer be used. 
2. I really am having a hard time justifying the existence of a `namespaces` portion of the config. I don't think this needs to exist at all. Happy to open a separate issue for that if you'd like.

## Why am I doing this?

I was looking at the changes of v3 this weekend, and just wanted to figure out how to get up and running. I started writing some instructions for how to do this, then realized that in order to start everything, we'd need two configs. I thought that was silly to write as it was going to change, so I sat out trying to actually make the change.

I think I have it to the point where the main application should run mostly like normal, but I haven't figure out config generation or anything like that yet. Hence the WIP.

## Notes for Matt

While I was doing this, I realized that we use configs in a lot of places. This change would have been made significantly easier if we had some rules about where configuration values were to be used. I think we should add this to the "golden rules" section of the v3 architecture documentation:

> Wherever possible, configuration values are used as initialization parameters to classes to make them easier to test. When a class is not appropriate, then module level defaults should be setup during module initialization via a `setup()` call.

Thoughts?


loganasherjones:
@hazmat345 Hey, I know this is a WIP, but before I went any further, I wanted you to take a look at what I have and let me know what you think.

src/app/beer_garden/bg_utils/mongo/__init__.py
```diff
@@ -17,15 +17,17 @@ def setup_database(config):
...
 
+    db_config = config.get("db")
+
     try:
         # Set timeouts here to a low value - we don't want to wait 30
         # seconds if there's no database
         conn = connect(
             alias="aliveness",
-            db=config.db.name,
+            db=db_config["name"],
```
hazmat345: Is there a specific reason for using dictionary-style lookup here?

hazmat345:
That's great you started this. Getting the config consolidated was actually the next 'real' thing on my list.

Totally fine with getting rid of the thrift and backend parts.

Also OK with getting rid of the "namespaces" section, kind of. The current form is definitely a stopgap to let me get something working. Currently you need to specify downstream connection information, which is not what we want. However, we'll need to migrate to some way to point at an upsteam (so maybe an `upstream` section would be appropriate).

I'm not opposed to the new golden rule regarding configuration usage. My only thought there is for eventual dynamic configuration reloading... if the config values are only set during initialization we should try to support an easy way to have all the subsystems reload their configuration at once.

hazmat345:
Before you work on this more check out #314

src/app/beer_garden/bg_utils/mongo/__init__.py
loganasherjones: nope

loganasherjones:
I'm not exactly sure how we are going to mark things and upstream/downstream. But yeah, we'll need some way to know about various beer-gardens. Even that probably shouldn't be in the configuration though. Ideally we should have some API to let downstream BG's register with an upstream.

Yes, for dynamic configuration reloading, we could use the watching capability in `yapconf`. But I agree, that is a thing that will be slightly more painful. One of the main reasons I want to move the config out to the edges is so that testing becomes much easier. I would be ok with a different way as long as we have a very simple, and straighforward way to write unit tests.

hazmat345:
A thing to note - when generating a logging config file it will currently **only** use cli args. I think this is just a relic from before we were using yapconf at all, but I wanted to verify.

To fix, in `beer_garden.config._parse_args` change this:

```
spec = YapconfSpec(_SPECIFICATION)
```

to this:

```
spec = YapconfSpec(_SPECIFICATION, env_prefix="BG_")
```

src/app/beer_garden/config.py
```diff
@@ -32,12 +32,16 @@ def load(args: List[str], force=False):
...
     Once loaded, the configuration can be reached with the `get` method.
 
-    :param args: Command line arguments.
-    :param force: Force a reload.
+    Args:
+        args: Command line arguments
+        force: Force a reload
+
+    Returns:
+        The configuration object
```
loganasherjones: Conceptually, I don't think we should be returning the configuration object here. The intent is always to access the config object through `config.get`. This opens us up to someone modifying the configuration object outside of the scope of the config subsystem.

src/app/beer_garden/config.py
```diff
@@ -66,21 +74,20 @@ def generate(args: List[str]):
...
+        vars(args), "ENVIRONMENT", exclude_bootstrap=True
+    )
 
     dump_data(
-        filtered_config,
-        filename=config.configuration.file,
-        file_type=_get_config_type(config),
+        config,
+        filename=bootstrap.configuration.file,
+        file_type=_get_config_type(bootstrap),
```
loganasherjones: do we still want to support `json` and `yaml`? I'm fine saying, our config files are yaml.

src/app/beer_garden/log.py
```diff
@@ -9,10 +10,24 @@
 _LOGGING_CONFIG = None
 
 
-def setup_application_logging(config):
+def load(config: Box, force=False) -> dict:
```
loganasherjones: I'd say let's make this type hint a `dict` rather than a `Box`.

src/app/beer_garden/config.py
hazmat345: :+1:

src/app/beer_garden/log.py
hazmat345: :+1:

src/app/beer_garden/config.py
hazmat345: I'm good with that. Even easier :smile:

hazmat345:
@loganasherjones I'm going to apply the same thing to the logging `load` method (don't return it, just set the global).

hazmat345:
@loganasherjones Ok, I made the yaml-only changes. Now all generated and updated config will be yaml.

We still load json files correctly since the yaml parser handles them, which is nice.

hazmat345:
Going to remove the WIP tag. Tests in `config_test.py` and `log_test.py` all pass.

src/app/beer_garden/config.py
```diff
@@ -0,0 +1,1042 @@
...
+from yapconf import YapconfSpec, dump_data
+
+from beer_garden.log import default_app_config
+
+__all__ = ["load", "generate_logging", "generate", "migrate", "get"]
+
+_CONFIG = None
+
+
+def load(args: Sequence[str], force=False) -> None:
```
loganasherjones: Type Annotation suggestion/question:

1. If we type annotate one thing, we should type annotate all things.
2. Should we annotate a return type of `None`?  We do here, but not for the rest of the functions.

src/app/beer_garden/config.py
```diff
@@ -0,0 +1,1042 @@
...
+    # Determine if a type conversion is needed
+    type_conversion = False
+    new_type = "yaml"
+    if current_type != new_type:
+        new_file = current_root + "." + new_type
+        type_conversion = True
+    else:
+        new_file = config.configuration.file
+
+    # logger.debug("About to migrate config at %s" % config.configuration.file)
```
loganasherjones: Remove this.

src/app/beer_garden/config.py
```diff
@@ -0,0 +1,1042 @@
...
+        config.configuration.file,
+        current_file_type=current_type,
+        output_file_name=new_file,
+        output_file_type=new_type,
+        update_defaults=True,
+        include_bootstrap=False,
+    )
+
+    if type_conversion:
+        # logger.debug("Removing old config file at %s" % config.configuration.file)
```
loganasherjones: remove this.

src/app/beer_garden/config.py
```diff
@@ -0,0 +1,1042 @@
...
+    spec, cli_vars = _parse_args(args)
+    filtered_args = spec.load_filtered_config(
+        cli_vars, include=["log.config_file", "log.file", "log.level"]
+    )
+
+    log = filtered_args.get("log", {})
+    logging_config = default_app_config(log.get("level"), log.get("file"))
+    log_config_file = log.get("config_file")
+
+    if log_config_file is not None:
```
loganasherjones: This could be replaced with a call to `dump_data`

src/app/beer_garden/config.py
```diff
@@ -0,0 +1,1042 @@
...
+        with open(log_config_file, "w") as f:
+            dumped = json.dumps(logging_config, indent=4, sort_keys=True)
+            f.write(six.u(dumped))
+    else:
+        print(json.dumps(logging_config, indent=4, sort_keys=True))
+
+    return logging_config
+
+
+def get(key: str):
```
loganasherjones: One thing I was thinking about this function call. We are still opening ourselves up to allowing people to change the config. Do you think we should be returning an immutable copy of the config?

loganasherjones: For example:

```
config.load()
log_config = config.get("log")
log_config["level"] = "this probably should not be allowed"
```

src/app/beer_garden/brew_view/__init__.py
```diff
@@ -35,33 +24,38 @@
...
 
-    app_logging_config = beer_garden.bg_utils.setup_application_logging(
-        config, get_default_logging_config(config.log.level, config.log.file)
-    )
+    beer_garden.log.load(beer_garden.config.get("log"))
     logger = logging.getLogger(__name__)
-    logger.debug("Logging configured. First post!")
 
     _setup_application()
+    logger.debug("Successfully loaded the brew-view application")
```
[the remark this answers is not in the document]
hazmat345: I'll just take the "brew-view" out... the logging relating to startup is going to be different soon.

src/app/beer_garden/config.py
hazmat345: Oh yeah, you mean for `force`? I agree, we should do all the args.

I don't know about the `None` return type. I don't really want to go though and add type hints to everything, so I figured that we could just add them as we touch things. My thinking was that explicit was better than implicit, but I don't know how useful this particular thing is.

src/app/beer_garden/config.py
hazmat345: :+1:

src/app/beer_garden/config.py
hazmat345: :+1:

src/app/beer_garden/config.py
hazmat345: Yeah, I'm good with that.

hazmat345: Do you think this should be done at load time? We can do something like this:

```python
_CONFIG = Box(spec.load_config(*config_sources).to_dict(), frozen_box=True)
```

src/app/beer_garden/config.py
hazmat345: :+1:

Could you add `dump_data` to `__all__`? :smile:

src/app/beer_garden/config.py
loganasherjones: I agree, I'm not sure how particularly useful it is to document a return type of `None` but, let's say, that if we decide to type a method signature that will mean the arguments as well as the return type, as a rule.

src/app/beer_garden/config.py
loganasherjones: Yep! I think that's great!

PR #319, opened 2019-09-16
Api

hazmat345:
This should only be merged after #313.


hazmat345:
So at this point it's possible to run Beergarden in HTTP or Thrift mode. Just pass either `--entry-http-enable` or `--entry-thrift-enable` when starting.

hazmat345:
This PR got out of control. Going to rebase.

PR #326, opened 2019-09-23
Tests

hazmat345:
This fixes some tests (and marks some as skip).


hazmat345:
@loganasherjones I'm going to go ahead and just merge this in. The only substantive change is that now the local plugin loader has config options passed to it during init instead of reaching into the config itself.

PR #330, opened 2019-09-25
Mongo Pruner Setup

hazmat345:
This PR moves the logic for setting up the mongo pruner into the pruner module. It also fixes / moves / skips some tests.


hazmat345:
I'm going to merge this in as it's pure refactoring / test changes.

PR #332, opened 2019-09-30
Removing static file handlers from http url map

hazmat345:
This PR removes the static file handlers from the Tornado configuration.


loganasherjones:
Just want to make sure we're clear. The end game is to have nginx or some other frontend actually server up our static content correct?  So this change forces developers to use the webpack development server correct?

hazmat345:
Yep. The `ui` directory is the place for the frontend now. Developers can run `npm run serve` for the webpack dev server, or they can use the Dockerfile there to build the nginx image.

loganasherjones:
This seems like a good a time as any to start documenting the developer guide. We should tell developers how we expect them to do the development. I'm up for creating a `DEVDOC.md` or just adding information to the `README.md` about how exactly to start this thing.

PR #339, opened 2019-10-16
Url prefix

hazmat345:
This PR removes an issue I've been bitten by a couple of times. It normalizes the url_prefix config values when they are loaded, which removes the need to call `normalize_url_prefix` everywhere the value is used.


hazmat345:
Oh is this already in `yapconf`? I should have done my homework :cry:

loganasherjones:
No sorry, I meant I should add that feature to yapconf

hazmat345:
Ahh ok. Well if you add that I'll be happy to rip this out :smile:

PR #340, opened 2019-10-17
DB Subsystem

hazmat345:
This PR creates a logical Mongo subsystem. This needs to be merged after beer-garden/brewtils#152.

This is ready to be looked at, but is not 100% complete. Things that I want to do before this PR should be accepted:
- ~Move the query logic for `Requests`~ Done

Things that need to be done eventually but should probably be part of their own PR:
- Event publishing still uses the mongo models. I figured this can change as part of the event subsystem rework.
- Things in the `api` package (thrift, http handlers) still use some mongo-specific things. It's mostly user and auth related, which again, will be getting an overhaul.

Things that can be improved:
- I punted on defining an overall "query language". Right now `filter_params` is a dictionary that gets passed directly to the Mongo QuerySet. Obviously that will need to change before we can actually support an alternate database, but I'd like to wait until actually attempting to implement that before commiting to anything. Basically, I'd like to avoid designing a "general" language when I only have one "specific" example.


src/app/beer_garden/requests.py
```diff
@@ -729,20 +734,18 @@ def process_request(
...
 
         try:
-            completed = request_map[request_id].wait(timeout=wait_timeout)
+            completed = request_map[request.id].wait(timeout=wait_timeout)
             if not completed:
-                raise TimeoutError(f"Timeout exceeded for request {request_id}")
+                raise TimeoutError(f"Timeout exceeded for request {request.id}")
 
-            request.reload()
+            request = reload(request)
```
loganasherjones: another good reason to namespace it as `db.reload(`

src/app/beer_garden/systems.py
```diff
@@ -32,19 +27,17 @@ def get_system(system_id: str) -> brewtils.models.System:
...
-    order_by: str = "name",
-    include_fields: Sequence[str] = None,
-    exclude_fields: Sequence[str] = None,
-    dereference_nested: bool = True,
-) -> List[brewtils.models.System]:
+    # filter_params: Dict[str, str] = None,
+    # order_by: str = "name",
+    # include_fields: Sequence[str] = None,
+    # exclude_fields: Sequence[str] = None,
+    # dereference_nested: bool = True,
```
loganasherjones: Delete these and change `Args` below to `Keyword Args:`

src/app/beer_garden/systems.py
```diff
@@ -62,28 +55,29 @@ def get_systems(
...
+    # if not dereference_nested:
+    #     query_set = query_set.no_dereference()
+    #
+    # filtered = query_set.filter(**(filter_params or {}))
+    #
+    # return SchemaParser.parse_system(
+    #     MongoParser.serialize_system(filtered, to_string=False, many=True),
+    #     from_string=False,
+    #     many=True,
+    # )
```
loganasherjones: delete this.

src/app/beer_garden/systems.py
```diff
@@ -150,7 +136,8 @@ def update_system(
                         f"different commands"
                     )
 
-                system.upsert_commands(new_commands)
+                # TODO - Move this somewhere
```
loganasherjones: TODO: this.

src/app/beer_garden/db/api.py
```diff
@@ -0,0 +1,9 @@
+# -*- coding: utf-8 -*-
+import beer_garden.db.mongo.api
+
+query = beer_garden.db.mongo.api.query
+query_unique = beer_garden.db.mongo.api.query_unique
+create = beer_garden.db.mongo.api.create
+update = beer_garden.db.mongo.api.update
+delete = beer_garden.db.mongo.api.delete
+reload = beer_garden.db.mongo.api.reload
```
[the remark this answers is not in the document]
hazmat345: It's only a builtin for 2.7, do we need to worry about it?

src/app/beer_garden/systems.py
hazmat345: Sigh. I spent 20 minutes trying to figure out why my local plugins didn't have commands and this was the reason. :man_facepalming:

src/app/beer_garden/local_plugins/manager.py
```diff
@@ -61,9 +66,11 @@ def start_plugin(self, plugin):
             self.registry.remove(plugin.unique_name)
             self.registry.register_plugin(new_plugin)
         else:
-            raise PluginStartupError("Plugin in an invalid state (%s)" % plugin.status)
+            raise PluginStartupError("Plugin in an invalid state (%s)" % plugin_status)
+
+        plugin_instance.status = "STARTING"
+        update(plugin_instance)
```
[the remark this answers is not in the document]
hazmat345: So to be clear you mean instead of doing this:

```python
from beer_garden.db.api import update

update(plugin_instance)
```

We should do this:
```python
import beer_garden.db.api as db

db.update(plugin_instance)
```

hazmat345:
@loganasherjones how opposed would you be to making instances and commands pure children of the `System` model? They really shouldn't have an independent existence and that change will make managing them much easier.

The only thing giving me pause is the transition - we'd have to update the mongo models, which is always a pain.

src/app/beer_garden/local_plugins/manager.py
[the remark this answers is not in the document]
hazmat345: This is done in 453a765

src/app/beer_garden/requests.py
hazmat345: See 453a765

hazmat345:
OK, this should be mostly done. I went ahead and added a `replace_commands` method to the API. It's not super clean, but I think the "correct" way to fix that is to remove Commands as a top-level collection. I'll open a PR so that can be addressed separately.

src/app/beer_garden/systems.py
hazmat345: Fixed by 407cdb4

src/app/beer_garden/local_plugins/loader.py
```diff
@@ -115,35 +117,37 @@ def load_plugin(self, plugin_path):
...
 
-            # Remove the current instances so they aren't left dangling
-            # TODO: Find a way to carry over instances that aren't changing
-            plugin_system.delete_instances()
+            for instance in plugin_instances:
+                existing_instance = plugin_system.get_instance(instance.name)
+                if existing_instance:
+                    instance.id = existing_instance.id
+
+            # TODO - Remove dangling instances
```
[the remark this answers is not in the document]
hazmat345: Fixed in ef31cb3

hazmat345:
I'm going to go ahead and merge this. Things still outstanding:

- `reload` method name. This is only a builtin in Python 2.7 so I didn't worry about renaming it, especially since we always refer to it as `db.reload` now. 
- The database api has a `replace_commands` method. I'm not very happy about that, but I think the correct way to deal with it is to fix our models. Whether we should do that is still up for debate since it has some consequences. Once that's resolved we can revisit this issue.

PR #344, opened 2019-10-28
Child commands

hazmat345:
This PR removes `Command` as a top-level database model. It's now an EmbeddedDocument owned by the `System` model.

Depends on https://github.com/beer-garden/brewtils/pull/232


hazmat345:
@TheBurchLog this is ready for review

PR #347, opened 2019-10-31
Queues

hazmat345:
This PR is the first pass at the queue subsystem. The API doesn't match what's in the v3 API doc yet. The difference is mostly that the functions operate on `queue_name` parameters instead of instances.

The initial `pika` and `pyrabbit` connection checks have been commented out because the connection definitions are no longer defined at that point. The logic for startup connection checks has already changed in #334, so there's no real benefit in recreating those changes in this PR.


hazmat345:
Going to go ahead and merge this in.

Issue #352, opened 2019-11-05
Rabbit message persistence


hazmat345:
It looks like we may not be persisting rabbit messages correctly across a broker restart. Need to verify that we are.

```python
delivery_mode=pika.spec.PERSISTENT_DELIVERY_MODE
```

Also, we may need to specify a hostname in `docker-compose.yml` for the rabbit container.

hazmat345:
Verified this was an issue, adding delivery mode flag fixed. Verified to work correctly with pika 0.13.1, 1.0.1, 1.1.0.

PR #357, opened 2019-11-05
Setting the hostname of the rabbitmq container

hazmat345:
This is part of the fix for #352. Rabbitmq uses the hostname when constructing its "node name" so the random hostname that docker assigns by default could may cause rabbit to get confused.

### DIFF
--- a/docker/docker-compose/docker-compose.yml
+++ b/docker/docker-compose/docker-compose.yml
@@ -84,6 +84,7 @@ services:
 
     rabbitmq:
         image: rabbitmq:management-alpine
+        hostname: rabbitmq
         networks:
             - bg-network
         environment:


TheBurchLog:
I do not know why this abstract ID that Docker provides would cause issues with Rabbit. It is using the host-name for routing and the provided host-name is valid. Please let me know why this is not a cosmetic change for readability.

Adding the host-name to the docker-compose file should not cause any issues for anyone deploying a single instance of RabbitMQ. 

If someone was to deploy a second instance of RabbitMQ to the same host it will cause a host-name conflict, but they would have to modify the YML regardless for the port changes, so I don't foresee this being an issue.

hazmat345:
That's a good point - this change messes with rabbit's scheme for assigning node names for potentially no reason. Let me check to see if this is really necessary or if beer-garden/bartender#74 fixed the issue already.

hazmat345:
So without this change messages can survive a container restart (`docker-compose restart rabbitmq`). However, taking down the container completely (`docker-compose down`) and bringing up a new one results in a completely blank broker. This change fixes that.

Issue #253, opened 2019-04-07
Event Listener


hazmat345:
We need a capability that will listen to beergarden events and update the database accordingly.

TheBurchLog:
Expanding this ticket to build a generic Event Manager/Listener. Where we can have multiple event types that we can trigger different actions off of.

Example:
When system starts up, mirror event to Parent to infom that downstream systems are available.

TheBurchLog:
@hazmat345 I am pretty sure this is done and checked into V3. Can we close this ticket?

PR #365, opened 2019-11-19
WIP: #253

TheBurchLog:
Building out functionality for centralize events manager. Objective is to create listeners for events that can spawn actions, such as parent notification, child notification, triggers, plugin communication.


src/app/beer_garden/config.py
```diff
@@ -790,6 +790,98 @@ def _parse_args(args: Sequence[str]) -> Tuple[YapconfSpec, dict]:
...
+                },
+                "ssl": {
+                    "type": "dict",
+                    "items": {
+                        "enabled": {
+                            "type": "bool",
+                            "default": False,
+                            "description": "Serve content using SSL",
+                            "previous_names": ["ssl_enabled"],
+                            "alt_env_names": ["SSL_ENABLED"],
```
hazmat345: We probably need to take these out.

TheBurchLog: These were removed

TheBurchLog:
The most recent Travis-CI failure is due to the annotations I placed on the DB layer. Any suggestions on how to handle this? We could move them down to the Mongo layer, but I wanted to keep them at a higher level.

TheBurchLog:
There was a three line add for the Brewtils/models.py file that I committed to v3. Did not do a new PR because it was such a small change.

src/app/beer_garden/__main__.py
```diff
@@ -26,6 +27,12 @@ def main():
     # Absolute first thing to do is load the config
     beer_garden.load_config(sys.argv[1:])
 
+    # Register the events handler before the application starts
+    #beer_garden.logger.info("Starting local events manager...")
+    #beer_garden.establish_events_manager()
+
+    beer_garden.establish_events_queue()
+
```
hazmat345: Let's move this out of this file - anything in `__main__` is really difficult to test, and I don't think there's a reason why we can't create the queue as part of the application startup.

src/app/beer_garden/__init__.py
```diff
@@ -36,6 +44,33 @@ def signal_handler(signal_number, stack_frame):
...
+def establish_events_queue(queue: Queue = None):
+    global events_queue
+
+    if queue is None:
+        context = multiprocessing.get_context("spawn")
+        queue = context.Queue()
+
+    events_queue = queue
+
+
```
hazmat345: I know we use this singleton pattern for other things, but that's mainly to avoid getting into a dependency-injection nightmare.

To me, having both the 'main' app and the entry points calling the same method make it less clear what's going on.

src/app/beer_garden/db/api.py
```diff
@@ -8,8 +10,25 @@
...
+
+
+@publish_event(Events.DB_CREATE)
+def update_function(obj):
+    return beer_garden.db.mongo.api.update(obj)
+
+
+@publish_event(Events.DB_DELETE)
+def delete_function(obj):
+    return beer_garden.db.mongo.api.delete(obj)
```
hazmat345: If we call these something like `_create` it'll be more apparent that they aren't meant to be accessed outside of this module.

src/app/beer_garden/__main__.py
TheBurchLog: I can work on moving this over to the app.py code.

src/app/beer_garden/__init__.py
TheBurchLog: Both Entry Points and the App need to call the same function. That way everything is configured the same for the Events Handler. I pulled it out of the Main App, but let's chat about the pro's and con's about them calling the same method.

src/app/beer_garden/db/api.py
TheBurchLog: Agreed, I will include these changes

TheBurchLog:
I changed the folder from "bg_event" to "events" but now that conflicts with the "events.py" file for the wrapper. So that got rolled up into events_manager.py. Then all of the files that use that dependency was converted over.

src/app/test/local_plugins/loader_test.py
```diff
@@ -152,6 +155,15 @@ def test_requirements_not_found(self, loader, registry):
...
 
+def mock_publish_event(event_type):
+    # TODO - This is kind of gross
+    @wrapt.decorator(enabled=True)
+    def wrapper(wrapped, _, args, kwargs):
+        return wrapped(*args, **kwargs)
+
+    return wrapper
+
+
```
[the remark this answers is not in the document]
TheBurchLog: Removed it

PR #370, opened 2019-12-02
Users

hazmat345:
This PR moves a bunch of the user and auth stuff out of the HTTP package.

PR #378, opened 2019-12-11
Issue 375

### LINKED ISSUE #375: Client Cert Login
We should be able to use client certs as an authentication method.

1. Allow users to auto login with certs instead of click the login option
2. Generate tokens based off the cert information
3. Make the cert extraction configurable

TheBurchLog:
This fixes #375. Allowing users to login with trusted certs

- [x] Verified that it still works when not in SSL
- [x] Verify that it still work when in SSL and user not in DB
- [x] Verify that auto login works
- [x] Verify that users can change to admin account in SSL
- [x] Verify that users can log into admin account not in SSL
- [x] Update SSL configuration documents

### DIFF
--- a/src/app/beer_garden/api/http/__init__.py
+++ b/src/app/beer_garden/api/http/__init__.py
@@ -156,7 +156,7 @@ def _setup_application():
     if http_config.behind_proxy.enabled:
         proxy_user_entity = http_config.behind_proxy.user_entity_object
         proxy_username = http_config.behind_proxy.user_entity_field
-        proxy_role_entity = http_config.behind_proxy.role_entity
+        proxy_role_entity = http_config.behind_proxy.role_entity_field
 
     tornado_app = _setup_tornado_app()
     server_ssl, client_ssl = _setup_ssl_context()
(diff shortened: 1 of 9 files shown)


loganasherjones:
Do we know if we're merging #370 in before this one or not? There will be lots of conflicts either way.

hazmat345:
I'd say probably this one before #370? That one is pretty raw and I'm not sure how much value it adds. It may be easier to just start from when this gets pulled into v3.

hazmat345:
@TheBurchLog would you like me to add a review on this? I haven't added any comments as it's marked WIP but I can take a look if you'd like.

TheBurchLog:
@hazmat345 We had talked about merging in the changes into this branch. Do you think we should still do that. I saw your comment about moving it over into v3 after these changes go in.

hazmat345 (commented):
Converting to draft - we'll need to completely revisit this.

TheBurchLog:
This is OBE now with the changes in https://github.com/beer-garden/beer-garden/pull/855 and NGINX proxy examples.

PR #383, opened 2019-12-22
Fixes #382

### LINKED ISSUE #382: UserPrincipal Not Iterable
Bug in v3.

Steps to reproduce:

```
$ git clone https://github.com/beer-garden/beer-garden
$ cd beer-garden/docker/docker-compose
$ docker-compose up -d
$ cd ../../src/app
$ python ./bin/app.py -c example_config/config.yaml
$ curl http://localhost:2337/api/v1/users/anonymous
```

loganasherjones:
This is really a problem for the anonymous principal. The
anonymous principal is actually a Brewtils object while the
MongoParser can only serialize Mongo objects. In the anonymous
case we need to fallback to the brewtils schema parser.

### DIFF
--- a/src/app/beer_garden/db/mongo/parser.py
+++ b/src/app/beer_garden/db/mongo/parser.py
@@ -30,7 +30,9 @@ class MongoParser(SchemaParser):
         }
     )
 
-    @staticmethod
-    def _get_schema_name(model):
+    @classmethod
+    def _get_schema_name(cls, model):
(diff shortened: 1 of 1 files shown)


hazmat345:
Looks good. Ideally once everything is completely stable we shouldn't ever be in the position where we need to serialize both brewtils and database flavors of the same model. But I think it's good to have a fallback instead of just raising an exception.

PR #385, opened 2019-12-27
WIP: Issue 362

### LINKED ISSUE #362: Child BG Registration
At start-up, BG instance should register with parent in a similar process to Remote Plug-ins / Systems.

Example Config:
```
parent:
    http:
        enable: false
        host: 0.0.0.0
        port: 2337
        public_fqdn: remotehost
        ssl:
          ca_cert: null
          ca_path: null
          client_cert_verify: NONE
          enabled: false
          private_key: null
          public_key: null
        url_prefix: /
        remote_username: bg_remote
        username: username
        password: password
```

TheBurchLog:
This ticket was originally allowing the child BG to register with the parent. While I was in there, it quickly steam rolled into creating all of the forwards for the events.

The PR does the following things:
- Register with parent BG and provides callback informaiton
- Forwards the following event types
  - Beer Garden Starting/Stoping
  - Request Start/Update/Compete
  - System Create/Update/Delete
  - Instance Initialized/Start/Update/Stopped
- Backs off when cannot connect
- Adds events back to queue if they cannot be processed 
  - Not sure how this will be impacted by things that have an order of operation (Request Start vs Compelted)
- When disconnected, does back off until it can reconnect
- Provides config option to skip specific event types

fixes #362


src/app/beer_garden/api/http/handlers/v1/garden.py
```diff
@@ -0,0 +1,156 @@
...
+from beer_garden.api.http.authorization import authenticated, Permissions
+from beer_garden.api.http.base_handler import BaseHandler
+
+
+class GardenAPI(BaseHandler):
+    @authenticated(permissions=[Permissions.SYSTEM_READ])
+    async def get(self, garden_name):
+        """
+        ---
+        summary: Delete a specific Instance
```
hazmat345: Need a new summary :smile:

src/app/beer_garden/config.py
```diff
@@ -1165,10 +1214,12 @@ def _parse_args(args: Sequence[str]) -> Tuple[YapconfSpec, dict]:
     "amq": _AMQ_SPEC,
     "application": _APP_SPEC,
     "auth": _AUTH_SPEC,
+    #"routing": _ROUTING_SPEC,
     "configuration": _META_SPEC,
     "db": _DB_SPEC,
     "entry": _ENTRY_SPEC,
     "event": _EVENT_SPEC,
+    "garden_name": _GARDEN_NAME_SPEC,
```
hazmat345: Can we reorganize this to:

```python
_GARDEN_SPEC = {
    "name": {
        ...
    }
}

...
"garden": _GARDEN_SPEC,
...
```

That way if we need to add anything to the garden part later we don't have to rearrange things.

src/app/beer_garden/api/http/handlers/v1/garden.py
TheBurchLog: What?!? That one wasn't good enough... Fine it has been changed.

src/app/beer_garden/config.py
TheBurchLog: That makes sense. I changed it. Should we move the Namespace information to this location as well?

PR #387, opened 2019-12-28
Mq refactor

hazmat345:
This PR moves some things out if the `instance` module into the `queue` package.

### DIFF
--- a/src/app/beer_garden/__init__.py
+++ b/src/app/beer_garden/__init__.py
@@ -2,7 +2,7 @@
 import logging
 import logging.config
 
-from brewtils.models import Request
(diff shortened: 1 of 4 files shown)


loganasherjones:
Sorry, in case my review is not clear. All the actual changes look good. I don't think this should be merged until the block comments are removed :)

hazmat345:
Ahh. Yeah, when I leave those in it's usually because the changes are part of a larger refactor and that functionality doesn't have a home yet. I don't want to forget about it so I just comment it out.

But I agree it's not really appropriate to have in a PR. I'll pull it out.

PR #392, opened 2020-01-04
Removing old, unnecessary v2 handlers

hazmat345:
This PR removes a bunch of the v2 handlers. I created those during pycon when I thought we would need a completely new API version. I don't think that's going to be necessary for the majority of endpoints, and even if we end up needing a v2 the current implementation of the handlers is wrong (because things have changed in the v1 handlers and v2 has not been kept in sync).

The only thing I didn't remove is the websocket (events module). I'm not sure how that will end up working, but the frontend is currently calling it, so I'm leaving it in for now.


TheBurchLog:
Looking at the difference between V1 and V2.
1) Inclusion of Namespace
2) Utilization of ExecutorClient

I do not believe that the Executor Client will work in the new structure of BGv3. The HTTP endpoint is spawned as a Stoppable Thread and cannot create new threads. It looks like the Executor Client creates a spawning pool and that should crash the first time it runs. 

Do you want to remove the client.py file as well?

PR #400, opened 2020-01-17
Event queues

hazmat345:
This PR moves and changes some things relating to the event system.

- The event processors have been mostly (I didn't move the http parent one) into the processors module. They've been renamed because there's nothing that's really Event specific and I'd like to use them for the logging queue stuff.
- We can now send events downstream to the entry points with a pipe connection. This is better than a queue in this case because it integrates with the asyncio nature of the HTTP entry point better.


hazmat345:
@TheBurchLog please let me know if you're opposed to this or see any issues that could impact you.

PR #404, opened 2020-01-23
Event changes

hazmat345:
This PR has more changes to the entry point and event systems.

I won't say these are 100% bulletproof, but with these changes they are getting pretty close to being stable I think.

We still need to really flesh out what an `Event` is and should look like.


hazmat345:
@TheBurchLog this is ready to review!

TheBurchLog:
@hazmat345 
Still working through the code review, but something you said earlier sparked a thought.

What you told me is that the local plugin kick off code waits for the Event from the HTTP Entry Points getting turned on, what happened when they disable the HTTP Entry point in the config? Does it just hang during initialization or does it not kick off the local plugins?

src/app/beer_garden/events/events_manager.py
```diff
@@ -1,59 +1,107 @@
...
+
+def publish_event(event_type):
+    @wrapt.decorator
+    def wrapper(wrapped, _, args, kwargs):
+        try:
+            result = wrapped(*args, **kwargs)
+
+            manager.create_event(event_type.name, result, False, args, kwargs)
+
+            return result
```
TheBurchLog: We were talking about this earlier and what values should be forwarded in an event. What you have here should be enough to derive any follow on action. If the event is a delete, then we can just look at the args to get the ID. I think adding any more variables to this would be overkill.

src/app/beer_garden/events/events_manager.py
```diff
@@ -1,59 +1,107 @@
...
+            ]
+        }
+
+    def register_forward(self, new_queue):
+        self.forward_queues.append(new_queue)
+
+    def register_callback(self, event_name: str, callback: Callable):
+        if event_name not in self.callbacks:
+            self.callbacks[event_name] = []
+        self.callbacks[event_name].append(callback)
```
TheBurchLog: When we are registering the callback, should it be a global declaration or function based?
Example: 
All request objects get ran the same way for callbacks
```
@publish_event(Events.REQUEST_CREATE)
def create_request(request):
``` 

Example 2:
Only events kicked off by this annotation get processed by ```do_function```
```
@publish_event(Events.REQUEST_CREATE, callback=do_function)
def create_request(request):
```

src/app/beer_garden/events/events_manager.py
```diff
@@ -1,59 +1,107 @@
...
             elif event.name in (Events.QUEUE_CLEARED.name, Events.SYSTEM_REMOVED.name):
                 event.payload = {"id": args[0]}
-            elif event.name in (Events.DB_CREATE.name, Events.DB_UPDATE.name):
-                event.payload = result
             elif event.name in (Events.DB_DELETE.name,):
                 event.payload = args[0]
-        finally:
-            upstream_queue.put(event)
 
-        return result
```
TheBurchLog: Thank you

src/app/beer_garden/app.py
```diff
@@ -183,38 +163,21 @@ def _startup(self):
...
         self.scheduler.start()
 
-        try:
-            self.events_manager.put(
-                brewtils.models.Event(name=Events.BARTENDER_STARTED.name)
-            )
-        except RequestException:
-            self.logger.warning("Unable to publish startup notification")
+        self.logger.debug("Publishing startup event")
+        publish(Event(name=Events.GARDEN_STARTED.name))
```
TheBurchLog: I just realized that in all of the places we try to publish an event we use  ```Events.EVENT_TYPE.name```. Should we just drop the ```.name```? Then we could do stuff like:
```
if event == event.EVENT_TYPE:
```
I don't foresee an scenario where we would not have access to the Event's object for this enumeration.

hazmat345: Yeah, we really only need the *string name* when we go to serialize it. I'm good with this, less to type :smile:

src/app/beer_garden/events/events_manager.py
hazmat345: Hahah :+1:

hazmat345:
@TheBurchLog ok, this is ready for another look.

- This reworks the event manager module to remove that entire class hierarchy.
- We no longer try to configure event handling / publishing in the actual `EntryPoint` class. There's just no good way to make that work for both sync and async entry point. So instead, I reverted to passing one end of the pipe into the main `run` function and letting the entry point itself deal with it. This is nice because we no longer have any extra threads running inside the http server process.

Unless you have major objections I'd like to merge this in on Monday. It's probably not perfect but I think we're at the point where we need to take the plunge and just fix any issues as we find them.

There's still work that needs to be done with regards to defining what an event should look like, but for the purposes of this PR that doesn't matter yet.

TheBurchLog:
When you get in, go ahead an merge the code.

PR #419, opened 2020-02-06
Consolidating routing logic in routing module

hazmat345:
This PR:

- moves all routing logic into the routing module
- streamlines error handling and argument passing in the routing logic

My thinking behind these changes is:
- I feel fairly strongly that the routing logic should be in one spot
- I don't think we need to (at least yet) pre-validate that a request we intent to route will eventually succeed. In other words, I don't think it's appropriate for the routing logic raise an exception to the effect of "You must provide an ID argument." The routing logic should only be concerned with invoking the correct business logic (or forwarding to the correct garden), and if the provided arguments are incorrect then the business logic should be raising an exception to that effect.

Other semi-random thought:
We should really talk about the Patch models. They're kind of a less-capable version of what we're trying to build out right now, and that puts them in a weird place. Basically, they exist because we can't directly express (for example) the difference between updating an instance's status and its heartbeat time directly though REST - we have to kind of shoehorn all possible update operations into the PATCH request. Hence, the Patch model.

However, what we're building out now is kind of like the Patch model on steroids - we need to be able to express EVERY possible operation in a single model. So I think we should try and see if we can confine the Patch operation to the REST api.

### DIFF
--- a/src/app/beer_garden/api/__init__.py
+++ b/src/app/beer_garden/api/__init__.py
@@ -92,7 +92,6 @@ get_all_queue_info = namespace_router(beer_garden.queues.get_all_queue_info)
 clear_queue = namespace_router(beer_garden.queues.clear_queue)
 clear_all_queues = namespace_router(beer_garden.queues.clear_all_queues)
 
-
 # Scheduler
 get_job = namespace_router(beer_garden.scheduler.get_job)
 get_jobs = namespace_router(beer_garden.scheduler.get_jobs)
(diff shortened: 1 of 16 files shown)


TheBurchLog:
Location Comment Response:
I went back and forth on where we should put the routing logic for the classes. At first I had it all in one class, but then I was thinking "Hey let's abstract this". If we placed it in the actual classes, maybe in the future we could do some class forwarding magic to say something like:
``` Python
class Route_Class(Enum):
    COMMAND = (brewtils.models.Command.schema, beer_garden.command)

    def __init__ (self, schema, classObj):
        self.schema = schema
        self.classObj = classObj

route_class.classObj.route_event(brewtils_obj=brewtils_obj, obj_id=obj_id, route_type=route_type, **kwargs)
```
Then we won't have to worry about all of the logic and the destination class takes it on.

Error Message Comment:
I agree with removing where I added the error message for ```obj_id```. If we do not have enough information to do the routing, we can make the determination that it is for a local action. That way the destination class can through the error instead.

Patch Model Comment:
Yes, this is the patch model on steroids for basic commands. 

I believe the PATCH model exists because at some point it was elected to not have a one to one mapping between the API and Service layer. This just happened to work out in my benefit. If we wanted to re-write how PATCH works in the future, let's revist how we invoke commands on plugins.

hazmat345:
Closing as this is superseded by #420.

PR #421, opened 2020-02-07
Adding system version column to request index page

### LINKED ISSUE #409: Request Index page show system version
We should add a column on the request index page with the system version.

hazmat345:
This fixes #409 - it adds a system version column to the request index page.


hazmat345:
Merging this.

PR #423, opened 2020-02-11
Events

hazmat345:
This PR publishes and processes events across the parent / child boundary.

It depends on beer-garden/brewtils#189.


src/app/beer_garden/events/__init__.py
```diff
@@ -0,0 +1,81 @@
...
+
+    Args:
+        event: The event to publish
+
+    Returns:
+        None
+    """
+    # Do some formatting / tweaking
+    if not event.garden:
+        event.garden = config.get("garden.name")
```
TheBurchLog: Chat with me about how the ```event.garden``` field would get populated outside of this. If this is only being used to send events from child to parent, then I don't see a problem with setting it, but we need to make sure we don't have any logic later in the process that looks up that garden_name. Anything past child relationships is obfuscated to the parent because they shouldn't care about grandchildren.

src/app/beer_garden/events/handlers.py
```diff
@@ -0,0 +1,62 @@
...
+                db.create(event.payload)
+
+            elif event.name in (
+                Events.REQUEST_STARTED.name,
+                Events.REQUEST_COMPLETED.name,
+                Events.SYSTEM_UPDATED.name,
+                Events.INSTANCE_UPDATED.name,
+            ):
+                db.update(event.payload)
+
```
TheBurchLog: This is the JAVA side having questions, why do you trust that the payload field is accurate? Should you run it through the respective module function to update the record?

``` Python
elif event.name in (
    Events.REQUEST_STARTED.name, 
    Events.REQUEST_COMPLETED.name
):
    beer_garden.system.update(event.payload)
```

PR #425, opened 2020-02-12
Abstract Routing

TheBurchLog:
Utilizing pointers to the module functions that will invoke the action requested by the Forward Request.


src/app/beer_garden/router.py
```diff
@@ -18,56 +17,178 @@
...
+    :return: Boolean response on if the object should be forwarded
+    """
+    # if forward.target_garden_name is None:
+    #     for arg in forward.args:
+    #         if isinstance(arg, (brewtils.models.Request, brewtils.models.RequestTemplate)):
+    #             forward.target_garden_name = get_system_mapping(
+    #                 name_space=arg.namespace,
+    #                 system=arg.system,
+    #                 version=arg.system_version,
+    #             )
```
hazmat345: Block-commented code needs to be removed before merging.

TheBurchLog: I can clean this up. I commented this to ensure it didn't run while testing. This should be functional after we get the plug-in registration working to make to `namespaces` and `garden_names`

TheBurchLog:
Changing the name to something like `Operation` sounds fine with me. The name `Forward` is not intuitive to what is actually occurring.

I just updated my PyCharm to use Google for docstrings. So I can work on converting them over to that format. By hey, I actually wrote them this time :)

hazmat345:
Hahahah, making progress :smile:

src/app/beer_garden/api/http/__init__.py
```diff
@@ -202,6 +203,7 @@ def _setup_tornado_app():
         # These are a little special - unpublished but still versioned
         # The swagger spec
         (rf"{prefix}api/v1/spec/?", misc.SpecHandler),
+        (rf"{prefix}api/v1/forward/?", v1.forward.ForwardAPI),
```
hazmat345: :+1:

PR #426, opened 2020-02-14
Plugin Default Namespace

TheBurchLog:
Provided API for plugins to grab the namespace and save with it.


TheBurchLog:
Going to cancel this one, we can re-write the API for the UI when we get done the dependant modules.

PR #430, opened 2020-02-19
System/Namespace registration

TheBurchLog:
When a system is created, it updates the garden mapping with the new information. It will also cache that information for the router module.

### DIFF
--- a/src/app/beer_garden/app.py
+++ b/src/app/beer_garden/app.py
@@ -20,8 +20,12 @@ from beer_garden.api.entry_point import EntryPoint
 from beer_garden.db.mongo.jobstore import MongoJobStore
 from beer_garden.db.mongo.pruner import MongoPruner
 from beer_garden.events import publish
-from beer_garden.events.handlers import local_callbacks, downstream_callbacks, \
-    system_mapping_callback, garden_mapping_callback
+from beer_garden.events.handlers import (
+    local_callbacks,
+    downstream_callbacks,
+    system_mapping_callback,
+    garden_mapping_callback,
+)
 from beer_garden.events.processors import (
     FanoutProcessor,
     HttpEventProcessor,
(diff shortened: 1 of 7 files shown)


src/app/beer_garden/app.py
```diff
@@ -20,7 +20,12 @@
 from beer_garden.db.mongo.jobstore import MongoJobStore
```
hazmat345: need to import beer_garden.garden here

src/app/beer_garden/app.py
```diff
@@ -257,6 +268,8 @@ def _setup_events_manager(self):
 
         event_manager.register(QueueListener(action=local_callbacks))
         event_manager.register(QueueListener(action=downstream_callbacks))
+        event_manager.register(QueueListener(action=system_mapping_callback))
+        event_manager.register(QueueListener(action=garden_mapping_callback))
```
hazmat345: Hmm, this makes me think we should talk about how to structure the callback listeners. I agree this is good for now because you want to apply the callback regardless of whether it's coming from the local garden or a downstream (right?)

src/app/beer_garden/router.py
```diff
@@ -26,13 +26,7 @@
...
 
 System_Garden_Mapping = {}
-child_connections: Dict[str, Garden] = {
-    "child": Garden(
-        name="child",
-        connection_type="HTTP",
-        connection_params={"public_fqdn": "localhost", "port": 2338},
-    )
-}
+child_connections: Dict[str, Garden] = dict()
```
hazmat345: :cry:

src/app/beer_garden/app.py
TheBurchLog: Yes, I want to handle it regardless of where it  is coming from. If a child forwards a system create, then I want to cache that and store it with the Garden object. We can sync about how we want to handle these callbacks. I think we could also drop the parent_http_processor.py module because you have it covered in your callback for events.

PR #432, opened 2020-02-21
Change Forward Event Garden

TheBurchLog:
When a Garden forwards an event to their parent, the event will have the forwarding Garden's name applied to it. This is because Parents have no concepts of Grand Children and would not know where the event came from if a Grand Child sent the event. This field is only for routing, so this is should be fine for these scenarios because the payload is not changed.

Also, added in the feature to skip events in the forwarding.

### DIFF
--- a/src/app/beer_garden/events/processors.py
+++ b/src/app/beer_garden/events/processors.py
@@ -104,14 +104,16 @@ class FanoutProcessor(QueueListener):
 class HttpEventProcessor(QueueListener):
     """Publish events using an EasyClient"""
 
-    def __init__(self, easy_client=None, **kwargs):
+    def __init__(self, easy_client=None, skip_events=[], **kwargs):
         super().__init__(**kwargs)
 
         self._ez_client = easy_client
+        self.skip_events = skip_events
 
-    def process(self, item: Event):
+    def process(self, event: Event):
         try:
-            item.garden = beer_garden.config.get("garden.name")
-            self._ez_client.publish_event(item)
+            if event.name not in self.skip_events:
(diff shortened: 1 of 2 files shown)


hazmat345:
@TheBurchLog how would you feel about having the `skip_events` functionality live in the more general case? I think that's something that *any* processor could be interested in, not just the `HttpEventProcessor`.

TheBurchLog:
I think we should have categories for the skip events. Making it a general field might have negative impacts, but make it general enough that all forward processing can use it, makes sense.

'''
skip_events:
    forward:
        - DB_UPDATE
"""

TheBurchLog:
Added the black_list to the base processor.

@hazmat345  What are your thoughts on making the assumption that everything we use the processor class for is of type `Event`? That is the assumption that I made for this.

src/app/beer_garden/events/processors.py
```diff
@@ -13,13 +14,15 @@
 class BaseProcessor(StoppableThread):
     """Base Processor"""
 
-    def __init__(self, action=None, **kwargs):
+    def __init__(self, action=None, black_list=list(), **kwargs):
         super().__init__(**kwargs)
 
         self._action = action
+        self._black_list = black_list
```
hazmat345: Two minor things:
1. It's standard to use the python syntax for empty arrays (and dicts) as opposed to the constructor-style.
2. It's not a good idea to have a mutable type be the default value for a parameter.

So this would be better as:
```python
def __init__(self, ..., black_list=None, ...):
   ...
   self._black_list = black_list or []
```

PR #433, opened 2020-02-21
Child Event Checks

TheBurchLog:
This checks the MongoDB model to determine any unique fields required for storage. Then queries the database to determine if the record already exists. If the record already exists, then it verifies that the records are the same based on the ID.

Child BG instances can only create new records in their parent and update records that they know of. If a child tries to overwrite something the parent or other child created (i.e. identical plugin deployments) then the parent will reject the event request.

** Also removed DB events, this caused a loop in the code.

### DIFF
--- a/src/app/beer_garden/db/api.py
+++ b/src/app/beer_garden/db/api.py
@@ -14,17 +14,14 @@ reload = beer_garden.db.mongo.api.reload
 replace_commands = beer_garden.db.mongo.api.replace_commands
 
 
-@publish_event(Events.DB_CREATE)
 def _create(obj):
     return beer_garden.db.mongo.api.create(obj)
 
 
-@publish_event(Events.DB_CREATE)
 def _update(obj):
     return beer_garden.db.mongo.api.update(obj)
 
 
-@publish_event(Events.DB_DELETE)
 def _delete(obj):
     return beer_garden.db.mongo.api.delete(obj)
 
--- a/src/app/beer_garden/db/api.py
+++ b/src/app/beer_garden/db/api.py
@@ -13,19 +13,6 @@ query = beer_garden.db.mongo.api.query
 reload = beer_garden.db.mongo.api.reload
 replace_commands = beer_garden.db.mongo.api.replace_commands
 
-
-def _create(obj):
-    return beer_garden.db.mongo.api.create(obj)
-
-
-def _update(obj):
-    return beer_garden.db.mongo.api.update(obj)
-
-
-def _delete(obj):
-    return beer_garden.db.mongo.api.delete(obj)
-
-
-create = _create
-update = _update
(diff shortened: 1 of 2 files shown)


hazmat345:
I'm not sure I'm a fan of the way this is organized... we're reaching down into mongo-specific things outside of the db package. Is there another way we can accomplish this?

TheBurchLog:
Matt, I added those changes. Let me know what you think.

src/app/beer_garden/db/api.py
```diff
@@ -14,17 +14,14 @@
 replace_commands = beer_garden.db.mongo.api.replace_commands
 
 
-@publish_event(Events.DB_CREATE)
```
hazmat345: If we're going to remove these decorators then we can remove these function defs and just assign the mongo-specific functions to the api ones (like above)

src/app/beer_garden/events/handlers.py
```diff
@@ -49,25 +53,53 @@ def downstream_callbacks(event: Event) -> None:
...
             return
 
-        try:
-            if event.name in (Events.REQUEST_CREATED.name, Events.SYSTEM_CREATED.name):
-                db.create(event.payload)
+        if not event.payload_type:
+            logger.error(
+                f"Unable to process event ({event} : {event.payload_type}: {event.payload}): No Payload Type"
+            )
+            return
```
hazmat345: Is there a reason we need this here? It seems reasonable that we could have an event without a payload at all (if we were only publishing fact-of occurrence or something).

src/app/beer_garden/events/handlers.py
```diff
@@ -49,25 +53,53 @@ def downstream_callbacks(event: Event) -> None:
...
+        ):
+
+            model_class = getattr(brewtils_models, event.payload_type)
+            record = db.query_unique(model_class, id=event.payload.id)
+
+            if record:
                 db.update(event.payload)
+            else:
+                logger.error(
+                    f"Unable to update ({event} : {event.payload_type} : {event.payload}): Object does not exists in database"
```
hazmat345: Should be

> ... does not exist in ...

src/app/beer_garden/events/handlers.py
TheBurchLog: I moved it down to the statements where I require that field. So now if we publish an update/delete event without the model type, we will throw an error stating that we can't process something we don't know what it is.

src/app/beer_garden/events/handlers.py
TheBurchLog: fixed

src/app/beer_garden/db/api.py
TheBurchLog: I removed the function definitions.

PR #440, opened 2020-02-25
Adding a reverse_delete_rule for the Command model's system field

### LINKED ISSUE #141: Mongo reverse_delete_rules
We currently have two `Reference` fields that don't define a `reverse_delete_rule`, so it defaults to doing nothing. We should be explicit about how we want to handle these deletes.

1. The `system` field in the `Command` model. I'm pretty sure this should be `CASCADE` since it doesn't really make sense to have 'orphan' commands. In fact, I believe that caused an issue for us at some point, so this may help address that.
2. The `parent` field of the `Request` model. I'm not too sure about this one, but I'm leaning towards `DENY` to prevent the ageoff from removing a parent request but leaving the child. If that were to happen now I'm pretty sure any attempt to pull that request out would result in an error when mongoengine tried to dereference the deleted parent.

hazmat345:
This PR fixes #141 by adding a `reverse_delete_rule` for the Command model's system field.

### DIFF
--- a/src/app/beer_garden/db/mongo/models.py
+++ b/src/app/beer_garden/db/mongo/models.py
@@ -198,6 +198,7 @@ class Command(MongoModel, Document):
     form = DictField()
     template = StringField()
     icon_name = StringField()
+    # reverse_delete_rule=CASCADE is set later, can't set until System is defined
     system = ReferenceField("System")
 
     def clean(self):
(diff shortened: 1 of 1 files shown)


hazmat345:
@TheBurchLog oh man, no. This is terrible :cry: 

The Request child field is a `DummyField` - it doesn't actually exist. We have to manually construct that field whenever we want to include it. If you look at `requests.get_request` you can see that we manually mangle that field before we return it.

I would support changing this to something more sane. :smile:

PR #441, opened 2020-02-26
Updated how Garden updates work

TheBurchLog:
Moved the PATCH back to the API
Updated Routing Logic to only accept events from the local garden or the one who forwarded thee event. 
Removed the PATCH logic from garden.py and added config update logic.


src/app/beer_garden/api/http/handlers/v1/garden.py
```diff
@@ -109,17 +111,38 @@ class GardenAPI(BaseHandler):
...
+                        args=[garden_name, "RUNNING"],
+                    )
+                )
+            elif operation == "config":
+                response = await self.client(
+                    Operation(
+                        operation_type="GARDEN_UPDATE_CONFIG",
+                        args=[
+                            garden_name,
+                            SchemaParser.parse_garden(op.payload, from_string=True),
```
hazmat345: Have you tested this? I don't believe a patch operation has a payload field.

src/app/beer_garden/garden.py
```diff
@@ -35,30 +35,13 @@ def get_gardens() -> List[Garden]:
...
-            garden = update_garden_status(garden_name, operation.upper())
-        elif operation == "heartbeat":
-            garden = update_garden_status(garden_name, "RUNNING")
-
-        else:
-            raise ModelValidationError(f"Unsupported operation '{op.operation}'")
-
-    return garden
+    update_garden(db_garden)
+    return db_garden
```
hazmat345: Every function in this module should be returning the result of the update:

```python
return update_garden(db_garden)
```

src/app/beer_garden/api/http/handlers/v1/garden.py
TheBurchLog: You are right, it should have been `op.value`

src/app/beer_garden/garden.py
TheBurchLog: Fixed

PR #443, opened 2020-02-27
Debugging Finds

TheBurchLog:
I was testing out v3 and discovered that we don't cache the current gardens at startup. So I updated that and also updated how we handle garden update events to cache the results.

Discovered that we were mapping the wrong field from the config for default namespace for plugins. Fixed that mapping

### DIFF
--- a/src/app/beer_garden/systems.py
+++ b/src/app/beer_garden/systems.py
@@ -72,7 +72,7 @@ def create_system(system: System) -> System:
             system.max_instances = len(system.instances)
 
     if system.namespace is None:
-        system.namespace = beer_garden.config.get("namespaces.name")
+        system.namespace = beer_garden.config.get("namespaces.local")
 
     system = db.create(system)
 
--- a/src/app/example_configs/config.yaml
+++ b/src/app/example_configs/config.yaml
@@ -104,7 +104,7 @@ event:
         public_key: null
       url_prefix: /
 garden:
-  name: default
+  name: parent_garden
 log:
   config_file: example_configs/logging-config.yaml
   file: null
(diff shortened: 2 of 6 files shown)


src/app/example_configs/config.yaml
```diff
@@ -104,7 +104,7 @@ event:
         public_key: null
       url_prefix: /
 garden:
-  name: default
+  name: parent_garden
```
hazmat345: This shouldn't be updated in version control

src/app/example_configs/config.yaml
```diff
@@ -116,7 +116,7 @@ metrics:
     port: 2339
     url: http://localhost:3000
 namespaces:
-  local: default
+  local: parent
```
hazmat345: This shouldn't be updated in version control

src/app/beer_garden/systems.py
```diff
@@ -72,7 +72,7 @@ def create_system(system: System) -> System:
             system.max_instances = len(system.instances)
 
     if system.namespace is None:
-        system.namespace = beer_garden.config.get("namespaces.name")
+        system.namespace = beer_garden.config.get("namespaces.local")
```
hazmat345: Should this be using the garden name instead? I think we should get rid of that whole namespaces config section as it doesn't really make sense anymore.

hazmat345:
@TheBurchLog I added a couple things, but I'm going to go ahead and merge this.

- I reverted the changes to the config file
- The entry points need to know about the gardens from the database as well.

src/app/example_configs/config.yaml
TheBurchLog: yeah, that should not have made it in

PR #446, opened 2020-02-27
Websocket reconnect

hazmat345:
This PR makes the frontend check the status of the websocket connection every five seconds. It will attempt to reconnect if not currently connected.


hazmat345:
Going to merge this.

PR #453, opened 2020-03-03
Added Delete System to routable

TheBurchLog:
Added the Delete System to forwarding options

### DIFF
--- a/src/app/beer_garden/router.py
+++ b/src/app/beer_garden/router.py
@@ -19,7 +19,12 @@ import beer_garden.systems
 from beer_garden.errors import RoutingRequestException, UnknownGardenException
 
 # These are the operations that we will forward to child gardens
-routable_operations = ["INSTANCE_START", "INSTANCE_STOP", "REQUEST_CREATE"]
+routable_operations = [
+    "INSTANCE_START",
+    "INSTANCE_STOP",
+    "REQUEST_CREATE",
+    "SYSTEM_DELETE",
+]
 
 # Processor that will be used for forwarding
 forward_processor = None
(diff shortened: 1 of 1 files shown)


hazmat345:
OK, I figured out the problem I was seeing. Try this:

1. Start the parent garden with Plugin A.
2. Stop the parent garden.
3. Start the parent garden with Plugin B.
4. Start the child garden with plugin C.

So now the parent garden should know about A, B, and C. A is stopped, B and C are running.

Now try to delete A. Instead of deleting it I see the parent garden attempt to route that to the child, which is wrong.

TheBurchLog:
Before I got through testing this, when you say "with Plugin A", do you mean that it is a local plugin? Then in the second start, you remove the local plugin and add a new one?

hazmat345:
Yes, that's accurate.

TheBurchLog:
Just ran this scenario with Remote  Plugins, and everything worked as expected. Going to try it with local plugins next.

TheBurchLog:
Just did the scenario:
1) Install Plugin A locally on Parent
2) Start Parent
3) Stop Parent
4) Uninstall Plugin A on Parent
5) Install Plugin B Locally on Parent
6) Start Parent
7) Install Plugin C Remotely on Child
8) Start Child
9) Delete Plugin A on Parent
10) Waited/refreshed, then it was removed successfully

TheBurchLog:
Followed the example you initially provided, and cannot replicate it.

hazmat345:
Alrighty, we can go ahead and merge this. I'm still not sure what's going on, but at this point I don't believe it's an issue with *these* changes.

PR #454, opened 2020-03-04
Added forwarding for Rescan requests

TheBurchLog:
This feature modifies the scope of the rescan button. We had two options for how we could handle this. I coded the first option.

1) When someone requests a `Re-Scan`, BG will run it locally and forward this request to all child `Gardens` because `Name Spaces` can span multiple `Gardens`.
2) When someone request a `Re-Scan`, BG will run it locally and  forward the request to all `Garden` deployments that currently host that `Name Space`.


hazmat345:
@TheBurchLog I think this is OBE - can we close?

hazmat345:
This is OBE - closing.

PR #455, opened 2020-03-04
Garden admin

hazmat345:
This PR adds a garden admin page and does some cleanup to the frontend.

It also changes the garden rest api to `/api/v1/gardens`


TheBurchLog:
I just tried to run it. I can now see the Garden page but it does not populate. Is that expected?

src/ui/src/js/configs/routes.js
```diff
@@ -66,24 +82,23 @@ export default function routeConfig($stateProvider, $urlRouterProvider, $locatio
       },
     })
     .state('login', {
-      url: '/login',
       templateUrl: basePath + 'login.html',
       controller: 'LoginController',
     })
     .state('base.namespace.systems', {
-      url: '/systems',
+      url: 'systems/',
```
TheBurchLog: Why did the change occur across all connections in this file?

src/ui/src/index.html
```diff
@@ -173,17 +173,23 @@
...
                 <li ng-show="hasPermission(user, 'bg-user-read')">
-                  <a ui-sref="user_admin" ui-sref-active="active">
+                  <a ui-sref="base.user_admin" ui-sref-active="active">
                     <i class="fa fa-user"></i>
                     <span>User Management</span>
                   </a>
                 </li>
                 <li ng-show="hasPermission(user, 'bg-role-read')">
-                  <a ui-sref="role_admin" ui-sref-active="active">
+                  <a ui-sref="base.role_admin" ui-sref-active="active">
```
TheBurchLog: We didn't have the prefix `base` here before. Why add it now? Was this an issue with how the UI worked previously or something that adding the `namespaces` to the UI introduced?

src/ui/src/index.html
```diff
@@ -173,17 +173,23 @@
...
                     <i class="fa fa-tag"></i>
                     <span>Role Management</span>
                   </a>
                 </li>
+                <li ng-show="hasPermission(user, 'bg-role-read')">
+                  <a ui-sref="base.garden_admin" ui-sref-active="active">
+                    <i class="fa fa-leaf"></i>
```
TheBurchLog: Perfect icon

hazmat345:
@TheBurchLog this is a great start! There are a couple of bugs I ran into, though.

- When I try to select a Connection Type things go crazy - a bunch of connected systems show up.
- The system name deconstruction breaks for the `echo-sleeper` system, I'm guessing because it has a hyphen?

Regarding the second bullet - I was thinking that it would be easier if we had the `systems` attribute of the Garden model be a list of actual systems, not just names. How would you feel about that?

TheBurchLog:
@hazmat345 
I was able to recreate the bug where it spams out a bunch of systems in the table. I will take a look. I probably reused a object name.

I like the idea of just having they `System` object on attached to the `Garden` object. That will also expose out the number of instances deployed and the current status. That way they could quickly see if one `Namespace` on a `Garden` is down.

TheBurchLog:
@hazmat345 
Fixed the table rendering issue, and I think the System Link issue. Can you test it in your environment?

hazmat345:
@TheBurchLog looks like those two issues are fixed for me.

Only things now:
- It looks like the connection info is setting the `hostname` field... this should probably be `host` to match all the other configuration values.
- The router is looking for that value in the `public_fqdn` field, it should be looking at the `host` field.

TheBurchLog:
@hazmat345 
Made the changes to convert them over `host` when stored in the database. It will still say `Host Name` on the UI, but it will be stored as `host`. I also made those changes in the router.py.

src/ui/src/js/controllers/admin_garden_view.js
```diff
@@ -0,0 +1,99 @@
...
+      var system = systems[key]
+      var systemObj = {};
+
+      systemObj['namespace'] = system.split(":")[0];
+
+      var systemVersion = system.split(":")[1]
+      var start = systemVersion.lastIndexOf('-');
+      var end = systemVersion.length;
+
+      systemObj['name'] = systemVersion.substring(0, start);;
```
hazmat345: Extra `;`

src/ui/src/js/controllers/admin_garden_view.js
```diff
@@ -0,0 +1,99 @@
...
+      systemObj['version'] = systemVersion.substring(start+1, end);
+
+      splitSystems.push(systemObj);
+    }
+
+    return splitSystems
```
hazmat345: Missing `;`

hazmat345:
I like how we're using schema form to do the form :+1:

TheBurchLog:
> I like how we're using schema form to do the form +1

Cool thing about using it is that if we add more parameters in the schema, the connection params field will automatically include it. So this should make it easy to add new connection types.

hazmat345:
LGTM!

PR #457, opened 2020-03-07
Refactoring

hazmat345:
This does a lot. Recommend looking at the commit messages.


src/app/beer_garden/app.py
```diff
@@ -287,8 +281,20 @@ def _setup_scheduler():
...
+        known_namespaces = beer_garden.namespace.get_namespaces()
+
+        garden = Garden(
+            name=beer_garden.config.get("garden.name"),
+            status="RUNNING" if event == Events.GARDEN_STARTED else "STOPPED",
+            systems=known_systems,
+            namespaces=known_namespaces,
+        )
 
         publish(Event(name=event.name, payload_type="Garden", payload=garden))
```
TheBurchLog: This could cause issues. When I was first looking at this, when you add all known systems and namespaces to the garden, then publish it, the `Router` class would take that are the new routing information. So it would never try to route to a downstream garden, it would always try to publish it locally.

To verify that this is not a problem, please give this a test:
1) Deploy Parent Garden
2) Deploy Child Garden w/ plugins
3) Task Child Plugin
4) Restart Parent Garden
5) Task Child Plugin

src/app/beer_garden/events/processors.py
```diff
@@ -113,21 +109,16 @@ def register(self, processor, start: bool = True):
 class HttpEventProcessor(QueueListener):
     """Publish events using an EasyClient"""
 
-    def __init__(self, easy_client=None, **kwargs):
+    def __init__(self, easy_client=None, black_list=None, **kwargs):
```
TheBurchLog: Sorry I forgot to move that

src/app/beer_garden/garden.py
```diff
@@ -133,3 +116,29 @@ def garden_add_system(system: System, garden_name: str):
...
+    """Handle garden-related events
+
+    For GARDEN events we only care about events originating from downstream. We also
+    only care about immediate children, not grandchildren.
+
+    Whenever a garden event is detected we should update that garden's database
+    representation.
+
+    This method should NOT update the routing module. Let its handler worry about that!
+    """
```
TheBurchLog: This is an interesting way to handle the events. So anything that would impact the DB layer from downstream events would be processed here.

src/app/beer_garden/requests.py
```diff
@@ -740,11 +693,11 @@ def complete_request(
     return request
 
 
-def cancel_request(request: Request) -> Request:
+def cancel_request(request_id: Request) -> Request:
     """Mark a Request as CANCELED
```
TheBurchLog: Do we need to have a Event type for cancellation?

src/app/beer_garden/requests.py
```diff
@@ -753,12 +706,31 @@ def cancel_request(request: Request) -> Request:
...
+                db.create(event.payload)
+
+        elif event.name in (Events.REQUEST_STARTED.name, Events.REQUEST_COMPLETED.name):
+            # When we send child requests to child gardens where the parent was on
+            # the local garden we remove the parent before sending them. Only setting
+            # the subset of fields that change "corrects" the parent
+            existing_request = db.query_unique(Request, id=event.payload.id)
+
+            for field in ("status", "output", "error_class"):
+                setattr(existing_request, field, getattr(event.payload, field))
```
TheBurchLog: This addresses my one comment above, now that I understand how you are handling the events. This should maintain the parent field, without the child needing to know it was a spawned request.

src/app/beer_garden/router.py
```diff
@@ -154,127 +183,182 @@ def forward(operation: Operation):
...
-    target_garden = None
-    for garden in gardens:
-        if garden.name == operation.target_garden_name:
-            target_garden = garden
-            break
-
-    if not target_garden:
+    try:
+        conn_type, conn_info = garden_connections[operation.target_garden_name]
+    except KeyError:
```
TheBurchLog: Still reading through, but would it be an error if `conn_type` and `conn_info` are `None`? If the admin has not configured the garden, then that will happen. We could raise an exception saying that Garden is not configured, please talk to your admins.

src/app/beer_garden/router.py
```diff
@@ -154,127 +183,182 @@ def forward(operation: Operation):
...
         )
 
-    if target_garden.connection_type is None:
-        target_garden = beer_garden.garden.get_garden(target_garden.name)
-        cache_gardens()
+    try:
+        if conn_type.casefold() == "http":
+            return _forward_http(operation, conn_info)
+        else:
+            raise RoutingRequestException(f"Unknown connection type {conn_type}")
```
TheBurchLog: What about `https`?

src/app/beer_garden/router.py
```diff
@@ -154,127 +183,182 @@ def forward(operation: Operation):
...
+    local_garden_name = config.get("garden.name")
 
+    # We do NOT want to load local garden information from the database as the local
+    # name could have changed
+    for garden in get_gardens():
+        if garden.name != local_garden_name:
+            if (
+                garden.connection_type is not None
+                and garden.connection_type.casefold() != "local"
+            ):
```
TheBurchLog: I will add local as a `connection_type` in the UI. That way we have a way for them to manage that if we store it in the database.

src/app/beer_garden/router.py
```diff
@@ -154,127 +183,182 @@ def forward(operation: Operation):
...
+                # Add to the connection lookup
+                garden_connections[garden.name] = (
+                    garden.connection_type,
+                    garden.connection_params,
+                )
+            else:
+                logger.warning(f"Adding garden with invalid connection info: {garden}")
 
+    # Now add the local systems
+    local_systems = get_systems(filter_params={"local": True})
```
TheBurchLog: I am guessing you added this to the model in Brewtils? Btw, I like this approach more. It is easier to figure out what is not local.

src/app/beer_garden/router.py
hazmat345: Consolidating this down to just http because the actual http vs https decision will be determined by the ssl boolean flag.

src/app/beer_garden/router.py
hazmat345: Yeah, I did. I'm not sure how much I like this - still very much a WIP

src/app/beer_garden/requests.py
hazmat345: Yeah, that would probably be helpful.

src/app/beer_garden/api/http/handlers/v1/request.py
```diff
@@ -85,18 +85,41 @@ class RequestAPI(BaseHandler):
...
+                elif op.path == "/output":
+                    operation.kwargs["output"] = op.value
+
+                elif op.path == "/error_class":
+                    operation.kwargs["error_class"] = op.value
+
+                else:
+                    raise ModelValidationError(f"Unsupported path '{op.path}'")
+            else:
+                raise ModelValidationError(f"Unsupported operation '{op.operation}'")
```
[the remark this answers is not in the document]
hazmat345: You cut it off one line too late :smile: 

Line 92 is the `for` loop that iterates though each operation in the Patch. We iterate over each "Patch Operation" and construct a "Router Operation". We then execute that Router Operation.

src/app/beer_garden/app.py
```diff
@@ -180,18 +180,12 @@ def _startup(self):
...
+        beer_garden.router.setup_routing()
+
         self.logger.debug("Starting forwarding processor...")
         beer_garden.router.forward_processor.start()
 
-        self.logger.debug("Setting up local garden information")
-        garden = beer_garden.garden.get_garden(beer_garden.config.get("garden.name"))
-        if garden is None:
-            garden = Garden(name=beer_garden.config.get("garden.name"))
-            beer_garden.garden.create_garden(garden)
```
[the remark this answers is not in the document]
hazmat345: I changed a lot of stuff in this area, but I want to make sure I didn't miss anything. Is the scenario that you're targeting here:

1. Stand up a new beer-garden with a completely new database.
2. Start a plugin in that garden.
3. Ensure that plugin is routable immediately

src/app/beer_garden/db/mongo/api.py
```diff
@@ -193,6 +201,8 @@ def query_unique(model_class: ModelType, **kwargs) -> Optional[ModelItem]:
         query_set = _model_map[model_class].objects.get(**kwargs)
         return to_brewtils(query_set)
     except DoesNotExist:
+        if raise_missing:
+            raise
```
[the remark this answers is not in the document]
hazmat345: The default in the parameter listing for this is `False`, so for places that are already using this function there's no change.

I added this because in some places I *need* to raise an exception when something isn't found.

src/app/beer_garden/db/mongo/api.py
```diff
@@ -279,7 +289,7 @@ def create(obj: ModelItem) -> ModelItem:
     if hasattr(mongo_obj, "deep_save"):
         mongo_obj.deep_save()
     else:
-        mongo_obj.save()
+        mongo_obj.save(force_insert=True)
```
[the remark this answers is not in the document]
hazmat345: They're unrelated. `deep_save` is something that we made up so that saving a `System` will save the nested models (instances and commands) as well.

This `force_insert` flag makes a save operation raise an exception if that model already exists in the database. Without it "save" basically means "save or update".

src/app/beer_garden/events/processors.py
hazmat345: No worries :smile:

src/app/beer_garden/app.py
hazmat345: This works as expected for me.

src/app/beer_garden/requests.py
hazmat345: Added an Event in the brewtils branch.

src/app/beer_garden/garden.py
hazmat345: Yeah, things got to the point where I wasn't able to reason about what was going on the event handling side of things. So I split things out so that each module has its own event handler function that will handle the events that it (and only it) cares about.

src/app/beer_garden/router.py
hazmat345: I added a check for this case.

src/app/beer_garden/app.py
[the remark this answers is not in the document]
TheBurchLog: If all of that works, then I am good

src/app/beer_garden/app.py
TheBurchLog: :+1:

src/app/beer_garden/router.py
TheBurchLog: :+1:

src/app/beer_garden/router.py
TheBurchLog: I made changes in the UI to reflect this as well.

src/app/beer_garden/db/mongo/api.py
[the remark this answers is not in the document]
TheBurchLog: ok, sounds good to me

src/app/beer_garden/db/mongo/api.py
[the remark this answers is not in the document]
TheBurchLog: sounds good

src/app/beer_garden/api/http/handlers/v1/request.py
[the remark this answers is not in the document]
TheBurchLog: ok, I must have just missed that one. Then this is good

src/app/beer_garden/router.py
TheBurchLog: Until you find something better, I am good with it.

src/app/beer_garden/router.py
```diff
@@ -84,29 +93,31 @@ def route(operation: Operation):
...
+    if operation.target_garden_name == config.get("garden.name"):
+        return execute_local(operation)
 
-def should_forward(operation: Operation) -> bool:
-    """Routing logic to determine if an operation should be forwardedl
+    # Otherwise, validate that it can be forwarded
+    if operation.operation_type not in routable_operations:
+        raise RoutingRequestException(
+            f"Operation type '{operation.operation_type}' can not be forwarded"
+        )
```
TheBurchLog: Should you run this check before executing local?

hazmat345: This is kind of the heart of this change. I want to inspect the operation and use it to determine which garden it SHOULD be routed to.

Then, if it turns out that the operation is not allowed to be routed, I can raise the correct exception here.

The alternative can be confusing. For example, say we decided to not allow forwarding for system delete events. If somebody tries to delete a system from a child garden and we did things the old way:
- We'd check if the operation was routable. It's not, so we'd attempt to execute it locally.
- That system doesn't exist locally, so we raise a "not found" exception
- The user is confused because the system is right there

This way allows us to be a little more accurate with the failure cases.

hazmat345:
Alrighty, merging this :crossed_fingers:

PR #463, opened 2020-03-09
Metrics server should only be started in main process

hazmat345:
This PR fixes an issue where both the main process and the HTTP entry point would attempt to start the metrics server. Since they both tried to use the same port the HTTP entry point would error.

Also, starting the same thing in both processes is not needed :smile:


hazmat345:
Merging this

PR #464, opened 2020-03-09
Config updates

hazmat345:
This PR makes some tweaks to the config subsystem:

- Removes the `brew_view` events section that is OBE
- No longer freezes the config when loading it. Freezing is nice, but it breaks the case where the server secret hasn't been set as we're unable to update it.
- The spec default for enabling sending event notifications to a parent needs to be `False` - the default needs to be standalone mode.

### DIFF
--- a/src/app/beer_garden/config.py
+++ b/src/app/beer_garden/config.py
@@ -769,7 +769,7 @@ _EVENT_SPEC = {
                     "items": {
                         "enable": {
                             "type": "bool",
-                            "default": True,
+                            "default": False,
                             "description": "Run an HTTP server",
                         },
                         "ssl": {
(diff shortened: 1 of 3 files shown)


src/app/beer_garden/config.py
```diff
@@ -811,7 +769,7 @@ def _parse_args(args: Sequence[str]) -> Tuple[YapconfSpec, dict]:
                     "items": {
                         "enable": {
                             "type": "bool",
-                            "default": True,
+                            "default": False,
                             "description": "Run an HTTP server",
                         },
```
TheBurchLog: Could you also change the description for this in your PR. `Run an HTTP server` is not accurate. This is to publish events to an HTTP server.

PR #465, opened 2020-03-09
Deployment

hazmat345:
This PR starts to bring the deployment artifacts in line with the new project structure:
- Services files have been removed (bartender and brew-view) and updated (beer-garden)
- The bartender and brew-view services in the compose file have been consolidated into a single beer-garden service
- The python version that's bundled with the rpm has been bumped to 3.7.4
- Files related to the python 2 build have been removed
- Pushes to the `v3` branch will build and push the `bgio/beer-garden:unstable` tag

Still TODO - currently there is no service in the compose file that will serve the UI. It's still necessary to run the npm dev-server to get the UI.

Also, there are a bunch of changes to `.travis.yml`. The ruby thing I used to encrypt the Docker credentials "helpfully" changed everything. The only real changes in that file are the actual credentials and the branch that `unstable docker deploy` runs on.


hazmat345:
We'll start auto-deploying the unstable docker images, yeah.

For the service file, we can make it even easier. But yes, that's another PR :smile:

PR #469, opened 2020-03-19
Webpack

hazmat345:
This PR updates frontend dependency versions. It also makes some changes to the `EventService` to accommodate removing a superfluous dependency.


TheBurchLog:
Nothing looks crazy, but I won't be able to run it till Monday to make sure the `npm install` works as expected. 

A test I want to do is make sure that the socket closes after the server shuts down. I see you are checking for this with `socketConnection.readyState == WebSocket.CLOSED`, but I have had experiences in other languages where the flag doesn't get set right. I also want to see if we have to refresh to re-establish the socket. Since we don't actually reload the page, this could get people into a weird state if they never close/refresh their browser but are waiting for updates to show.

hazmat345:
So there's definitely reconnect logic - if you look in `run.js` I'm calling `connect` every 5 seconds. This does nothing if the connection is already established, but if the server is actually down it logs an error and there's no way to prevent that. It'd probably be better to implement a progressive backoff with a maximum similar to what the plugins do to minimize that but I just didn't bother.

It does work correctly if the beer-garden server goes down and comes back up. I haven't read the websocket spec enough to understand how it determines a connection is severed, but I haven't seen any issues where it should have tried to reconnect and didn't.

PR #470, opened 2020-03-19
Frontend cleanup

hazmat345:
This is a minor PR that removes an unnecessary `div` and stops the browser from requesting `favicon.ico` before we can tell it the actual favicon file name.


hazmat345:
Gonna merge this one

PR #471, opened 2020-03-20
Added button to download output

### LINKED ISSUE #361: Download request output
We should make an easy way for UI users to download or view the raw output in another tab / window.

west270:
Add button to download output as a json, txt, or html file.

Fixes #361

### DIFF
--- a/src/ui/src/index.js
+++ b/src/ui/src/index.js
@@ -155,6 +155,7 @@ angular.module('bgApp',
 .config(['localStorageServiceProvider', function(localStorageServiceProvider) {
   localStorageServiceProvider.setStorageType('sessionStorage');
 }])
+//Adds urls to whitelist to prevent undefined url for download link
 .config(['$compileProvider', function ($compileProvider) {
   $compileProvider.aHrefSanitizationWhitelist(/^\s*(https?|ftp|mailto|tel|file|data):/);
 }])
--- a/src/ui/src/index.js
+++ b/src/ui/src/index.js
@@ -155,6 +155,9 @@ angular.module('bgApp',
 .config(['localStorageServiceProvider', function(localStorageServiceProvider) {
   localStorageServiceProvider.setStorageType('sessionStorage');
 }])
+.config(['$compileProvider', function ($compileProvider) {
+  $compileProvider.aHrefSanitizationWhitelist(/^\s*(https?|ftp|mailto|tel|file|data):/);
+}])
 .service('APIInterceptor', interceptorService)
 .service('authInterceptorService', authInterceptorService)
 .animation('.slide', slideAnimation)
--- a/src/ui/src/js/controllers/request_view.js
+++ b/src/ui/src/js/controllers/request_view.js
@@ -33,6 +33,30 @@ export default function requestViewController(
     request) {
   $scope.service = RequestService;
 
+ $scope.download = function() {
+      var filename = $scope.data.id;
+      var output;
+      var outputType = $scope.data.output_type;
+      if(outputType.toLowerCase() == "string"){
+        output = $scope.rawOutput;
+        filename = filename+".txt"
+      } else if(outputType.toLowerCase() == "html"){
+        output = $scope.htmlOutput;
+        filename = filename+".html"
+      } else if(outputType.toLowerCase() == "json"){
+        output = $scope.jsonOutput;
+        filename = filename+".json"
+      }
+      var element = document.createElement('a');
+      element.setAttribute('href', 'data:text/plain;charset=utf-8,' + encodeURIComponent(output));
+      element.setAttribute('download', filename);
+      element.style.display = 'none';
+      document.body.appendChild(element);
+      element.click();
+      document.body.removeChild(element);
+      return;
+ };
+
   $scope.instanceStatus = undefined;
   $scope.timeoutRequest = undefined;
 
--- a/src/ui/src/js/controllers/request_view.js
+++ b/src/ui/src/js/controllers/request_view.js
@@ -37,7 +37,10 @@ export default function requestViewController(
       var filename = $scope.data.id;
       var output;
       var outputType = $scope.data.output_type;
-      if(outputType.toLowerCase() == "string"){
+      if($scope.data.output === undefined || $scope.data.output == null){
+        window.alert("No output to download");
+        return;
+      } else if(outputType.toLowerCase() == "string"){
         output = $scope.rawOutput;
         filename = filename+".txt"
       } else if(outputType.toLowerCase() == "html"){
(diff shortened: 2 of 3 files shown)


src/ui/src/partials/request_view.html
```diff
@@ -94,7 +94,11 @@ <h1 class="page-header">
     <div class="row">
       <div class="col-md-6">
         <div style="margin-bottom: 10px;">
-          <span class="h3">Output</span>
+          <span class="h3">Output
+            <button type="button" class="btn pull-right" ng-click="download();">
```
TheBurchLog: The semicolon after `download()` is not needed. We use that when calling straight JS. When referencing functions in the HTML tags, we have not been using it. Both ways still work.

src/ui/src/js/controllers/request_view.js
```diff
@@ -33,6 +33,33 @@ export default function requestViewController(
     request) {
   $scope.service = RequestService;
 
+ $scope.download = function() {
+      var filename = $scope.data.id;
+      var output;
+      var outputType = $scope.data.output_type;
```
hazmat345: It's a *lot* easier to declare variables with `let` instead of `var`. The rules regarding when and where `var` variables exist are really complicated and not worth it. Trust me, just use `let` everywhere :smile:

src/ui/src/js/controllers/request_view.js
```diff
@@ -33,6 +33,33 @@ export default function requestViewController(
...
+      if($scope.data.output === undefined || $scope.data.output == null){
+        window.alert("No output to download");
+        return;
```
hazmat345: This is a good check! How about we also hide the button until we have something to download? That way we can prevent the users from making this error (which is nicer than alerting them (which is nicer than breaking :smile: )) 

See if you can figure out how to make this happen from the AngularJS docs, but take a look at the "Pour it again" button for something similar if you get stuck.

src/ui/src/partials/request_view.html
```diff
@@ -94,7 +94,11 @@ <h1 class="page-header">
...
+              <i class="fa fa-download" ></i> Download
```
hazmat345: Let's see how this looks without the "Download" text - the download icon is pretty ubiquitous so we can probably get away with something subtler.

src/ui/src/js/controllers/request_view.js
```diff
@@ -33,6 +33,33 @@ export default function requestViewController(
...
+      } else if(outputType.toLowerCase() == "string"){
+        output = $scope.rawOutput;
+        filename = filename+".txt"
+      } else if(outputType.toLowerCase() == "html"){
+        output = $scope.htmlOutput;
+        filename = filename+".html"
+      } else if(outputType.toLowerCase() == "json"){
+        output = $scope.jsonOutput;
+        filename = filename+".json"
+      }
```
hazmat345: It's definitely good to use the output type to determine the file extension. But I think you can simplify this and just use `$scope.rawOutput` everywhere. The other things are for presentation / formatting, and if someone is downloading the output we need to give them the *real* (raw) output, not the version we modified to look pretty for them.

src/ui/src/js/controllers/request_view.js
```diff
@@ -33,6 +33,33 @@ export default function requestViewController(
...
+      var element = document.createElement('a');
+      element.setAttribute('href', 'data:text/plain;charset=utf-8,' + encodeURIComponent(output));
+      element.setAttribute('download', filename);
+      element.style.display = 'none';
+      document.body.appendChild(element);
+      element.click();
+      document.body.removeChild(element);
+      return;
```
hazmat345: I'm not sure what's going on here, we'll have to chat about this one :smile:

west270: It creates a hyperlink for download then clicks hyperlink, then removes hyperlink. I will change the download button to a hyperlink of the code above since it does not need to create a hyperlink just to delete the hyperlink.

src/ui/src/js/controllers/request_view.js
west270: Will do, I did not think of hiding button which is better than an alert.

src/ui/src/partials/request_view.html
```diff
@@ -95,9 +95,8 @@ <h1 class="page-header">
       <div class="col-md-6">
         <div style="margin-bottom: 10px;">
           <span class="h3">Output
-            <button type="button" class="btn pull-right" ng-click="download();">
-              <i class="fa fa-download" ></i> Download
-            </button>
+              <a class="fa fa-download pull-right" ng-href="{{downloadHref}}" download="{{filename}}"
+                 style="visibility:{{downloadVisible}};"></a>
```
hazmat345, line 99: Angular has a special directive for this, `ng-show` - let's use that instead of setting the style.

src/ui/src/js/controllers/request_view.js
```diff
@@ -113,19 +86,21 @@ export default function requestViewController(
     $scope.formatErrorMsg = undefined;
 
     let rawOutput = $scope.data.output;
-
+    let downloadHref = 'data:text/plain;charset=utf-8,' + encodeURIComponent($scope.data.output);
```
hazmat345, line 89: This is much cleaner :smile:

src/ui/src/index.js
```diff
@@ -157,6 +157,9 @@ angular.module('bgApp',
 .config(['localStorageServiceProvider', function(localStorageServiceProvider) {
   localStorageServiceProvider.setStorageType('sessionStorage');
 }])
+.config(['$compileProvider', function ($compileProvider) {
+  $compileProvider.aHrefSanitizationWhitelist(/^\s*(https?|ftp|mailto|tel|file|data):/);
```
hazmat345, line 161: Let's add a little comment here explaining that this is for the download link.

src/ui/src/js/controllers/request_view.js
```diff
@@ -84,21 +84,22 @@ export default function requestViewController(
...
+        $scope.filename = $scope.request.id+".html";
         $scope.htmlOutput = rawOutput;
         $scope.formattedAvailable = true;
         $scope.showFormatted = true;
       } else if ($scope.request.output_type == 'JSON') {
         try {
           let parsedOutput = JSON.parse(rawOutput);
           rawOutput = $scope.stringify(parsedOutput);
-
+          $scope.filename = $scope.data.id+".json";
```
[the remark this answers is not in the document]
TheBurchLog: Totally forgot about that. Good catch.

TheBurchLog:
I am going to say this is ready to be checked in. @hazmat345 do you have any objections? 

@west270 If we don't hear back from @hazmat345 by COB, go ahead and merge in your code.

PR #473, opened 2020-03-23
Consolidate Indexs

TheBurchLog:
When viewing System records, it is now in a table view that allows the user to click on the breadcrumbs at the top to jump around. This will be added to the Commands page in another PR to make everything feel consistent.


src/ui/src/partials/system_index.html
```diff
@@ -1,44 +1,39 @@
...
-    <div class="panel-body" style="padding-top: 0;">
-      <div>
-        <button type="button" class="btn btn-primary btn-lg btn-block"
-            ng-click="exploreSystem(system)" style="margin-top: 10px;">
-          Explore {{system.name}} commands
-        </button>
-      </div>
-    </div>
+<div ng-if="responseState(response) === 'success'">
+  <h1 class='page-header' ng-bind-html="getBreadCrumbs()"></h1>
```
hazmat345: I think this should be tweaked before merging. It's really constructing HTML in the controller, which we should avoid if we can (I know there are places where we don't follow this rule, but I think we should try to if possible). 
```
ng-bind-html="getBreadCrumbs()"
```

We do something similar in the request view for children requests:

```

  <ol class="tagline breadcrumb" ng-show="request_tree.length > 1">
    <li ng-repeat="request in request_tree = getParentTree(data)" ng-class="{active: $last}"
      ><span ng-if="!$last"><a ui-sref="base.namespace.request({requestId: request.id})">{{request.command}}</a></span
      ><span ng-if="$last">{{request.command}}</span>
    </li>
  </ol>
```

TheBurchLog:
I have narrowed down where the error message is getting throw. It is when the table is generated. If I comment out the `datatable="ng"` from the following snippet, the error goes away. Still trying to debug to figure out why this is happening.

```html
    <table
        datatable="ng"
        dt-options="dtOptions"
        class="table table-striped table-bordered"
        style="width: 100%">
```

src/ui/src/partials/system_index.html
TheBurchLog: Changed this to pulled the variables from the `$scope` object. Also wrote it so if we copy and past it somewhere else, it "should" work, like in on the commands page.

PR #477, opened 2020-03-24
Moved Menu Around

TheBurchLog:
~~This changed the side menu to only have system information. Everything else is at the top.~~

~~Issue:~~
~~The Namespaces would not collapse for me. I don't know if this is because it is a dynamic list.~~

~~Another Note:~~
~~I just randomly put them in an order, if we want to re-arrange them we can.~~

This now moves everything to the top bar. Giving the user the entire width of their display. The logo in the top left is now a drop down menu to allow them to quickly navigate to a system. It searches all fields of the plugin, including description. Hoverover of the systems give the system description as well.

### DIFF
--- a/src/ui/src/index.html
+++ b/src/ui/src/index.html
@@ -26,7 +26,11 @@
       <ul class="nav navbar-nav navbar-top-links navbar-right">
 
         <li ng-show=!isUser(user)>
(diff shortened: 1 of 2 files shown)


hazmat345:
@TheBurchLog I'm going to go ahead and merge this!

PR #479, opened 2020-03-26
Merge commands system view

TheBurchLog:
Now the System View and Commands Index use the same source code. This depends on #473


hazmat345:
I'm going to close this as it's really part of #473.

PR #493, opened 2020-06-08
Create sub menus

TheBurchLog:
Trying to clean up how we have our drop down menu in the top right.

Previous the menu was as followed:
```
About
---------
default
slate
--------
Systems
Queues
Users
Roles
Gardens
```

Now the menus are broken out into three sections: Main, Themes, and Admin. When you click on the option it will open the sub-menu. Not sure if only Themes should be like this, but I went ahead and did the Admin actions the same way to see how it looks.

### DIFF
--- a/src/ui/src/index.html
+++ b/src/ui/src/index.html
@@ -105,7 +105,7 @@
                   <div style="padding: 3px 15px; cursor: pointer;"
                      ng-click="setMenuPage('main')">
                   <span class="fa fa-chevron-left"></span>
-                  <span>Themes</span>
+                  <span><b>Themes</b></span>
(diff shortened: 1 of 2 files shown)


hazmat345:
If we're going to keep the titles on the submenus I think they need to be more different from the rest of the options - it's hard to tell that they're titles right now. Maybe bold?

TheBurchLog:
Now it has been bolded

TheBurchLog:
@hazmat345 At COB today, I plan on merging in this code. If you have any more comments on how the menu looks please let me know. If not, this is something we can update later. All functionality works as expected.

PR #494, opened 2020-06-09
Spill large String output to GridFS storage

### LINKED ISSUE #407: Use GridFS to handle large output
Mongo does not allow storing documents larger than 16MB, so any output larger than that currently causes the request to fail.

It would be nice to transparently use GridFS to make outputs larger than 16MB work.

This comes from #308.

TheBurchLog:
Within the Mongo model for Requests we check the size of the output field. If it is larger that 16MB, then we spill it over to a File object within Mongo. Then when we dump that object for serialization we convert back over the object to the String field. 

fixes #407 

I will need to create an example plugin that replicates this. For now, I have this shoved into one of my plugins

```python

    @command()
    def large_output(self):

        print("Running Large Output")
        mb = 16
        output = ""
        N = 10000
        json_obj = {"long_array":[]}
        sub_obj = {"Animals":["Dog","Cat","Farret","Wombat","Snake"],
                   "Location":"Behind You",
                   "Worth":1000,
                   "Scare_Level":5001}

        while len(output.encode('utf-16')) < (1000000 * mb):
            for x in range(0, N):
                json_obj['long_array'].append(sub_obj)

            output = json.dumps(json_obj)

        print(len(output.encode('utf-16')) / 1000000)

        return output
```

### DIFF
--- a/src/app/beer_garden/db/mongo/models.py
+++ b/src/app/beer_garden/db/mongo/models.py
@@ -88,6 +88,8 @@ class MongoModel:
     def clean_update(self):
         pass
 
+    def pre_serialize(self):
+        pass
 
 # MongoEngine needs all EmbeddedDocuments to be defined before any Documents that
 # reference them. So Parameter must be defined before Command, and choices should be
@@ -284,7 +286,7 @@ class Request(MongoModel, Document):
     )
     children = DummyField(required=False)
     output = StringField()
-    output_gridfs = FileField(required=False)
+    output_gridfs = FileField()
     output_type = StringField(choices=BrewtilsCommand.OUTPUT_TYPES)
     status = StringField(choices=BrewtilsRequest.STATUS_LIST, default="CREATED")
     command_type = StringField(choices=BrewtilsCommand.COMMAND_TYPES)
@@ -361,12 +363,11 @@ class Request(MongoModel, Document):
 
     logger = logging.getLogger(__name__)
 
-    def dumps(self, *args, **kwargs):
+    def pre_serialize(self):
         """If string output was over 16MB it was spilled over to the GridFS storage solution"""
         if self.output_gridfs:
-            self.output = self.output_gridfs.read()
+            self.output = self.output_gridfs.read().decode("utf-8")
             self.output_gridfs = None
-        super(Request, self).dumps(*args, **kwargs)
 
     def save(self, *args, **kwargs):
         self.updated_at = datetime.datetime.utcnow()
--- a/src/app/beer_garden/db/mongo/models.py
+++ b/src/app/beer_garden/db/mongo/models.py
@@ -376,7 +376,7 @@ class Request(MongoModel, Document):
 
         # If the output size is too large, we switch it over
         # Max size for Mongo is 16MB, switching over at 15MB to be safe
-        if self.output and sys.getsizeof(self.output) < (1000000 * 15):
+        if self.output and sys.getsizeof(self.output) > (1000000 * 15):
             self.output_gridfs.put(self.output, encoding="utf-8")
             self.output = None
 
--- a/src/app/beer_garden/db/mongo/models.py
+++ b/src/app/beer_garden/db/mongo/models.py
@@ -372,8 +372,8 @@ class Request(MongoModel, Document):
         self.updated_at = datetime.datetime.utcnow()
 
         # If the output size is too large, we switch it over
-        if self.output and len(self.output.encode('utf-8')) > (1000000 * 16):
-            self.output_gridfs.put(self.output, encoding='utf-8')
+        if self.output and len(self.output.encode("utf-8")) > (1000000 * 16):
+            self.output_gridfs.put(self.output, encoding="utf-8")
             self.output = None
 
         super(Request, self).save(*args, **kwargs)
--- a/src/app/beer_garden/db/mongo/models.py
+++ b/src/app/beer_garden/db/mongo/models.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 import datetime
 import logging
+import sys
 
 import pytz
 import six
@@ -374,7 +375,8 @@ class Request(MongoModel, Document):
         self.updated_at = datetime.datetime.utcnow()
 
         # If the output size is too large, we switch it over
-        if self.output and len(self.output.encode("utf-8")) > (1000000 * 16):
+        # Max size for Mongo is 16MB, switching over at 15MB to be safe
+        if self.output and sys.getsizeof(self.output) < (1000000 * 15):
             self.output_gridfs.put(self.output, encoding="utf-8")
             self.output = None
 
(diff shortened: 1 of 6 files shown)


src/app/beer_garden/garden.py
```diff
@@ -130,7 +130,7 @@ def handle_event(event):
     This method should NOT update the routing module. Let its handler worry about that!
     """
     if event.garden != config.get("garden.name"):
-        if event.name in (Events.GARDEN_STARTED.name, Events.GARDEN_UPDATED.name,):
+        if event.name in (Events.GARDEN_STARTED.name, Events.GARDEN_UPDATED.name):
```
hazmat345, line 133: I think this is causing a merge conflict...

I'm not *opposed* to changing stuff like this (especially if it fixes the linting) but I think for things like this we should just make the change directly in `v3` and push it up. That way we don't complicate PRs with "extra" changes. I think we can skip the review for this stuff :smile:

src/app/beer_garden/db/mongo/models.py
```diff
@@ -358,8 +360,21 @@ class Request(MongoModel, Document):
 
     logger = logging.getLogger(__name__)
 
+    def dumps(self, *args, **kwargs):
+        """If string output was over 16MB it was spilled over to the GridFS storage solution"""
+        if self.output_gridfs:
+            self.output = self.output_gridfs.read()
+            self.output_gridfs = None
```
hazmat345, line 367: I really wish we had a way to tell if this has a significant impact on our db performance. I'm scared this is incurring an entire DB roundtrip for every request we load, which *could* be very bad, but I don't know that for sure.

TheBurchLog:
Once we get the scheduler code working, I will kick off a job that does this every 5 seconds, then we will see how the database looks after an hour.

src/app/beer_garden/garden.py
TheBurchLog: I agree, I normally try to not check in these formatting  changes when it doesn't impact my code. This one got slipped in on accident. I will revert this one.

TheBurchLog:
One impact to performance I have seen is the loading of the Request View page. It takes a minute for the object to get rendered when it is a JSON blob around 16MB.

TheBurchLog:
I just started to test with larger file sizes and it stopped working. Going to investigate why it is not running the dump code.

src/app/beer_garden/db/mongo/models.py
TheBurchLog: This does cause for a long delay when trying to pull down a single request. Rendering a file that large just takes a while.

When we load the Requests page we have no additional latency because the output field is not returned from the query. 

Can we think of any other times we load requests from the database.

TheBurchLog:
Got it working, but I don't like it. I added a pre_serialization function to the base MongoModel. Then before we serialize any object for Brewtils, I call that function. The issue was that not all objects had that function, even though I added it to the base mode. So I had to verify that the function existed before trying to execute it.

TheBurchLog:
@hazmat345 I had all kinds of memory issues when trying to test this. I kept using functions that bloated the in memory size. If you can think of a better way to generate the output, let me know.

It was god awful slow trying to render the larger output, but it eventually loaded. The thing that I thought was funny is we actually have code to say that the output is to large to display, but it only counts the keys in the top level of the dictionary, instead of getting an accurate sizing. If that was working, maybe bringing back the data is fine, but we just don't display it and throw that error message?

hazmat345:
I tweaked `large_output` to be more reliable. Unfortunately it's more boring now, though :smile: 

I'd definitely be on board with expanding that check to work better / be applicable across all output types.

My hesitation as it relates to *this* PR is that previously we had a hard limit of 16MB. So no matter how inefficient our frontend is we would never be overly taxed from a pure size perspective. This removes that, so we could theoretically have output in the GBs.

Regardless of how much beyond the limit a specific request is, it'd be nice to see that before the frontend requests the entire thing (since we couldn't display it anyway). It'd be more of a "sorry, this output is too big. Click here to download it" kind of thing instead of waiting for the whole request to "load" but then not displaying it.

hazmat345:
Also, I'm not sure why the Travis status isn't reporting but it looks like tests are failing here.

TheBurchLog:
@hazmat345 I agree with having a different PR to handle the UI. This is purely the ability to store it, which we should now be able to do that. This does make it possible to deal with large amounts of data. Honestly, how cool would it be if someone used BG plugins to do pre-processing of data for machine learning or something, In those scenarios they would need to be able to store large outputs, but don't actually want to view the results.

TheBurchLog:
@hazmat345 I have no clue why this travis is not running. I think this is the one that has been hanging on me for a while.

TheBurchLog:
@hazmat345 I just found some unit tests that are failing because of the gridfs. Going to see if fixing them resolves our issue.

src/app/beer_garden/db/mongo/models.py
```diff
@@ -358,8 +364,21 @@ class Request(MongoModel, Document):
...
+        if self.output_gridfs:
+            self.output = self.output_gridfs.read().decode("utf-8")
+            self.output_gridfs = None
+
     def save(self, *args, **kwargs):
         self.updated_at = datetime.datetime.utcnow()
+
+        # If the output size is too large, we switch it over
+        # Max size for Mongo is 16MB, switching over at 15MB to be safe
+        if self.output and sys.getsizeof(self.output) < (1000000 * 15):
```
hazmat345, line 378: How was this demonstrated to work? This needs to be `>`.

TheBurchLog: hmm.... You are correct..... I don't know why I have it backwards....

TheBurchLog:
@hazmat345 If you still have that large request in your database, could you do a test for me?

In ../controllers/request_view.js could you go to line 145 and update it to the following?

```javascript
$scope.successCallback = function(request) {
    request.output = null;
    $scope.request = request;
```

Then see if you get the long load time for the page? I am curious if it is the loading the object into the browsers memory is the issue or processing it to be viewed. I am guessing that the latency is on the loading in the browser.

hazmat345:
@TheBurchLog nulling out the output means that everything works ok (things are snappy). The v3 branch in example plugins has the fixed large_output command.

TheBurchLog:
@hazmat345 That is good news. So if we implement the fix on the UI for check the size before rendering then we should be in the clear.

hazmat345:
Yeah, I'm pretty sure this is a browser or angular issue... at some point there's just too much text to deal with.

TheBurchLog:
ok, I couldn't  help myself. I added the check in the browser to determine the in-memory footprint of an object. Then if it exceeds a certain amount we give them the warning. I don't know what that threshold should be at the moment. So I have it set to 5 MBs. Then I did my test with creating a 50MB request and it loads the UI in a reasonable amount of time. BTW, 50MB string creates a memory footprint of double that apparently.

hazmat345:
Hahahah my evil plan succeeded!

This works a lot better. It still takes a little bit of time to render the page even when it doesn't end up showing the output. And using data urls for the download is not realistic for output on the order of megabytes (waiting minutes to download something is way too long). So there are still a couple problems, but in the interest of not moving the goal posts *again* I'll open new issues for them.

PR #495, opened 2020-06-11
Return Local Garden Info

TheBurchLog:
When viewing the Gardens on the UI we couldn't see the local garden information. These changes will inject the Local Garden object when someone requests all of the Gardens. When someone attempts to make changes to the Garden it will not persist because it does not have an ID. When you restart with a new Garden name it will change that Garden info. If you select a Garden Name of a downstream Garden, you will see both in the UI but with different comments next to them (Local vs Remote).

### DIFF
--- a/src/app/beer_garden/garden.py
+++ b/src/app/beer_garden/garden.py
@@ -111,7 +111,6 @@ def create_garden(garden: Garden) -> Garden:
         The created Garden
 
     """
-    garden.status = "INITIALIZING"
     garden.status_info["heartbeat"] = datetime.utcnow()
 
     return db.create(garden)
--- a/src/app/beer_garden/garden.py
+++ b/src/app/beer_garden/garden.py
@@ -9,6 +9,7 @@ from brewtils.models import Events, Garden, System
 import beer_garden.config as config
 import beer_garden.db.api as db
 from beer_garden.events import publish_event
+from beer_garden.namespace import get_namespaces
 from beer_garden.systems import get_systems
 
 logger = logging.getLogger(__name__)
@@ -25,7 +26,8 @@ def get_garden(garden_name: str) -> Garden:
 
     """
     if garden_name == config.get("garden.name"):
-        return get_local_garden()
+        return local_garden()
+
     return db.query_unique(Garden, name=garden_name)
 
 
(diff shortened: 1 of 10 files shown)


src/app/beer_garden/garden.py
```diff
@@ -33,7 +36,34 @@ def get_gardens() -> List[Garden]:
...
+        Garden Object
+
+    """
+    local_garden = Garden(
+        name=config.get("garden.name"),
+        connection_type="LOCAL",
+        systems=get_systems(filter_params={"local": True}),
+        status="RUNNING",
+        status_info="This is your local Garden record, not persisted in the Database. "
+        "Changes will not be reflected",
```
hazmat345, line 60: I think we should take this out - this field isn't actually a string. This feels like something we should just display in the frontend when see the `LOCAL` connection type.

TheBurchLog: We will drop this field completely. I will see about throwing an alert when they go to edit a local instance.

TheBurchLog:
@hazmat345 
I added an alert window to the Admin Garden Viewer to alert the user that the field was auto generated at startup and is not stored in the database. Let me know if that message is clear enough and then we can merge this in.

TheBurchLog:
@hazmat345 I will try running here in a bit. We do need both controllers because one handles the overall view of the Gardens and the other is the unique view of each Garden. They both use the same Service layer. Just being consistent with how we render the other pages.

TheBurchLog:
@hazmat345 I got the browser issues fixed and updated the warning message. So it all should be working. This is what I get for writing code that I "know" works without testing it. But I got it running this time.

hazmat345:
I'm seeing a bunch of console errors on the specific garden page. I cleared my garden collection and then stood up a new child garden. I can see that garden on the main page and it's correctly marked as a remote, but when I view the specific page for it the table that should allow me to change the connection info isn't shown.

hazmat345:
@TheBurchLog this is ready for you to review... can't mark you as a reviewer since you opened this.

TheBurchLog:
@hazmat345 Sorry, this should have been next on my list to work on. I blew away my database and rebuilt everything. The child BG instance was running with the latest V3 code base. The parent BG instance was running with the latest from this branch. I am getting an error on the Edit Page for the Child BG instance. 

child_garden_errors

hazmat345:
@TheBurchLog yeah, it's because there's no `options` specified for that table. I already fixed it in v3 c74b1ea2 (it wasn't an issue related to this PR).

hazmat345:
@TheBurchLog fysa - I have some additional changes I've been working on related to how the garden and routing modules work... there were some corner cases we weren't handling correctly. But it turns out that's a larger rabbit hole than I have the time to explore right now, so I've put it on hole. I'm gonna go ahead and merge this now and I'll pick back up later.

PR #496, opened 2020-06-11
508 Compliance

### LINKED ISSUE #23: Make BEERGARDEN Accessible
We should verify that BEERGARDEN can be used by everybody

TheBurchLog:
This is the first round of 508 Compliance testing. Utilized the ANDI package sweet to inspect each page, then used Google's LightHouse application to get a score.

Pages Checked:

- [x] Systems = 98%
- [x] Requests = 90%
- [x] Commands = 98%
- [x] Scheduler (blank) = 100%
- [x] Scheduler (Populated) = 
- [x] Create Job System Index = 98%
- [x] Create Job Command Index = 98%
- [x] Create Job Page = 90%
- [x] Create Request = 90%
- [x] View Request = 87%
- [x] Admin Systems = 98%
- [x] Admin Queues = 100%
- [x] Admin Users = 100%
- [x] Admin Roles = 100%
- [x] Admin Gardens = 100%
- [x] About = 97%

Fixes #23

### DIFF
--- a/src/ui/src/index.html
+++ b/src/ui/src/index.html
@@ -30,7 +30,7 @@
         <li class="dropdown">
           <span data-toggle="dropdown" style="padding: 9px 15px; display: block;">
             <i class="fa fa-search"></i>
-            <input ng-model="searchText">
+            <input ng-model="searchText" aria-label="Search System Names">
           </span>
 
           <ul class="dropdown-menu">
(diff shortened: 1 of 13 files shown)


TheBurchLog:
Found an error with creating jobs that we need to fix, so was unable to score that page. Using the projection from the other pages like it, it would be around 98%. With that in mind, our average score now is 96% accessibility according to the Google Audit tool Light House.

Light house can be used in Google Chrome by Developers Tool UI, then going to the Audit Tab. I reduced the check list to "Accessibility" only.

TheBurchLog:
@west270 I believe that your changes should be independent enough that we can check these into v3 separately. We should not see any major merge conflicts.

Issue #170, opened 2018-09-05
Resize output / parameter displays


hazmat345:
It would be useful to be able to control the size / fullscreen those displays on the request view page.

hazmat345:
Also, when the JSON output is nested deeply the horizontal scroll is messed up.

west270:
![maximize_example1](https://user-images.githubusercontent.com/61320354/84502528-7fec2900-ac86-11ea-8543-7db756c289fb.png)

![maximize_example2](https://user-images.githubusercontent.com/61320354/84502514-76fb5780-ac86-11ea-82c7-a24c0d4e36a8.png)
@TheBurchLog here are the two examples we talked about for make output cell larger.

TheBurchLog:
@west270 Looking at the two, the first one (with request table) looks nicer. The primary reason why this was asked for was to display custom html outputs. Could you run a command that creates an html table with 10 columns and maybe 10 rows? Then we can point them to this post to see what to expect.

Here is a quick example:

```html
<table>
  <tr>
    <th>header</th>
    <th>header</th>
    <th>header</th>
    <th>header</th>
    <th>header</th>
    <th>header</th>
    <th>header</th>
    <th>header</th>
    <th>header</th>
    <th>header</th>
  </tr>
  <tr>
    <td>data</td>
    <td>data</td>
    <td>data</td>
    <td>data</td>
    <td>data</td>
    <td>data</td>
    <td>data</td>
    <td>data</td>
    <td>data</td>
    <td>data</td>
  </tr>
</table>
```

west270:
![maximize_example3](https://user-images.githubusercontent.com/61320354/84505543-e0ca3000-ac8b-11ea-8e48-d589ad9a091e.png)
![maximize_example4](https://user-images.githubusercontent.com/61320354/84505547-e32c8a00-ac8b-11ea-9442-964292c5cf48.png)
@TheBurchLog here are the examples with html.

TheBurchLog:
Perfect, I like the one with the table still in screen. So if you want to check that in and I will do the code review. Thank you!

PR #497, opened 2020-06-12
Added button to expand output and parameter.

west270:
Added buttons to make output or parameter larger.

Fixs #170


src/ui/src/partials/request_view.html
```diff
@@ -15,6 +15,7 @@ <h1 class="page-header">
   </span>
 </h1>
 
+
```
TheBurchLog, line 18: Remove extra white space

PR #498, opened 2020-06-12
Job events

TheBurchLog:
Changed the Job Scheduler to use Events. This was required since we moved the handlers into  threads.

Key Changes:
- We no longer wait for the request to be complete. This should prevent us from running into Threading issues when users spawn to many requests that take a long period of time to complete.
- beer_garden.application.scheduler is only modified when it sees a local JOB event or REQUEST_COMPLETED event associated with a local JOB ID.

This depends on the brewtils PR: beer-garden/brewtils#207


TheBurchLog:
Moved over to #499

PR #499, opened 2020-06-12
Updated the scheduler to use Events for updates

TheBurchLog:
Changed the Job Scheduler to use Events. This was required since we moved the handlers into  threads.

Key Changes:
- We no longer wait for the request to be complete. This should prevent us from running into Threading issues when users spawn to many requests that take a long period of time to complete.
- beer_garden.application.scheduler is only modified when it sees a local JOB event or REQUEST_COMPLETED event associated with a local JOB ID.

This depends on the brewtils PR: beer-garden/brewtils#216


TheBurchLog:
I reverted the run function back to how it was. Now it will wait for the request to be complete to hold open the JOB so it doesn't reschedule. I tested this by running the sleeper command for 30 seconds every 5 seconds. I hit the warning, so it still works.

I ran all of the unit tests locally and everything passed. We will see what happens this time in Travis.

TheBurchLog:
Just realized why it is failing. It is looking for this PR in Brewtils: https://github.com/beer-garden/brewtils/pull/216

TheBurchLog:
The Scheduler was never closing its runs because the request mapper that was handled by requests.py because we were working in two different threads.

Scheduler runs in the Main Thread and kicks off the request. Then the Plugin responds on the RabbitMQ Handler Entry Point thread so they did modify the same in memory object for `request_map`. So we just needed an Event Handler that would push back those results.

src/app/beer_garden/requests.py
```diff
@@ -735,3 +736,8 @@ def handle_event(event):
                 setattr(existing_request, field, getattr(event.payload, field))
 
             db.update(existing_request)
+
+    # Required if the main thread spawns a wait Request
```
hazmat345, line 740: process, not thread

src/app/beer_garden/requests.py
```diff
@@ -735,3 +736,8 @@ def handle_event(event):
...
+    if event.name == Events.REQUEST_COMPLETED.name:
+        if str(event.payload.id) in request_map:
+            request_map[str(event.payload.id)].set()
```
hazmat345, line 743: OK, I think I understand why we have to do this here. We also do this in the `complete_request` method, but that was based on the assumption that the same entry point (aka process) would be used to create the request and to complete it.

That assumption doesn't hold for this case, because the main process is creating the request so the request_map entry lives there. Then the HTTP entry point is handling the completion and is unable to find it.

I'm going to see if removing the old code works now that you've added this so we don't have duplicate functionality laying around.

Also, we should probably make this shared memory or something so we don't have to think about this anymore :thinking:

hazmat345:
Also, if you delete the only existing job in the UI it doesn't really work well.

hazmat345:
Hmm, also, shutting down while a job is in progress does not work well.

src/app/beer_garden/requests.py
hazmat345: Should have thought about it more. Until it's shared memory we actually need it in both.

TheBurchLog:
What are you experiencing when you deleted the last scheduled job?

When I clicked the delete button on the Job View page, it brings me back to the default scheduler page.

src/app/beer_garden/requests.py
TheBurchLog: We can make that a future task for shared memory, until then, I will keep this in here.

TheBurchLog:
@hazmat345 At COB tomorrow (6/23), I plan on merging in this request. All functionality is working as expected and we were unable to replicate the issue you had in the UI. Please let me know if you have any objections.

TheBurchLog:
@hazmat345 I am going to merge this in.

PR #500, opened 2020-06-12
Provide PYTHONPATH in beer.conf

### LINKED ISSUE #492: Plugin Defined Python Env
This request came from our GITTER conversation with Keith Ancowitz. 

In the Plugin configuration we should include the ability specify the Python environment for the plugin to be executed against. That way users can run both 2.X and 3.X plugins in the same environment. 

As the Beer Garden framework we are not responsible for managing any user defined environments. They must understand that the Beer Garden dependencies must be included in their environment at a minimum. 

To make this possible we should look at how we execute the SubProcess library when running local plugins. Then if a user provides the the ENV in the beer.conf we use that.

Example
```
NAME="echo-sleeper"
VERSION="1.0.0.dev0"
DESCRIPTION="A plugin that's annoying AND lazy."
PLUGIN_ENTRY="-m echo_sleeper"
REQUIRES=['echo', 'sleeper']
PYTHON_ENV="/PATH/TO/PYTHON"
```

TheBurchLog:
This change will allow developers to self manage the Python environment. If they do not provide a Python Path then it will default to path the Beer Garden is using.

Fixes #492

### DIFF
--- a/src/app/beer_garden/local_plugins/manager.py
+++ b/src/app/beer_garden/local_plugins/manager.py
@@ -41,6 +41,7 @@ class ConfigKeys(Enum):
     DISPLAY_NAME = 11
     METADATA = 12
     NAMESPACE = 13
+    PYTHONPATH = 14
 
 
 class PluginManager(StoppableThread):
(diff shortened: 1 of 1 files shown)


TheBurchLog:
That was an easy change. Now it is INTERPRETER_PATH

Issue #377, opened 2019-12-09
3.0 Beer-garden Breaking Changes


hazmat345:
This should track breaking changes in 3.0.

Here's how this works.

Comments to this can be used to jot down notes or ideas or things to look into (better than post-it notes smile). However, the Authoritative List of breaking changes should managed by editing this description.

Below there's a list of checkboxes, each describing something.

    When the code that addresses that something is merged into the v3 branch, check the box
    When documentation describing the something has been added to the upgrade guide, edit the line and make a strikethrough (surround with ~)

We'll know we're ready to release when all the boxes are checked and struck through!

- [x] Bartender and Brew-view consolidated into a single application
- [x] Removing the "depends on" / "requires" functionality for local plugins
- [x] Removing support for Centos 6 and Python 2
- [x] Rename `amq` config section to `mq`

hazmat345:
Version 3 has been out for a good while. Closing.

PR #501, opened 2020-06-15
Migrated the usage of AMQ to MQ.

TheBurchLog:
Migrating over the usage of AMQ to MQ. 

In response to #377


hazmat345:
I should have caught this earlier - the MQ options need to specify the old names as previous names or they won't translate over. I'll fix this.

PR #503, opened 2020-06-19
Made changes to support multiple outputs.

### LINKED ISSUE #13: Define Command Outputs
For each command, it would be nice if we were able to tell what the plugin was going to respond with

Let's say I have a plugin **`foo`** that responds to command **`bar`** with JSON:
```python
{
  "foo": "bar",
  "baz": "bat",
}
```

There is no good way to tell BEERGARDEN that the above if the expected output. If there was a way to tell BEERGARDEN that, we would suddenly know a lot more about how a plugin is operating, and whether or not what it responds with is expected or unexpected. This would enable-us to create a flow-based GUI/system.

west270:
Made changes to display multiple outputs.
Sample:
image

Dependent on PR beer-garden/brewtils#218.
Fixes #13

### DIFF
--- a/src/app/beer_garden/db/mongo/models.py
+++ b/src/app/beer_garden/db/mongo/models.py
@@ -200,6 +200,8 @@ class Command(MongoModel, Document):
     parameters = ListField(EmbeddedDocumentField("Parameter"))
     command_type = StringField(choices=BrewtilsCommand.COMMAND_TYPES, default="ACTION")
     output_type = StringField(choices=BrewtilsCommand.OUTPUT_TYPES, default="STRING")
+    output_types = ListField()
+    output_labels = ListField()
     schema = DictField()
     form = DictField()
     template = StringField()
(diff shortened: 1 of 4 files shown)


TheBurchLog:
I created a plugin to help me test the different use cases for this. I went through all of the use cases I thought of and ran into a couple issues.

- When returning a list I was getting a error that that JSON was not formatted properly
- I had issues with the following commands. 
  - nothing_defined: Doesn't display results
  - nothing_defined_list: Breaks apart results when it should not have
  - single_output_labels: Label was not applied
  - single_output_type_multi_results: Breaks apart results when it should not have
  - single_output_types_multi_results: Breaks apart results when it should not have

Plugin:
https://github.com/beer-garden/example-plugins/tree/v3/output-types-labels

hazmat345:
I'm going to mark this as needs discussion but I'm opposed to this being merged. I recommend closing and deleting this branch.

I don't believe this PR actually addresses #13, so I think the question of whether to close that issue is a separate discussion.

PR #504, opened 2020-06-23
Added env options for Parent connections

TheBurchLog:
Added env options to allow for docker-compose deployment of children. That way it is easier to stand up example instances for users.

### DIFF
--- a/docker/docker-compose/docker-compose.yml
+++ b/docker/docker-compose/docker-compose.yml
@@ -11,9 +11,11 @@ x-definitions: &env
     BG_PUBLISH_HOSTNAME: localhost
 
     # IMPORTANT!!
-    # If you are connecting to a Parent BG instance, change this the resolvable DNS entry
     # If you are connection to a Parent BG instance, toggle true
     BG_PARENT_HTTP: false
+    
+    # IMPORTANT !!
+    # If you are connecting to a Parent BG instance, change this the resolvable DNS entry
     BG_PARENT_HOST: localhost
 
 services:
--- a/docker/docker-compose/docker-compose.yml
+++ b/docker/docker-compose/docker-compose.yml
@@ -12,11 +12,11 @@ x-definitions: &env
 
     # IMPORTANT!!
     # If you are connection to a Parent BG instance, toggle true
-    BG_PARENT_HTTP: false
+    BG_PARENT_HTTP_ENABLE: false
     
     # IMPORTANT !!
     # If you are connecting to a Parent BG instance, change this the resolvable DNS entry
-    BG_PARENT_HOST: localhost
+    BG_PARENT_HTTP_HOST: localhost
 
 services:
     beer-garden:
--- a/docker/docker-compose/docker-compose.yml
+++ b/docker/docker-compose/docker-compose.yml
@@ -10,6 +10,11 @@ x-definitions: &env
     # Should be resolvable from inside and outside the docker network
     BG_PUBLISH_HOSTNAME: localhost
 
+    # IMPORTANT!!
+    # If you are connecting to a Parent BG instance, change this the resolvable DNS entry
+    # If you are connection to a Parent BG instance, toggle true
+    BG_PARENT_HTTP: false
+    BG_PARENT_HOST: localhost
 
 services:
     beer-garden:
--- a/src/app/beer_garden/config.py
+++ b/src/app/beer_garden/config.py
@@ -741,7 +741,6 @@ _PARENT_SPEC = {
                             "type": "bool",
                             "default": False,
                             "description": "Serve content using SSL",
-                            "cli_separator": "_",
                         },
                         "private_key": {
                             "type": "str",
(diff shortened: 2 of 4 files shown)


src/app/beer_garden/config.py
```diff
@@ -771,6 +771,7 @@ def _parse_args(args: Sequence[str]) -> Tuple[YapconfSpec, dict]:
                             "type": "bool",
                             "default": False,
                             "description": "Publish events to parent garden over HTTP",
+                            "alt_env_names": ["PARENT_HTTP"],
```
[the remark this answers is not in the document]
TheBurchLog: Do we need to use Alt Env Names for when we are going to be passing those variables in through the command line with a different name?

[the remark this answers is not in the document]
hazmat345: @TheBurchLog no, the alt_env_names only affect how that config is resolved from the environment. If you're worried about the CLI name for this being overly long (`--event-parent-http-enable`) we could consider moving this section out from the events section - I'm not sure it really belongs there anyway.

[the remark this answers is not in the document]
TheBurchLog: sorry, I misspoke in my response. I need those to resolve from the environment variables. That is how I am defining it in the Docker Compose file. That way it was easier for me to initialize Beer Garden.

I agree that looking at the config it is not intuitive for the Parent connect information to be under events. We should probably pull that up to a top level config. I will look at moving that up.

src/app/beer_garden/config.py
```diff
@@ -723,6 +723,95 @@ def _parse_args(args: Sequence[str]) -> Tuple[YapconfSpec, dict]:
...
+                    "alt_env_names": ["PARENT_HTTP"],
+                },
+                "ssl": {
+                    "type": "dict",
+                    "items": {
+                        "enabled": {
+                            "type": "bool",
+                            "default": False,
+                            "description": "Serve content using SSL",
+                            "cli_separator": "_",
```
hazmat345, line 745: Do we need this to be `_`?

src/app/beer_garden/config.py
```diff
@@ -723,6 +723,95 @@ def _parse_args(args: Sequence[str]) -> Tuple[YapconfSpec, dict]:
...
+    "type": "dict",
+    "items": {
+        "http": {
+            "type": "dict",
+            "items": {
+                "enable": {
+                    "type": "bool",
+                    "default": False,
+                    "description": "Publish events to parent garden over HTTP",
+                    "alt_env_names": ["PARENT_HTTP"],
```
hazmat345, line 736: I *guess* I can be ok with this. I'd prefer to just set the normal name `PARENT_HTTP_ENABLE` over adding an alt name that is the same as the parent_http config item. But if you think it makes a difference and it works then :shrug:

src/app/beer_garden/config.py
```diff
@@ -723,6 +723,95 @@ def _parse_args(args: Sequence[str]) -> Tuple[YapconfSpec, dict]:
...
+                    "type": "str",
+                    "default": "/",
+                    "description": "URL path prefix",
+                    "required": False,
+                },
+                "host": {
+                    "type": "str",
+                    "default": "0.0.0.0",
+                    "description": "Host for the HTTP Server to bind to",
+                    "alt_env_names": ["PARENT_HOST"],
```
hazmat345, line 796: This alt name should be removed - it's setting us up for problems whenever we add additional transport mechanisms. If you need to set this from compose just use the normal name `PARENT_HTTP_HOST`.

src/app/beer_garden/config.py
```diff
@@ -733,7 +733,7 @@ def _parse_args(args: Sequence[str]) -> Tuple[YapconfSpec, dict]:
                     "type": "bool",
                     "default": False,
                     "description": "Publish events to parent garden over HTTP",
-                    "alt_env_names": ["PARENT_HTTP"],
+                    "alt_env_names": ["PARENT_HTTP_ENABLE"],
```
hazmat345: We don't need to specify this here - "PARENT_HTTP_ENABLE" *is* the environment name.

src/app/beer_garden/config.py
```diff
@@ -793,7 +793,7 @@ def _parse_args(args: Sequence[str]) -> Tuple[YapconfSpec, dict]:
                     "type": "str",
                     "default": "0.0.0.0",
                     "description": "Host for the HTTP Server to bind to",
-                    "alt_env_names": ["PARENT_HOST"],
+                    "alt_env_names": ["PARENT_HTTP_HOST"],
```
hazmat345: Same with this - "PARENT_HTTP_HOST" is already the environment name for this config item.

docker/docker-compose/docker-compose.yml
```diff
@@ -10,6 +10,13 @@ x-definitions: &env
     # Should be resolvable from inside and outside the docker network
     BG_PUBLISH_HOSTNAME: localhost
 
+    # IMPORTANT!!
+    # If you are connection to a Parent BG instance, toggle true
+    BG_PARENT_HTTP: false
+    
+    # IMPORTANT !!
+    # If you are connecting to a Parent BG instance, change this the resolvable DNS entry
+    BG_PARENT_HOST: localhost
```
hazmat345, line 19: These need to change - I'll change them.

PR #507, opened 2020-07-08
Multi plugins

hazmat345:
This PR fixes issues associated with multi-instance plugins.

### DIFF
--- a/src/app/beer_garden/systems.py
+++ b/src/app/beer_garden/systems.py
@@ -118,7 +118,7 @@ def update_system(
         system = db.replace_commands(system, new_commands)
 
     if add_instances:
-        if len(system.instances) + len(add_instances) >= system.max_instances:
+        if len(system.instances) + len(add_instances) > system.max_instances:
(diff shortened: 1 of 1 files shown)


src/app/beer_garden/systems.py
```diff
@@ -118,13 +118,13 @@ def update_system(
...
     if add_instances:
-        if len(system.instances) + len(add_instances) >= system.max_instances:
+        if len(system.instances) + len(add_instances) > system.max_instances:
             raise ModelValidationError(
                 f"Unable to add instance(s) to {system} - would exceed "
                 f"the system instance limit of {system.max_instances}"
             )
 
-        system.instances.append(add_instances)
+        system.instances += add_instances
```
TheBurchLog, line 127: I guess this works instead of append, just a little different then what we normally do.

hazmat345: `add_instances` is a list, if you use append you end up with `[instance1, [instance2]]`. Using concatenation is what we need.

PR #520, opened 2020-07-22
Added handler for unresponsive child gardens

### LINKED ISSUE #513: Request forwarding fails silently
If you have an incorrectly-configured connection to a child garden (wrong port, for example) then attempts to publish a request to that child will silently fail. The request will exist in a `CREATED` state but will give no indication that anything is wrong.

TheBurchLog:
When a failure occurs when publishing to a down stream Garden, we can now throw an event that can be handled by correctly for each Payload Type. In this commit we are handling the Garden Status change and Request Creates.

Fixes #513 

Requires https://github.com/beer-garden/brewtils/pull/220


hazmat345:
@TheBurchLog OK, I figured out why none of the Tornado logs were showing up, at least. See 92bfad3a.

I'd recommend rebasing this onto v3 for debugging purposes.

TheBurchLog:
@hazmat345 
I see what is happening. I only handled what happens when the connection fails. Not what happens when the Admin didnt' configure the connection.

TheBurchLog:
@hazmat345 

Try it out now. I did the following use cases

1. New BG connected and connection info not set.
1. New BG connected and connection info set properly
1. New BG connected and connection info set properly, then changed to wrong connection info

TheBurchLog:
@hazmat345 
Can you run through the test criteria again and let me know if you are still seeing the issue you documented?

TheBurchLog:
@hazmat345 Updated the Status and Events for the Gardens to reflect the different scenarios we could be in.

hazmat345:
@TheBurchLog this needs to be rebased onto v3 to test... I tried to do this and made issue/513 branches in both beer-garden and brewtils. Check them out and see if they look reasonable, but feel free to ignore them :smile: 

I did test in v3 to see if this is still a problem and it is.

src/app/beer_garden/requests.py
```diff
@@ -757,6 +757,23 @@ def handle_event(event):
...
+                Events.GARDEN_NOT_CONFIGURED.name,
+            )
+            and event.payload.model_type == "Request"
+        ):
+            complete_request(
+                event.payload.model.id,
+                status="ERROR",
+                error_class=event.error_message,
+            )
+
```
hazmat345, line 776: I really don't like doing this in the garden module. Can we just call `complete_request` directly in the error handling of the `forward` function?

src/app/beer_garden/requests.py
```diff
@@ -757,6 +757,23 @@ def handle_event(event):
...
+                Events.GARDEN_UNREACHABLE.name,
+                Events.GARDEN_ERROR.name,
+                Events.GARDEN_NOT_CONFIGURED.name,
+            )
+            and event.payload.model_type == "Request"
+        ):
+            complete_request(
+                event.payload.model.id,
+                status="ERROR",
+                error_class=event.error_message,
```
hazmat345, line 774: The error class is supposed to be a literal class name. If you want to make the error message part of the request you should put it in the output, but make sure you look at how brewtils does that. There should be a format function that you need to call to account for if the output should be JSON or not.

src/app/beer_garden/router.py
```diff
@@ -443,35 +455,69 @@ def _determine_target_garden(operation: Operation) -> str:
...
+            Event(
+                name=Events.GARDEN_ERROR.name,
+                payload_type=operation.schema,
+                payload=operation,
+                error_message=f"Attempted to forward operation to garden "
+                f"'{operation.target_garden_name}' but an error occurred."
+                f"Please talk to your system administrator.",
+            )
         )
+        raise ex
```
hazmat345, line 520: If you're re-raising an exception like this the statement needs to be just `raise`, not `raise ex`. The way it is now is going to be reported in the stacktraces as the exception handling for the first exception encountering a problem and raising a *new* exception (that happens to be identical to the first one). That's how we get these long chained exceptions with "error encountered while processing exception..." in the logs.

hazmat345 (changes requested):
Ok, this definitely fixes the problem. Instead of a request that hangs forever in CREATED the request errors. But there are a couple issues to be addressed.

hazmat345 (approved):
This looks good, only thing is that the output should be run through the same code path as plugin-generated errors in order to match what the plugins do (this prevents JSON parse errors for system clients that are expecting JSON).

However, I'm not sure that's still what we should be doing, so I'm going to open a new issue to address.

PR #521, opened 2020-07-22
Nameless Pour Fix

### LINKED ISSUE #515: Namespace-less requests Pour It Again
Upgrading from v2 to v3 will result in existing requests not having a namespace. This breaks the Pour it Again functionality. We need to determine what to do / how to handle.

TheBurchLog:
When we have people migrate over from v2 this is to ensure that their old requests will still work. The Pour Again feature will grab the local garden name for the Namespace. If that doesn't work, then we won't guess which plugin they actually ran.

I would hope that we could do some database migration to fix this, but this is our fallback. 

Fixes #515

### DIFF
--- a/src/ui/src/js/controllers/request_view.js
+++ b/src/ui/src/js/controllers/request_view.js
@@ -11,6 +11,7 @@ requestViewController.$inject = [
   'RequestService',
   'SystemService',
   'EventService',
+  'GardenService',
 ];
 
 /**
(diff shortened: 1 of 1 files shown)


hazmat345:
@TheBurchLog I think this is overkill - see #536.

I definitely don't think we should be modifying the v2 requests in the database to fix this.

hazmat345:
Going to close this as it's OBE after #536

PR #522, opened 2020-07-23
Application will now wait for all Entry Points to start

### LINKED ISSUE #518: Multiple entry point startup procedure
Right now we're being kind of lazy about how the local plugin startup process is kicked off. We wait for the entry point started event to be handled by the app module and immediately start the local plugins.

We should instead track the status of *all* entry point startups and only start the local plugins when all are ready.

We should probably also publish the garden started event at that point as well, rather than waiting until the very end of the application startup.

TheBurchLog:
When Beer Garden starts up, it will wait for all Entry Points to be in the RUNNING status before proceeding. This is done by providing each entry point two call back functions, first monitors events for START/STOP events to reflect in it's status. The second is the standard callback to provide events to the main process.

Fixes #518

### DIFF
--- a/src/app/beer_garden/api/entry_point.py
+++ b/src/app/beer_garden/api/entry_point.py
@@ -74,19 +74,20 @@ class EntryPoint:
         self._logger = logging.getLogger(__name__)
         self._process = None
         self._ep_conn, self._mp_conn = Pipe()
+        self.event_callback = event_callback
 
-        # We use two action events. The first one executes to manage START/STOP events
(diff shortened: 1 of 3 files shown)


TheBurchLog:
The Entry Point will evaluate each Event it publishes before forwarding it along. This allows for the Event Start and Stop to change the status of the Entry Point.

hazmat345 (commented):
Converting this to a draft - this depends on #529 so let's revisit after that's merged.

hazmat345:
This doesn't apply cleanly anymore 😢

PR #523, opened 2020-07-23
Flex logging

hazmat345:
This PR reworks how plugin logging configuration works. Depends on / enables beer-garden/brewtils#221.

### DIFF
--- a/src/app/beer_garden/api/http/handlers/v1/logging.py
+++ b/src/app/beer_garden/api/http/handlers/v1/logging.py
@@ -6,11 +6,34 @@ from brewtils.schema_parser import SchemaParser
 from beer_garden.api.http.base_handler import BaseHandler
 
 
+class LoggingAPI(BaseHandler):
+    async def get(self):
+        """
+        ---
+        summary: Get plugin logging configuration
+        description: |
+          Will return a Python logging configuration that can be used to configure
+          plugin logging.
+        responses:
+          200:
+            description: Logging Configuration for system
+          50x:
+            $ref: '#/definitions/50xError'
+        tags:
+          - Logging
+        """
+        response = await self.client(Operation(operation_type="PLUGIN_LOG_READ"))
+
+        self.set_header("Content-Type", "application/json; charset=UTF-8")
+        self.write(response)
+
+
 class LoggingConfigAPI(BaseHandler):
     async def get(self):
         """
         ---
         summary: Get the plugin logging configuration
+        deprecated: true
         parameters:
           - name: system_name
             in: query
@@ -25,9 +48,9 @@ class LoggingConfigAPI(BaseHandler):
           50x:
             $ref: '#/definitions/50xError'
         tags:
-          - Config
+          - Deprecated
         """
-        response = await self.client(Operation(operation_type="LOG_READ"))
+        response = await self.client(Operation(operation_type="PLUGIN_LOG_READ_LEGACY"))
 
         self.set_header("Content-Type", "application/json; charset=UTF-8")
         self.write(response)
@@ -36,6 +59,7 @@ class LoggingConfigAPI(BaseHandler):
         """
         ---
         summary: Reload the plugin logging configuration
+        deprecated: true
         description: |
           The body of the request needs to contain a set of instructions detailing the
           operation to make. Currently supported operations are below:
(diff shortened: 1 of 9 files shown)


src/app/example_configs/plugin-logging.yaml
```diff
@@ -1,30 +1,30 @@
...
+#  plugin logging
+#  file:
+#    backupCount: 5
+#    class: "logging.handlers.RotatingFileHandler"
+#    encoding: utf8
+#    formatter: default
+#    level: INFO
+#    maxBytes: 10485760
+#    filename: "logs/$instance_name.log"
+
```
TheBurchLog, line 20: Just wanted to make that we document this. The filename is in relation to where the plugin is executed. So just kicked off a remote plugin and it built it in the directory I was running in. So if someone wants to use this for local and remote, both directory locations must be available.

src/app/beer_garden/log.py
```diff
@@ -345,18 +365,17 @@ def _validate_formatters(formatters):
             return {"default": {"format": LoggingConfig.DEFAULT_FORMAT}}
 
         for formatter_name, _formatter_info in six.iteritems(formatters):
+            # TODO - WTF
```
TheBurchLog, line 368: What needs to be done here?

src/app/beer_garden/log.py
```diff
@@ -111,16 +111,23 @@ def setup_entry_point_logging(queue):
...
 
 
-def get_plugin_log_config(system_name=None):
-    return _PLUGIN_LOGGING.get_plugin_log_config(system_name=system_name)
+def get_plugin_log_config(**_) -> dict:
+    """Get the plugin logging configuration
 
+    Args:
+        **_: Eventually you will be able to select a plugin logging config based on
+            selectors (system name, etc.)
```
TheBurchLog, line 119: When we chatted about this, part of me was thinking what would happen if we added the logging configuration to the System model? Then when a System is created the plugin would pass the flag is a local config is being used or not. If a local config was not being used, then we could grab the default config and populate it.

This would allow us to later be able to create an admin interface that could change the logging configuration. Just kick off an Admin command and then update the model. Then we would have to determine if a server "forced" logging change overwrites locally provided configs and if they would persist on reboot.

This is not an ask for this PR, just wanted to put my thoughts some where.

src/app/example_configs/plugin-logging.yaml
hazmat345: Yeah, there's really no way around that, unfortunately. We need a way to modify or use different configurations based on the plugin. Which was the original goal of this, but we're not there yet.

src/app/beer_garden/log.py
hazmat345: :laughing: 

It needs to be ripped out.

src/app/beer_garden/log.py
hazmat345: Hmm, that's a really interesting idea. We'd probably need some sort of granularity at the Instance level too, but I like this idea.

hazmat345:
@TheBurchLog Ok, this is ready for another look.

I ripped out all the old crap, but I wanted to make sure it wouldn't break for anybody using it since it's exposed API.

To get around this I essentially mangle the config to make it look like the old style. I deprecated the current API that allows the plugins to get that data (/api/v1/config/logging) and made a new one to get the new format (/ap1/v1/logging).

src/app/beer_garden/api/http/handlers/v1/logging.py
```diff
@@ -71,7 +91,9 @@ async def patch(self):
         response = None
         for op in patch:
             if op.operation == "reload":
-                response = await self.client(Operation(operation_type="LOG_RELOAD"))
+                response = await self.client(
+                    Operation(operation_type="PLUGIN_LOG_RELOAD")
+                )
```
TheBurchLog, line 96: Where do we invoke this API command? Because I can't find reference to it. They way I am reading this command is that if we change the plugin logging, if we call this endpoint then all local plugins get this update. But it does not work the same for remote plugins. They won't get it until restart.

hazmat345: @TheBurchLog there's definitely room for improvement here. This reload operation probably shouldn't even exist - all it does is redo what the application does on startup. It doesn't send any updates to plugins, it only updates the configuration that will be returned if any new plugins ask.

Basically, if you want to change the configuration you'll need to update the file and then hit this endpoint (and we currently don't expose any way to hit this endpoint from the UI). After doing that any new plugins that come up will get the updated config.

We should probably just watch the plugin logging config file and execute this operation if it ever changes. If we do that then this API doesn't need to exist.

PR #528, opened 2020-07-29
Added Namespace to System Admin Page

### LINKED ISSUE #527: No concept of Namespace on System Admin Page
On the System Admin page we need a way to display the namespace that a plugin is deployed on. Currently you don't know which one you are going to delete.

TheBurchLog:
On the admin page we needed a way to display the namespace. This just adds it to the hyperlink.

fixes #527


west270:
Looks good to go.

PR #529, opened 2020-07-29
Adds Stomp Entry-point

west270:
Adds new entry-point for stomp communication. The stomp library used is https://github.com/jasonrbriggs/stomp.py and the docker image used is https://hub.docker.com/r/vromero/activemq-artemis.

The entry-point sends events to an events topic called beergarden/events and receives messages from topic beergarden/operations. When beergarden receives a message from stomp it sends the operation result to the events topic or to reply-to destination defined in received header.

Can start stomp communication with ```docker-compose up -d activemq``` and access management console from http://0.0.0.0:8161/console. The default username and password  are username: beer_garden password: password


src/app/beer_garden/api/stomp/server.py
```diff
@@ -0,0 +1,65 @@
...
+
+class Connection:
+    def __init__(self, hosts=None):
+        if hosts is None:
+            hosts = [('localhost', 61613)]
+        global conn
+        conn = stomp.Connection(host_and_ports=hosts, heartbeats=(10000, 0))
+        conn.set_listener('', OperationListener())
+        conn.connect(username='beer_garden', passcode='password', wait=True, headers={'client-id': 'BeerGarden'})
+        conn.subscribe(destination='beergarden/operations', id='beer_garden', ack='auto',
```
TheBurchLog, line 47: We do not hard code in username/passwords. You pass those in through the config file.

src/app/beer_garden/api/stomp/server.py
```diff
@@ -0,0 +1,65 @@
...
+            if 'reply-to' in headers:
+                conn.send(body=result, headers=response_headers, destination=headers['reply-to'])
+            else:
+                conn.send(body=result, headers=response_headers, destination='beergarden/events')
+
+
+class Connection:
+    def __init__(self, hosts=None):
+        if hosts is None:
+            hosts = [('localhost', 61613)]
```
TheBurchLog, line 42: should this come from the config?

src/app/example_configs/config.yaml
```diff
@@ -36,6 +36,8 @@ entry:
       private_key: null
       public_key: null
     url_prefix: /
+  stomp:
+    enable: true
   thrift:
```
TheBurchLog, line 41: Here is where we need to have the host name, port, user name, password, etc.

src/app/beer_garden/api/stomp/server.py
```diff
@@ -0,0 +1,65 @@
...
+                operation.kwargs['wait_timeout'] = 0
+        result = beer_garden.router.route(operation)
+        if result is not None:
+            many = isinstance(result, list)
+            model_class = result.__class__.__name__
+
+            if many:
+                model_class = result[0].__class__.__name__
+            response_headers = {'model_class': model_class, 'many': many}
+            response_headers = append_headers(headers=response_headers)
```
TheBurchLog, line 31: Can this have two inputs? 
```python 
response_headers = append_headers(response_headers=response_headers, request_headers=headers)
```

Then in the function we just don't do anything with the request_headers.

src/app/beer_garden/api/stomp/__init__.py
```diff
@@ -0,0 +1,85 @@
...
+
+io_loop = None
+
+
+def run(ep_conn):
+    global logger, conn, client, io_loop
+    logger = logging.getLogger(__name__)
+    logger.info("Starting Stomp server")
+    hosts = [('localhost', 61613)]
+    conn = Connection(hosts=hosts)
```
TheBurchLog, line 29: These variables should be getting passed in from the config

src/app/example_configs/config.yaml
```diff
@@ -36,6 +36,8 @@ entry:
       private_key: null
       public_key: null
     url_prefix: /
+  stomp:
+    enable: true
```
TheBurchLog, line 40: This needs to be disabled by default

src/app/requirements.txt
```diff
@@ -1,3 +1,4 @@
+stomp
```
TheBurchLog, line 1: I don't like the placement of this. Let's get a version locked in for this and move it down with the others.

TheBurchLog: This doesn't work. You need to use `stomp.py` not `stomp`

src/app/beer_garden/api/stomp/server.py
```diff
@@ -0,0 +1,65 @@
...
+    def on_message(self, headers, message):
+        global conn
+        operation = SchemaParser.parse_operation(
+            message, from_string=True
+        )
+        if hasattr(operation, 'kwargs'):
+            if 'wait_timeout' in operation.kwargs and operation.operation_type == "REQUEST_CREATE":
+                operation.kwargs['wait_timeout'] = 0
+        result = beer_garden.router.route(operation)
+        if result is not None:
```
TheBurchLog, line 24: You need a way to handle exception and decide what to do when we throw an exception. Do we want to publish that out or just leave it in the logs?

src/app/beer_garden/api/stomp/server.py
```diff
@@ -0,0 +1,150 @@
...
+
+        result = None
+        if error_msg is None:
+            try:
+                result = beer_garden.router.route(operation)
+            except:
+                error_msg = "Failed to route operation"
+                error = sys.exc_info()
+        if error is not None:
+            send_error_msg(error, headers)
```
TheBurchLog, line 83: You never use `error_msg`, are you meaning to forward that message to the user?

src/app/beer_garden/api/stomp/server.py
```diff
@@ -0,0 +1,150 @@
...
+                body=error.__str__(),
+                headers=error_headers,
+                destination=stomp_config.event_destination,
+            )
+    pass
+
+
+class OperationListener(stomp.ConnectionListener):
+    def on_error(self, headers, message):
+        print("received an error:", headers)
```
TheBurchLog, line 61: We do not use Print statements in our baseline. You should use the logger like you did in the Connection class lower in the file.

src/app/beer_garden/api/stomp/server.py
```diff
@@ -0,0 +1,150 @@
...
+            error_msg = "Failed to parse message"
+            error = sys.exc_info()
+
+        result = None
+        if error_msg is None:
+            try:
+                result = beer_garden.router.route(operation)
+            except:
+                error_msg = "Failed to route operation"
+                error = sys.exc_info()
```
TheBurchLog, line 81: We have a bunch of different exceptions that couple be raised during this. We should properly reflect which exception was raised.We raise the exception So you can just capture the error message and return the message.

``` 
except Exception as e:
    error_msg = str(e)
```

src/app/beer_garden/api/stomp/processors.py
```diff
@@ -0,0 +1,29 @@
+from brewtils.models import Event, Events, Request, Operation, System
```
TheBurchLog, line 1: Not used

TheBurchLog (commented):
Here are my comments before running the code.

TheBurchLog:
I have no future comments on this PR.

@hazmat345 Do you have any comments?

src/app/beer_garden/api/stomp/server.py
```diff
@@ -0,0 +1,136 @@
...
+                operation.kwargs.pop("wait_timeout", None)
+            result = beer_garden.router.route(operation)
+            send_message(result, headers)
+        except Exception as e:
+            send_error_msg(str(e), headers)
+
+
+class Connection:
+    @staticmethod
+    def __init__():
```
hazmat345, line 76: Having the `__init__` method be static is kind of odd. I'm actually surprised that python doesn't yell about this... was this just a testing thing or is there a specific need driving this?

src/app/beer_garden/api/stomp/server.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def on_message(self, headers, message):
+        try:
+            operation = SchemaParser.parse_operation(message, from_string=True)
+            if hasattr(operation, "kwargs"):
+                operation.kwargs.pop("wait_timeout", None)
+            result = beer_garden.router.route(operation)
+            send_message(result, headers)
+        except Exception as e:
+            send_error_msg(str(e), headers)
```
hazmat345, line 71: We need to do some logging here. When I was testing I attempted to create a request for a system version that didn't exist and there was no indication in the application logs that anything happened.

src/app/beer_garden/api/stomp/server.py
```diff
@@ -0,0 +1,136 @@
...
+                body=error_msg,
+                headers=error_headers,
+                destination=stomp_config.event_destination,
+            )
+    pass
+
+
+class OperationListener(stomp.ConnectionListener):
+    def on_error(self, headers, message):
+        logger = logging.getLogger(__name__)
```
hazmat345, line 60: This logger should probably be a module-level thing (on the same level as `conn` and `bg_active`).

src/app/beer_garden/api/stomp/server.py
```diff
@@ -0,0 +1,136 @@
...
+            conn.set_ssl(
+                for_hosts=host_and_ports,
+                key_file=stomp_config.private_key,
+                cert_file=stomp_config.cert_file,
+            )
+        conn.set_listener("", OperationListener())
+
+    @staticmethod
+    def connect(connected_message=None):
+        global conn
```
hazmat345, line 92: `global` statements are only needed if you're *assigning* to that variable inside of the function. If you're just accessing / using the value they're not needed.

It's because without them if you tried to assign something to the name `conn` inside the function Python will create a normal function-scoped name for you. The `global` tells Python to not do that and instead assign to the "global" namespace.

If you're just accessing the name you don't have to worry about this because the normal name lookup will look for the name in the function scope, fail to find it there, look in the module scope, and find it.

src/app/beer_garden/api/stomp/__init__.py
```diff
@@ -0,0 +1,96 @@
...
+
+from beer_garden.api.stomp.processors import EventManager
+from beer_garden.api.stomp.server import Connection
+from beer_garden.events import publish
+from beer_garden.events.processors import QueueListener
+
+logger = None
+conn = None
+
+io_loop = None
```
hazmat345, line 20: The `io_loop` code shouldn't be necessary, and it may actually be causing some performance losses. `stomp.py` is not an asyncio thing, so there's no need to start up an event loop.

This should make things simpler than the HTTP entry point - everywhere that does something like `io_loop.add_callback` should just be able to call the function directly. For example, in the signal handler

```python
io_loop.add_callback_from_signal(shutdown)
```

can be 

```python
shutdown()
```

hazmat345 (changes requested):
I think a couple more things are needed here, but I think this is getting very close. I'm pretty confidant this is good for 3.0.

src/app/beer_garden/api/stomp/__init__.py
west270: I removed the `io_loop` and created a thread to get events from `ep_conn`

src/app/beer_garden/api/stomp/server.py
west270: python was complaining about it not being static because I was not using `self`

hazmat345 (approved):
Sweet, this looks good. :fireworks::fireworks::fireworks:

BUT :smile:

Important: Please add `stomp.py` to the requirements.txt and setup.py files. It's already in requirements.in, which is good, but it needs to be in all three.

Less important: I know that Travis has been misbehaving lately so you may not have seen this, but there's a linting error :smile:

Going forward there's some organizational changes that I think should be made with regards to the `Connection` class, but I'm OK with accepting the PR and opening an issue for them.

hazmat345:
@west270 the versions in requirements.txt and setup.py are mutually exclusive.

PR #537, opened 2020-08-04
Adding content-type for JS output

hazmat345:
Adds the correct content-type for JS output.


TheBurchLog:
@hazmat345 Can you add this change to the request_view.js? Right now it won't populate the output field because we are doing a hard filter on output types.

hazmat345:
@TheBurchLog are you sure? It just shows up as normal text for me, which is what I would expect. I didn't make any effort to prettify it since the only use-case I see for this is for "helper" commands that shouldn't be called from the UI anyway.

TheBurchLog:
I didn't actually write a plugin to test this.... so I was just assuming it would fail.... sorry about that. We shouldn't make a helper class to pretty it. I see this PR as a go now :smile:

hazmat345:
I'm going to close this in favor of #539

PR #541, opened 2020-08-06
Added the ability to cache the results of Requests

TheBurchLog:
This implements the ability to ask for the results of a Request to be cached for X time period. If the record is already cached, then the cached instance will be returned. If the record is not cached, then it will be added to the cache for the determined time period.

Options:
Cache Time = -1 (Don't use Cached results)
Cache Time = 0 (Use cached results, but don't cache this)
Cache Time > 0 (Cache results for this many seconds)

Depends on: https://github.com/beer-garden/brewtils/pull/225

### DIFF
--- a/src/app/beer_garden/api/http/handlers/v1/request.py
+++ b/src/app/beer_garden/api/http/handlers/v1/request.py
@@ -384,6 +384,12 @@ class RequestListAPI(BaseHandler):
             description: Max seconds to wait for request completion. (-1 = wait forever)
             type: float
             default: -1
+          - name: cache
+            in: query
+            required: false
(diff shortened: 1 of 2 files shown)


hazmat345 (commented):
We need to think long and hard whether this is something we want to do. Converting to draft awaiting further discussion.

hazmat345:
I'm not comfortable adding something like this without **extremely** extensive testing in place to support it. Additionally, a change this significant needs an associated issue for discussion.

Since this isn't currently on the roadmap and nobody is asking for it I don't think we get value from keeping this around - I'm going to close and remove.

PR #546, opened 2020-08-07
Reordering beer.conf processing steps

### LINKED ISSUE #542: beer.conf $BG_PLUGIN_PATH not working
In a beer.conf environment specification `$BG_PLUGIN_PATH` should be interpreted... it's currently not working.

hazmat345:
Fixes #542 - Needed to ensure all entries in the env dict are strings before expanding `$VAR` items.

### DIFF
--- a/src/app/beer_garden/local_plugins/env_help.py
+++ b/src/app/beer_garden/local_plugins/env_help.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 import os
+from typing import Dict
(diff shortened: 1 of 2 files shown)


hazmat345:
The substitution was failing because it was trying to replace `$BG_PLUGIN_PATH` with a non-string value (a PosixPath in this case). I was going to go through and ensure that all values were strings before calling that function, but then I realized that we were already doing that, just not early enough.

PR #547, opened 2020-08-07
Issue/544

### LINKED ISSUE #544: Request view output pane size persistent value
We should store the state of the request view page panels in browser local storage so they persist between requests. This will prevent a user from needing to expand the `output` pane every time they make a request.

hazmat345:
This fixes #544 by persisting values in local storage.

### DIFF
--- a/src/ui/src/js/controllers/request_view.js
+++ b/src/ui/src/js/controllers/request_view.js
@@ -77,12 +77,13 @@ export default function requestViewController(
     return RequestService.isComplete(request);
   };
   
-  $scope.Resize = function(resizeCell) {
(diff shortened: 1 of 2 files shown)


TheBurchLog:
Using local storage opens up a couple options with the UI that could consider. For example, keeping state on the Theme the user has selected. In this PR could you also add this ability?

A more advance would be allowing users to favorite commands. We should chat about this feature and discuss if it is a good idea, and if so, make a new PR for it.

hazmat345:
@TheBurchLog not sure what you mean by theme tracking, do you mean loading the theme when the user accesses the page?

Sounds like a new issue to me :smile:

PR #550, opened 2020-08-10
Added ability to search commands along with Systems

### LINKED ISSUE #543: UI Search Bar Commands
The overall search bar on the UI should find commands in addition to systems.

TheBurchLog:
Updated the top search bar to search commands as well as systems.

Fixes #543


TheBurchLog:
@hazmat345 
I fixed the command link and it now redirects properly. Now the dropdown will not populate unless you have values in the search bar.

I could not get the commands to look the same as systems. So instead I restructured the output to make it look like they build on each other.

PR #554, opened 2020-08-12
Removing Admin Queue Page

### LINKED ISSUE #533: Consolidate Systems and Queues admin pages
We should consolidate the queue admin page functionality into the system admin page:

- See how many messages are on a specific instance's queue*
- Clear a specific instance's queue
- Clear all queues

* We should make this functionality require an affirmative action rather than just passively being update every couple of seconds. Getting the message counts for all queues is a resource-intensive operation for the rabbit broker, so we shouldn't be doing it all the time for every user on the admin page.

TheBurchLog:
Moved all of the features in the Admin Queue page over to the Admin System page. The page no longer polls every 5 seconds for updates. When having the Queue prompt up, it will poll every 30 seconds for that one queue. This required a new endpoint to pull based off the instance. Could have done this off the Queue Name, but I went with this instead.

Also, I dropped the Admin Queue page all together since we have this functionality else where.

Fixes #533


src/app/beer_garden/api/http/__init__.py
```diff
@@ -181,6 +181,7 @@ def _setup_tornado_app():
         (rf"{prefix}api/v1/commands/(\w+)/?", v1.command.CommandAPI),
         (rf"{prefix}api/v1/instances/(\w+)/?", v1.instance.InstanceAPI),
         (rf"{prefix}api/v1/instances/(\w+)/logs/?", v1.instance.InstanceLogAPI),
+        (rf"{prefix}api/v1/instances/(\w+)/queues/?", v1.instance.InstanceQueuesAPI),
```
hazmat345, line 184: :+1:

src/app/beer_garden/api/http/handlers/v1/instance.py
```diff
@@ -254,3 +254,36 @@ async def get(self, instance_id):
...
+
+
+class InstanceQueuesAPI(BaseHandler):
+    @authenticated(permissions=[Permissions.QUEUE_READ])
+    async def get(self, instance_id):
+        """
+        ---
+        summary: Retrieve all queue information
+        parameters:
+          - name: queue_name
```
hazmat345, line 266: Think this should be instance_id

src/app/beer_garden/api/http/handlers/v1/instance.py
```diff
@@ -254,3 +254,36 @@ async def get(self, instance_id):
...
+            in: path
+            required: false
+            description: The instance ID to pull queues for
+            type: string
+        responses:
+          200:
+            description: List of all queue information objects
```
hazmat345, line 273: "List of queue info objects for this instance"

TheBurchLog:
@hazmat345 
I got all of the changes you asked for. The wait time is now ten seconds. Took my a hot minute to figure out the mouse over icon, but that is now resolved. I am going to merge this in

PR #558, opened 2020-08-19
Persist Themes (No Login Required)

TheBurchLog:
Now when a user selects a Theme from the drop down menu the selection will persist when refreshing the page. This previously only worked for logged in users. Now any person who accesses BG can make this selection and it will persist that to local storage.

(ignore commit message)

### DIFF
--- a/src/ui/src/js/run.js
+++ b/src/ui/src/js/run.js
@@ -91,8 +91,11 @@ export default function appRun(
         // coalescePermissions [0] is roles, [1] is permissions
         user.permissions = RoleService.coalescePermissions(user.roles)[1];
 
-        let theme = _.get(user, 'preferences.theme', 'default');
-        $rootScope.changeTheme(theme);
+        // If user is logged in, change to their default theme selection
+        if (user.id != null){
+          let theme = _.get(user, 'preferences.theme', 'default');
+          $rootScope.changeTheme(theme);
+        }
 
         $rootScope.user = user;
       }, (response) => {
@@ -250,6 +253,12 @@ export default function appRun(
       updateInstance(event.payload);
     }
   });
+
+  // If a Theme is stored, switched to Theme in the session
+  let theme = localStorageService.get('currentTheme');
+  if (theme != null){
+    $rootScope.changeTheme(theme);
+  }
 
   $interval(function() {
     EventService.connect(TokenService.getToken());
--- a/src/ui/src/js/run.js
+++ b/src/ui/src/js/run.js
@@ -92,10 +92,13 @@ export default function appRun(
         user.permissions = RoleService.coalescePermissions(user.roles)[1];
 
         // If user is logged in, change to their default theme selection
-        if (user.id != null){
-          let theme = _.get(user, 'preferences.theme', 'default');
-          $rootScope.changeTheme(theme);
+        let theme;
+        if (user.id) {
+          theme = _.get(user, 'preferences.theme', 'default');
+        } else {
+          theme = localStorageService.get('currentTheme') || 'default';
         }
+        $rootScope.changeTheme(theme, false);
 
         $rootScope.user = user;
       }, (response) => {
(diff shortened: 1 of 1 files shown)


src/ui/src/js/run.js
```diff
@@ -87,8 +87,11 @@ export default function appRun(
         // coalescePermissions [0] is roles, [1] is permissions
         user.permissions = RoleService.coalescePermissions(user.roles)[1];
 
-        let theme = _.get(user, 'preferences.theme', 'default');
-        $rootScope.changeTheme(theme);
+        // If user is logged in, change to their default theme selection
+        if (user.id != null){
```
hazmat345, line 91: Is this to exclude the anonymous user?

TheBurchLog: Yes, if the user is anonymous then use what is cached in their browser storage. If not, then use what we store with that user account.

TheBurchLog:
@hazmat345 
I fixed the message. Didn't realize I had to do a force push on it.

Documenting for future me:
```
git commit --amend -m "<NEW MESSAGE>"
git push --force
```

hazmat345:
@TheBurchLog this doesn't seem to work like I'd expect. If I set the theme to slate and then close and reopen the tab it just uses the default theme.

TheBurchLog:
@hazmat345 I am getting the same results. I have the same issue when I try maximizing the output field for Requests. It looks like Storage is not persisting to cookies like we are expecting.

TheBurchLog:
@hazmat345 Found the issue. We are storing these values based on the Session. So it will not persist across tabs.

TheBurchLog:
To fix this we need to update the index.js to the following
```Javascript
.config(['localStorageServiceProvider', function(localStorageServiceProvider) {
  localStorageServiceProvider.setStorageType('localStorage');
}])
```

TheBurchLog:
@hazmat345 The main question we should be asking is if we want Tokens to be Session or Browser based.

If the tokens are changed to be Browser based, then the user can open multiple tabs with the same authentication token. This might be a problem when they refresh because each session will request a refresh and could cause them to become out of sync. So we would need something to manage that. 

We could make two different storage solutions, Session and Local. Then anything like browser preferences we store in the cookies and tokens we store in the session?

TheBurchLog:
I found a solution for this. We can directly interact with the cookies instead of the storage solution. 
Docs:
https://www.npmjs.com/package/angular-local-storage/v/0.1.2#cookie

hazmat345:
@TheBurchLog the problem with cookies is that they're going to be sent along with every request. That's why we moved away from them.

Is there a reason why we can't use both types of storage?

TheBurchLog:
@hazmat345 This is utilizing both solutions. Here are the two ways we interact with local storage now

``` python
# Example 1 
# Using the local storage object as configured (Session Based)
    localStorageService.set('key', value);
    let value = localStorageService.get('key');

# Example 2
# Going directly to the Cookies instead of using the Local Storage Settings
    localStorageService.cookie.set('key', value);
    let value = localStorageService.cookie.get('key');
```

TheBurchLog:
@hazmat345 I wasn't aware that cookies get sent forward with REST commands. That is good to know.

As it stands the `localStorageService` that was are using is limited to one or the other. We would have to find a way to inject a second storage service in our baseline.

hazmat345:
@TheBurchLog from the angular-local-storage docs:

>You can also dynamically change storage type by passing the storage type as the last parameter for any of the API calls. For example: `localStorageService.set(key, val, "sessionStorage");`

I'm good with making the default localStorage since that seems like it'll be the majority of what we need. Then we can selectively use session when we need to.

I'd also be OK with just removing the login capability from the UI for now until it's in a better state, but that's a discussion for another time :smile:

hazmat345:
Converting to a draft until we pick this up again.

hazmat345:
I've rebased this onto current develop and dropped the commit that used cookies. Since this PR is pretty much brand new I'm going to close this and open a new one.

Issue #532, opened 2020-08-03
Auto-reload plugin logging config when file changed


hazmat345:
Right now we have an API endpoint for reloading the plugin logging config. This seems fairly unnecessary - if the plugin logging config file is changed then it's almost certain we'd want to use it immediately without having to hit a REST endpoint to kick it.

We should think about watching the file and reloading the config whenever it changes (assuming it passes validation).

The other option here is to add an endpoint that will actually allow for updating the configuration directly.

TheBurchLog:
Which logging configuration are you talking about here
1. Beer Garden Logging Config
1. Local Plugin Logging Config

If we are talking about auto reloading the Local Logging config, then I agree with you. That way a system admin can go in and update the file and without having to reload the plugin.

If we are talking about the Beer Garden Logging config, I would be nervous to auto push that out. We had discussed having different levels of logging configs in #525. So if we are doing this logging config, it would only be applied to plugins that don't have logging in the database or local.

hazmat345:
The `plugin-logging.yaml` file is the one I'm talking about. You should be able to use `watchdog` to set up a little file watcher thing. I'll need to be set up in the main app and the entry points, I think.

TheBurchLog:
@hazmat345 
Here is my thought on how to get this working properly. Once BG successfully validates the new logging config, BG requests all "local" plugins to re-initialize their logging configuration. That way each plugin will do their own internal checks to determine if they need the remote logging configuration. That way it removes the logic off of BG proper. 

This will probably need to be a new Admin command that requests `_initialize_logging` to be ran.

hazmat345:
I'm not opposed to that, but you're going further down the path than what this ticket is meant to address.

Right now, the application will read `plugin-logging.yaml` on startup and keep it in memory. Whenever a plugin comes up it will ask beer-garden for a logging configuration. The application won't re-read the file, it'll just hand the plugin what it already has in-memory.

This ticket is meant to just address what should happen if that file is changed while the application is running. Currently, nothing happens automatically. Any plugins that ask for a config will still get the version that was originally loaded during startup.

There's currently a REST endpoint you can hit that will cause the application to re-load this file. However, it's annoying because that endpoint isn't exposed anywhere in the UI, only in the swagger docs. Once you hit that endpoint the application will re-load the file and any new plugins that ask for a configuration will get the new one.

This ticket is only to remove that last part, not to automatically push anything to the plugins.

TheBurchLog:
[...]

PR #571, opened 2020-08-20
Added auto-loader for plugin logger, including plugin updates

TheBurchLog:
Created a generic monitoring tool for files. Then set this to monitor the logging-config.yaml and spawn an event when it changes. This event is captured by the Log.py handler and reloads the file. It also spawns commands to local plugins to reload their logging if they are Running. That way the changes can be immediate.

Easiest way to test this is to add and remove the file logging option. If you change the file directory or name you will see that change happen semi-immediately. 

The cool thing is now the logging file doesn't have to exist at startup. They can create it later and then those configurations will be applied automatically. 

Depends on: https://github.com/beer-garden/brewtils/pull/230
Fixes https://github.com/beer-garden/beer-garden/issues/532

### DIFF
--- a/src/app/beer_garden/log.py
+++ b/src/app/beer_garden/log.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import copy
+import os
 
 import brewtils.log
 import logging
@@ -9,6 +10,7 @@ from ruamel import yaml
 from ruamel.yaml import YAML
 
 import beer_garden.config as config
+from brewtils.models import Events
 
 _APP_LOGGING = None
 
@@ -169,7 +171,7 @@ class PluginLoggingManager(object):
         Returns:
             None
         """
-        if filename:
+        if filename and os.path.exists(filename):
             with open(filename) as log_config_file:
                 cls._PLUGIN_LOGGING = yaml.safe_load(log_config_file)
         else:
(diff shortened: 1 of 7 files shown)


hazmat345:
@TheBurchLog it's cool that you're able to reload the configurations, but that shouldn't happen automatically. Reloading the logging configuration of running plugins needs to be initiated by an admin user.

TheBurchLog:
@hazmat345 
ok.... I will remove the auto reload of configurations but I am going to leave everything stubbed out so a follow on PR can enable it through the UI.

TheBurchLog:
@hazmat345 Going to merge this in.

src/app/beer_garden/log.py
```diff
@@ -169,7 +171,7 @@ def load(cls, filename: str, default_config) -> None:
         Returns:
             None
         """
-        if filename:
+        if filename and os.path.exists(filename):
```
hazmat345, line 174: This needs to be backed out. If an incorrect file is specified an error should be raised. This change will make the system silently fall back to the default configuration.

TheBurchLog: What if instead we still raise the error but allow them to proceed with the defaults. That way if they for some reason go back later and add the configuration they don't need to restart Beer Garden to apply it.

TheBurchLog: I backed it out and created a new PR for that change. #587

PR #574, opened 2020-08-20
Correctly setting MAX_INSTANCES in beer.conf

### LINKED ISSUE #568: Propagate MAX_INSTANCES to local plugin environment
We need to be pushing the max_instances value into local plugin environments.

hazmat345:
Fixes #568


TheBurchLog:
@hazmat345 
I get it now. I do agree with setting max instances to -1. Go ahead and close out this ticket and we can get the -1 PR going.

PR #583, opened 2020-08-27
Issue/576

### LINKED ISSUE #576: Better routing logic for non-routable operations
Currently the routing logic is slightly backwards. We "parse" and operation and determine which garden an operation should be addressed to. Only right before we attempt to forward the operation do we check to see if it's in the list of "allowed" forwardable operations (and raise if it isn't).

We should just assume that everything that *isn't* forwardable is addressed to the local garden from the start.

This ticket was originally only related to routing the SYSTEM_UPDATE event:
~In the router, move System Updates to local garden events only. Plugins are the only BG entities that call this function and they can only be local.~

TheBurchLog:
The moves the `SYSTEM_UPDATE` routing action to be ran against Systems that are running on the Local Garden instance. Any other actions will be against the actual Instance object, so this is safe to assume.

Fixes #576

### DIFF
--- a/src/app/beer_garden/router.py
+++ b/src/app/beer_garden/router.py
@@ -310,14 +310,14 @@ def _determine_target_garden(operation: Operation) -> str:
         or "GARDEN" in operation.operation_type
         or "JOB" in operation.operation_type
         or operation.operation_type
-        in ("PLUGIN_LOG_RELOAD", "SYSTEM_CREATE", "SYSTEM_RESCAN")
+        in ("PLUGIN_LOG_RELOAD", "SYSTEM_CREATE", "SYSTEM_RESCAN", "SYSTEM_UPDATE")
     ):
         return config.get("garden.name")
 
     # Otherwise, each operation needs to be "parsed"
     target_system = None
 
-    if operation.operation_type in ("SYSTEM_DELETE", "SYSTEM_RELOAD", "SYSTEM_UPDATE"):
+    if operation.operation_type in ("SYSTEM_DELETE", "SYSTEM_RELOAD"):
         target_system = db.query_unique(System, id=operation.args[0])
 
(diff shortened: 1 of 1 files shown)


hazmat345:
I don't really have an issue with this, but it doesn't fix the underlying issue. I'm running this rebased onto the `child_commands` branch to have the best chance of this being fixed. And while it does seem to help, I see the same issue crop up in other places now:

> brewtils.errors.SaveError: {'message': "Unknown target garden for <Operation: operation_type=INSTANCE_INITIALIZE, source_garden_name=default, target_garden_name=None, model_type=None, model=None, args=['5f4e6af4b0fb07896ccc1c56'], kwargs={'runner_id': 'JqCQLuOodC'}>"}

I've seen the issue outside of the routing module as well. Which is bad news, because that means the solution can't be to play whack-a-mole and declare `INSTANCE_INITIALIZE` is now non-routable as well. If we do that we're still going to see the issue when these methods try to pull things out of the database themselves.

I'm pretty confidant the issue here is our database. I'm wondering if there's something we need to be doing with the read and write concerns that could give us better consistency.

hazmat345:
@TheBurchLog I'm going to just go ahead and close this... there are like 4 lines of changes but there are merge conflicts. I don't think it's worth the headache, especially since this didn't end up being necessary to fix the underlying problem we were seeing.

I'm going to modify the ticket referenced here to expand the scope a little to modify the routing order to match the behavior we discussed the other day.

hazmat345:
@TheBurchLog to close the loop on this, after updating #576 I think it's correct that this be closed rather than merged. Just because the SYSTEM_UPDATE operation is not routable does not mean that we should assume that it's intended for the local garden, which is what this PR would do. It's a subtle distinction, but it does affect the error that will be raised if someone were to attempt to issue a SYSTEM_UPDATE against a non-local system.

PR #585, opened 2020-08-27
Black Formatter Upgrade Issues

TheBurchLog:
Some new linting issues are appearing in Travis. Some of these files were not in recent PRs, so we need to address them separately. 

To recreate the same black errors I had to reinstall black
`pip install --ignore-installed black`

### DIFF
--- a/src/app/beer_garden/monitor.py
+++ b/src/app/beer_garden/monitor.py
@@ -107,7 +107,8 @@ class MonitorFile:
             self.observer.start()
 
     def on_created(self, event):
-        """ When a user VIM edits a file it DELETES, then CREATES the file, this captures that use case"""
+        """ When a user VIM edits a file it DELETES, then CREATES the file, this
+        captures that use case"""
         if self.create_event:
             publish(self.create_event)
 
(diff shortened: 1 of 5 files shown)


src/app/beer_garden/monitor.py
```diff
@@ -107,7 +107,8 @@ def __init__(
             self.observer.start()
 
     def on_created(self, event):
-        """ When a user VIM edits a file it DELETES, then CREATES the file, this captures that use case"""
+        """When a user VIM edits a file it DELETES, then CREATES the file, this
+        captures that use case"""
```
hazmat345, line 111: This actually violates a PEP: https://www.python.org/dev/peps/pep-0257/#id16. Specifically, if you're going to use a oneline docstring then it should really be one line.

However, more importantly, this is good information to have in this method's docstring. It tells me why we bother to implement this at all, which is important since my first impression was "why are we doing this, we should never allow them to just create a config file in the middle of running." But that info isn't really appropriate for a summary line (the first line in a multi-line docstring), that line should be used to describe the function itself.

 So something like this would be preferable:

> """Callback invoked when the file is created
>
> Some text editors (like VIM) will delete and create a file during interactive editing. Implementing this method allows for capturing this case."""

src/app/beer_garden/monitor.py
```diff
@@ -122,7 +123,8 @@ def on_moved(self, event):
             publish(self.moved_event)
 
     def on_deleted(self, event):
-        """ This captures if the file was deleted (Be warned that VIM does this by default during write actions)"""
+        """This captures if the file was deleted (Be warned that VIM does this by
+        default during write actions)"""
```
hazmat345, line 127: Same as above

PR #594, opened 2020-09-02
Multi instance

hazmat345:
This fixes the issue with multiple instance local plugins failing to start


TheBurchLog (approved):
Should we also use this logic for any mongo models that use the ReferenceField link?

hazmat345:
> Should we also use this logic for any mongo models that use the ReferenceField link?

Yes, probably. We should also probably change all places that use `update` to `modify`. But that's a project for another day / PR :smile:

PR #598, opened 2020-09-08
Adding ordering to system admin page

hazmat345:
Adds some order to the system admin page

### DIFF
--- a/src/ui/src/partials/admin_system.html
+++ b/src/ui/src/partials/admin_system.html
@@ -35,7 +35,7 @@
     </div>
 
     <div class="panel-body" style="padding-bottom: 0px;">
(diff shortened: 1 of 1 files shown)


TheBurchLog (commented):
I agree that we need ordering on this page and Version would be the ideal way to group them.

The issue I have with this is the way we read the list of systems. It currently is Namespace then Version. So it would make more sense to order by Namespace, then Version. Unless we want a different way a different way of presenting the hyper link.

hazmat345:
@TheBurchLog I just need some kind of sorting on the page for my own sanity :smile:

Taking namespace into account is definitely better, but that requires more cycles than I was willing to dedicate to this (since I think we should really just overhaul that page).

PR #599, opened 2020-09-09
Sync and Delete Gardens

### LINKED ISSUE #582: Delete Gardens
We need a way to delete a Garden from the database. Just in case someone redeploys a garden with a new name and the old one is still listed. The delete should remove all Systems from the Database associated with the Garden and then the Garden itself.

This should be able to be done on active Gardens. That child Garden should continue to echo out a status message and re-establish the information, but this is a good way to purge old systems for a Garden that did not get cleaned up in the database.

TheBurchLog:
This adds the ability to Delete Child Garden Instances and request Syncs to occur. 

A sync can be request for an individual Beer Garden or all Beer Gardens. When a Sync is executed against a child Beer Garden, all Gardens past that child will also be invoked to sync. That way the most complete picture can be provided.

Syncs only updates the Garden and System records. This does not sync Requests.

Fixes #582 
Depends on https://github.com/beer-garden/brewtils/pull/241


hazmat345:
@TheBurchLog I haven't looked at the code yet, but the test I ran didn't seem to work correctly. Here's what I did:

- Start a parent and child garden, correctly configured. Child garden has a running plugin.
- Verify that the parent can see the child plugin and the status is RUNNING
- Stop the parent garden
- Stop the child plugin using the child garden. It fails to update the parent (that works correctly)
- Start the parent garden
- Verify that the status of the child plugin is still RUNNING or UNRESPONSIVE
- Run a Sync
- Verify that the status has updated

The last part failed for me - the status didn't update.

Also, the child garden threw a bunch of errors trying to update the parent garden. Do we have any way to handle that yet? Just so a child garden with an offline parent doesn't completely fill up its logs.

TheBurchLog:
@hazmat345 

I would like to bet that the System Status was updated but it didn't refresh on the page. This is because the Garden record is saved to the database and the REF objects are updated automatically. This does not spawn a System Update event and will not kick off something for the Websocket to listen to.

We could iterate through the Systems on the Garden and publish an update?

We do not have a say to stop the logs from filling up when the connection is disconnected from the Parent Garden. I believe we do a gradual backoff but it still emits the whole error in the logs. We can look at creating a smaller error message or only emitting it the first time.

TheBurchLog:
@hazmat345 
Added the Event Publisher for the System objects saved to the Database. This should dynamically load the System updates in the UI as you were expecting. 

The logger error you mentioned will be resolved by #608

hazmat345:
@TheBurchLog so I now see the status update if I refresh the page, but not before. Let's talk about this - I think we may need to rework some of the frontend event handling stuff to do this correctly, and this may tie into that.

TheBurchLog:
@hazmat345 

I updated how we handle events on the System Admin page. Let me know if you are still having the same issue. We now watch the `$rootScope.systems` object for changes to invoke the `groupSystems` function. So there should be no race condition between the two executions now.

src/app/beer_garden/garden.py
```diff
@@ -57,6 +58,33 @@ def local_garden() -> Garden:
...
+                    status="RUNNING",
+                    systems=get_systems(),
+                    namespaces=get_namespaces(),
+                ),
+            )
+        )
+    else:
+        router.forward(
+            Operation(operation_type="GARDENS_SYNC", target_garden_name=garden)
+        )
```
hazmat345, line 79: This is an issue - the router module should be the only module making routing decisions. Pushing that logic down into the actual functions is not going to scale.

src/app/beer_garden/api/http/handlers/v1/garden.py
```diff
@@ -207,3 +214,64 @@ async def post(self):
         self.set_status(201)
         self.set_header("Content-Type", "application/json; charset=UTF-8")
         self.write(response)
+
+    @authenticated(permissions=[Permissions.SYSTEM_UPDATE])
```
hazmat345, line 218: Not that it really matters, but this should be `GARDEN_UPDATE`

hazmat345 (changes requested):
Think there needs to be one pretty important change on this one.

src/app/beer_garden/api/http/handlers/v1/garden.py
TheBurchLog: We don't have permissions setup for the Gardens. There is also another PR out to remove all of these and reduce it down to the CRUD + Admin permissions.

src/app/beer_garden/garden.py
TheBurchLog: Moved this up into a routing function in `route.py` function. Went back and forth in doing everything here, but I restricted it to only the logic on where to execute the function, not executing the function.

src/app/beer_garden/api/http/handlers/v1/garden.py
hazmat345: That's fair. Can we just remove this from the PR then?

src/app/beer_garden/router.py
```diff
@@ -40,6 +41,34 @@
...
+                    beer_garden.garden.sync_garden()
+                else:
+                    forward(
+                        Operation(
+                            operation_type="GARDEN_SYNC", target_garden_name=garden.name
+                        )
+                    )
+
+        # Final sync runs just in case this doesn't have any children
+        beer_garden.garden.sync_garden()
```
hazmat345, line 69: Won't this run the `sync_garden` (and so publish the event) twice? Should it just be:

```python
                if garden.name != config.get("garden.name"):
                    forward(
                        Operation(
                            operation_type="GARDEN_SYNC", target_garden_name=garden.name
                        )
                    )
```

and then run the last sync_garden?

src/app/beer_garden/garden.py
```diff
@@ -57,6 +57,22 @@ def local_garden() -> Garden:
     )
 
 
+def sync_garden():
+
+    publish(
+        Event(
+            name=Events.GARDEN_UPDATED.name,
```
hazmat345, line 64: Let's create a new event for this - I think it will be better for flexibility if we can treat updates that happen "organically" differently from sync operations.

hazmat345 (changes requested):
Couple of comments here.

Also, after looking at the most recent changes I see why you had the routing logic where you did. You weren't really routing anything, forwarding the "sync" request to the child gardens was how the "sync" was implemented. I don't think it's necessary to change it back though.

I think calling both of these things "SYNC_OPERATION" is what caused my confusion. One operation is basically "please publish your current garden status" and the other is "please send the request to publish the current garden status to a child."

Actually, now that I write that out, it seems like they're the same thing, but we should just be manually assigning the target garden in the latter case. Is there a reason why we don't do that? Instead of passing the `garden_name` as a kwarg when you're making the Operation in the Handler could you just make that the target_garden_name?

src/app/beer_garden/garden.py
TheBurchLog: Done

TheBurchLog:
@hazmat345 
The change requests have been made and I now have a PR in brewtils to go along aside with this one.
https://github.com/beer-garden/brewtils/pull/241

src/app/beer_garden/api/http/handlers/v1/garden.py
TheBurchLog: I am going to leave it, just in case we don't get the other PR merged in.

hazmat345 (approved):
Everything looks good

PR #601, opened 2020-09-09
Db efficiency

hazmat345:
This PR lets the router add kwargs to the operation. It uses this to add any database models it looks up in the course of determining where to route the operation.

This then modifies the business logic to use those kwargs if available.

### DIFF
--- a/src/app/beer_garden/systems.py
+++ b/src/app/beer_garden/systems.py
@@ -64,7 +64,8 @@ def create_system(system: System) -> System:
 
 @publish_event(Events.SYSTEM_UPDATED)
 def update_system(
-    system_id: str,
+    system_id: str = None,
+    system: System = None,
     new_commands: Sequence[Command] = None,
     add_instances: Sequence[Instance] = None,
     description: str = None,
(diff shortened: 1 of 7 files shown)


src/app/beer_garden/systems.py
```diff
@@ -147,34 +149,36 @@ def reload_system(system_id: str) -> None:
...
+        system: The System
 
     Returns:
         The removed System
 
     """
-    system = db.query_unique(System, id=system_id)
+    system = system or db.query_unique(System, id=system_id)
 
     db.delete(system)
```
TheBurchLog, line 176: Had an error when attempting to delete a System. I don't know if is this caused by the System object being passed in or something else.

```
mongoengine.errors.ValidationError: 'default:echo-1.0.0.dev0' is not a valid ObjectId, it must be a 12-byte input or a 24-character hex string
```

TheBurchLog (commented):
I am good with the approach to passing in variables if we already pulled them from the database. Had one issue when testing, that is below.

src/app/beer_garden/systems.py
hazmat345: Whoops. Needed to be `db.delete(system=system)`

PR #605, opened 2020-09-11
Updated event handling for Garden shutdown

### LINKED ISSUE #602: Schedule Jobs when Plugin is Down
We need to verify that Beer Garden does not hang during shutdown when a scheduled job is running for a plugin that is turned off.

It looks like the process that is handling the Request for the job is not closing properly.

TheBurchLog:
Changed the Event handling for Beer Garden Shutdown event to close open connections for Requests if the Shutdown event is spawned from the local garden environment. This will ensure all processes close during shutdown, and handles the edge case of Jobs tasking commands that are offline.

Fixes #602

### DIFF
--- a/src/app/beer_garden/requests.py
+++ b/src/app/beer_garden/requests.py
@@ -733,8 +733,8 @@ def cancel_request(request_id: Request) -> Request:
 
 
 def handle_event(event):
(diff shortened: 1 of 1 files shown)


hazmat345 (approved):
LGTM

PR #608, opened 2020-09-14
Child Gardens Back-Off

### LINKED ISSUE #607: Improve Child Disconnect Error
When a child disconnects from the parent via HTTP, the child continues to attempt to publish. We need to go into a back-off state until we can reconnect and skip any events there could not be published.

TheBurchLog:
When a child garden disconnects from the parent the child will go into a back-off state until it is reconnected. At the disconnect the queue is purged and no new events are accepted into the processor. Then once reconnected it runs any reconnect actions provided (Publish Garden Start Event) and begin accepting new events.

Fixes #607

### DIFF
--- a/src/app/beer_garden/events/processors.py
+++ b/src/app/beer_garden/events/processors.py
@@ -20,10 +20,12 @@ class BaseProcessor(StoppableThread):
         super().__init__(**kwargs)
 
         self._action = action
+        self._processing = True
 
     def process(self, item):
         try:
-            self._action(item)
+            if self._processing:
+                self._action(item)
         except Exception as ex:
             logger.exception(f"Error processing: {ex}")
 
@@ -42,7 +44,8 @@ class QueueListener(BaseProcessor):
         Args:
             item: New item
         """
-        self._queue.put(item)
+        if self._processing:
+            self._queue.put(item)
 
     def clear(self):
         """Empty the underlying queue without processing items"""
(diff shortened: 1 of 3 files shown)


src/app/beer_garden/events/processors.py
```diff
@@ -20,10 +20,12 @@ def __init__(self, action=None, **kwargs):
         super().__init__(**kwargs)
 
         self._action = action
+        self._processing = True
```
hazmat345, line 23: This shouldn't be in the base class

src/app/beer_garden/events/processors.py
```diff
@@ -125,16 +128,47 @@ def register(self, processor, manage: bool = True):
 class HttpEventProcessor(QueueListener):
```
hazmat345, line 128: I think this should be renamed and moved at this point... it's a lot more than a simple processor. More like a "HttpParentUpdater".

You may want to go ahead and think of the best way to organize the class structure here to separate "parent updater" stuff from "http" stuff. That'll be important to support other transports in the near future.

src/app/beer_garden/events/processors.py
```diff
@@ -125,16 +128,47 @@ def register(self, processor, manage: bool = True):
...
+            if self._ez_client.can_connect():
+                self._processing = True
+
+                if self._reconnect_action:
+                    self._reconnect_action()
+
+            else:
+                self.logger.warning(
+                    "Waiting %.1f seconds before next attempt", wait_time
+                )
```
hazmat345, line 172: This should either be taken out or changed to debug level.

src/app/beer_garden/events/processors.py
```diff
@@ -125,16 +128,47 @@ def register(self, processor, manage: bool = True):
...
             if event.name not in self._black_list:
                 event.garden = beer_garden.config.get("garden.name")
                 self._ez_client.publish_event(event)
         except Exception as ex:
             logger.exception(f"Error publishing EasyClient event: {ex}")
+
+            self.reconnect()
+
+    def reconnect(self):
+
```
hazmat345, line 152: Should add an warning-level log message indicating that we're in this reconnect state.

src/app/beer_garden/events/processors.py
```diff
@@ -125,16 +128,47 @@ def register(self, processor, manage: bool = True):
...
+        # Purge the current Queue
+        while not self._queue.empty():
+            self._queue.get()
+
+        # Back-off connection from EZ Client
+        wait_time = 0.1
+        while not self.stopped() and not self._processing:
+            if self._ez_client.can_connect():
+                self._processing = True
```
hazmat345, line 164: Should add an warning-level log message indicating that we've successfully reconnected and are exiting this reconnect state.

Even though this feels like an info-level I'd say it's better to match the level of when we entered the reconnect state. That way if the logging level is set to warning it won't get lost.

src/app/beer_garden/events/processors.py
```diff
@@ -125,16 +128,47 @@ def register(self, processor, manage: bool = True):
...
+
+                if self._reconnect_action:
+                    self._reconnect_action()
+
+            else:
+                self.logger.warning(
+                    "Waiting %.1f seconds before next attempt", wait_time
+                )
+                self.wait(wait_time)
+                wait_time = min(wait_time * 2, 30)
```
hazmat345, line 174: Hmm, we should probably make _progressive_backoff a member of `StoppableThread` instead of the application. We'd have to tweak it to allow setting the log level of the output messages, though. Maybe not worth the effort.

hazmat345 (changes requested):
Couple of changes

src/app/beer_garden/events/processors.py
TheBurchLog: Done

src/app/beer_garden/events/processors.py
TheBurchLog: Done

src/app/beer_garden/events/processors.py
TheBurchLog: Done, moved to the `HttpParentUpdater` class only

src/app/beer_garden/events/processors.py
TheBurchLog: I am not going to do that for this PR.

hazmat345 (approved):
:+1:

PR #609, opened 2020-09-14
Embedded instance

hazmat345:
This PR makes the Instance model an embedded document. The main benefit of this is getting rid of the "deep" methods on the system model (e.g. `deep_save`) that are unsafe without transaction protection.

This doesn't make any changes to the REST API. Those are in another PR, but I'm not certain those are warranted yet.


src/app/beer_garden/plugin.py
```diff
@@ -203,15 +206,17 @@ def update(
     updates = {}
 
     if new_status:
-        updates["set__status"] = new_status
-        updates["set__status_info__heartbeat"] = datetime.utcnow()
+        updates["set__instances__S__status"] = new_status
+        updates["set__instances__S__status_info__heartbeat"] = datetime.utcnow()
```
TheBurchLog, line 210: It wouldn't let me put a comment above, but I thought he DB query for instances by ID would no longer work. (line 198)

TheBurchLog (commented):
Just the one comment this time.

src/app/beer_garden/plugin.py
hazmat345: Arg, I missed one. Good catch, thank you!

hazmat345: Ok, this should be resolved

src/app/beer_garden/systems.py
```diff
@@ -280,3 +343,15 @@ def handle_event(event: Event) -> None:
...
         elif event.name == Events.SYSTEM_REMOVED.name:
             db.delete(event.payload)
+
+        elif event.name == Events.INSTANCE_UPDATED.name:
+            if not event.payload_type:
+                logger.error(f"{event.name} error: no payload type ({event!r})")
+                return
+
+            record = db.query_unique(Instance, id=event.payload.id)
+
```
TheBurchLog, line 353: Why are you not using something like `_from_kwargs` function that you made in `plugin.py` to pull back the instance record OR the `get_instance` function above?

hazmat345: This is just another one I missed. I'll need to tweak how this works.

TheBurchLog (approved):
Once I merged in the latest from V3 everything ran as expected.

hazmat345:
@TheBurchLog I just tested that as well and it worked for me too

PR #610, opened 2020-09-15
Roles Redesign

TheBurchLog:
Before we add complexity to support Namespace filtering, we need to simplify the roles we have in Beer Garden. We are removing Nested Roles and API specific Permissions. 

Depends on: https://github.com/beer-garden/brewtils/pull/237


src/app/beer_garden/api/http/authorization.py
```diff
@@ -16,43 +16,24 @@
 
 
 class Permissions(Enum):
+    # Redesign of the permissions simplifies the grouping. Admin permissions are
```
hazmat345, line 19: I'd take out the line about the redesign, it doesn't really help anyone understand the code as it currently is.

hazmat345 (commented):
This mostly LGTM. Only nit is a comment.

Have you tested what happens when upgrading from v2 to this? I'm concerned that there's no migration plan in place.

TheBurchLog:
@hazmat345 
We would have to write something for this migration plan. If we wanted to roll people forward from V2 to V3, then we would have to flatten all of the roles. So the permissions assigned either through the current role or nested would get applied.

hazmat345:
@TheBurchLog it's up to you whether you think a real migration plan is worth the effort. I'm just concerned that going from a v2 database to v3 (with these changes) is going to fail when we do our startup database checks because what's in the database doesn't match the new model. If we don't handle that case (by at least dropping the roles collection) then we will never recover.

src/app/beer_garden/db/mongo/util.py
```diff
@@ -152,8 +136,18 @@ def ensure_users(guest_login_enabled):
             ).save()
 
 
-def ensure_embedded_command():
-    """Ensures that the Command model is an EmbeddedDocument
+def ensure_model_migration():
+    """Ensures that the Role model is an flatten and Command model is an
```
hazmat345, line 140: is flattened

src/app/beer_garden/db/mongo/util.py
```diff
@@ -164,19 +158,24 @@ def ensure_embedded_command():
     it would be better if we could seamlessly move the existing commands into existing
     Systems.
     """
-    from beer_garden.db.mongo.models import System
+    from .models import Role, System
```
hazmat345, line 161: Is there a reason why you switched to relative imports? Just curious if you hit an issue that could crop up other places.

hazmat345 (approved):
One wording nitpick, otherwise LGTM (except for the tests failing)

src/app/beer_garden/db/mongo/util.py
TheBurchLog: This was cause by a thing called "Lazy Developer Syndrome", I allowed PyCharm to auto create the import for me. I will switch it back.

PR #612, opened 2020-09-21
Uses ThreadPool to shutdown plugins

### LINKED ISSUE #567: Local plugin shutdown concurrency
When we go to shut down local plugins we wait the shutdown timeout consecutively for all the plugins. We should be doing that wait concurrently.

TheBurchLog:
Updates the shutdown process to concurrently run 10 plugin shutdowns until they are all closed. 

Fixes #567

### DIFF
--- a/src/app/beer_garden/local_plugins/manager.py
+++ b/src/app/beer_garden/local_plugins/manager.py
@@ -206,7 +206,9 @@ class PluginManager(StoppableThread):
     @classmethod
     def stop_all(cls):
 
-        shutdown_pool = ThreadPoolExecutor(1 if len(cls.runners) < 1 else len(cls.runners))
+        shutdown_pool = ThreadPoolExecutor(
+            1 if len(cls.runners) < 1 else len(cls.runners)
+        )
         futures = []
 
         for runner in cls.runners:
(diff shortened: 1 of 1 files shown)


hazmat345 (commented):
I'm skeptical it'll be this easy, unfortunately.

Have you tested this with the sleeper plugin? Specifically, making a request that will take much longer than the shutdown timeout to complete and then asking beer-garden to shutdown? It should wait 10 seconds (or whatever `shutdown_timeout` is) and then forcibly terminate the plugin process. The log messages should be enough to explain what's going on to an administrator who doesn't understand what's happening.

Should also probably make the max number of threads on the executor the number of active runners the plugin manager has, not 10.

TheBurchLog:
@hazmat345 

I setup a scheduler to run the sleeper plugin for 30 seconds. Then once it run, I shutdown Beer Garden. After 10 seconds it forces Beer Garden to shutdown.

I also added your comment about adding threads to match the total number running. If nothing is running, instead of setting it to 0, we set it to 1 just in case something weird happens.

hazmat345:
Sweet, that's a pleasant surprise. LGTM then!

PR #615, opened 2020-09-21
Updated Unknown Garden Exception

### LINKED ISSUE #563: Better response for request creation error
Attempting to create a request for a system/version/namespace combo that doesn't exists results in a "can't determine the correct garden" error. It should be a "this combo doesn't exist" error.

TheBurchLog:
Added some verbiage to make it more clear for a user who accidentally mistypes a manual request.

Fixes #563


hazmat345 (approved):
Can we add a "target" or "destination" in there somewhere? So like, "could not determine the target garden..."

Issue #611, opened 2020-09-18
Overall Performance


hazmat345:
Right now the following makes the application unusably slow:

1. Start with the normal example plugins
2. Make the complex plugin have 9 instances instead of 2
3. Make that complex plugin have 6 versions instead of just 1

This is not an unreasonable amount of plugins, so this needs to be better.

PR #616, opened 2020-09-21
Making MongoParser work with updated SchemaParser

hazmat345:
This is part of beer-garden/beer-garden#611. It depends on and enables https://github.com/beer-garden/brewtils/pull/239.


TheBurchLog (approved):
Looks good to me.

PR #617, opened 2020-09-21
Misc Updates

hazmat345:
This PR has a couple of things that fell out of investigating #611.

- Adds type hints in some places
- Names some threads to help with debugging
- Adds a `__main__` module to the HTTP entry point package. This allows running *only* the http entry point without any of the overhead of the main process.


hazmat345:
This branch is named poorly, closing this PR.

PR #618, opened 2020-09-21
Misc updates

hazmat345:
This PR has a couple of things that fell out of investigating #611.

- Adds type hints in some places
- Names some threads to help with debugging
- ~Adds a __main__ module to the HTTP entry point package. This allows running only the http entry point without any of the overhead of the main process.~ jk, that needs more work

### DIFF
--- a/src/app/beer_garden/api/entry_point.py
+++ b/src/app/beer_garden/api/entry_point.py
@@ -72,7 +72,11 @@ class EntryPoint:
         self._logger = logging.getLogger(__name__)
         self._process = None
         self._ep_conn, self._mp_conn = Pipe()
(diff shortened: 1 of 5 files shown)


TheBurchLog:
The first two bullets look good. I am not seeing where you added a main function to http entry point.

Also, you need to fix the linting error.

hazmat345:
@TheBurchLog that's because I ripped that commit out of this branch :facepalm:. That functionality needs a bunch of extra changes to actually work correctly and I don't feel like getting them into a state to actually merge them in. So I moved that commit over to my debugging branch.

hazmat345:
Formatting fixed!

PR #621, opened 2020-09-22
Not using ThreadPoolExecutor for everything in HTTP package

hazmat345:
This depends on #620, and is part of #611 

This transforms the `ExecutorClient` into a `SerializeHelper`. It no longer executes the routing operation (and then the actual operation) inside of a `ThreadPoolExecutor`. The route operation is now just called directly and the client is used as a serialization helper.

Future work could be to either remove this class completely or pull it up to the entry point level (or the router) where it can be used by all entry points.


TheBurchLog (approved):
I agree that we should eventually roll this up to the Entry Point level.

PR #622, opened 2020-09-24
Heartbeats

hazmat345:
This PR is part of #611.

It splits out the instance heartbeat method so an event isn't published every time.

### DIFF
--- a/src/app/beer_garden/api/http/handlers/v1/instance.py
+++ b/src/app/beer_garden/api/http/handlers/v1/instance.py
@@ -149,11 +149,7 @@ class InstanceAPI(BaseHandler):
 
             elif operation == "heartbeat":
                 response = await self.client(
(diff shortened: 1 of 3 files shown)


TheBurchLog:
@hazmat345 
I am conflicted on this one. Yes, not publishing these events will save a lot of unnecessary bandwidth. The downside is that this only solves a problem for local Beer Gardens. When we move to the Parent/Child scenario, the parent will never get the latest active time and will make the plugin as inactive. So we need the Garden Sync function running on a timer that is equal to the Timeout or have the `PluginMonitor` class only check plugins that are running locally.

So yes, your code is good. It just causes issues in other areas that we need to address.

hazmat345:
@TheBurchLog good point - the plugin status monitor should only be monitoring local systems.

PR #623, opened 2020-09-25
Routing

hazmat345:
Part of #611 

tl;dr - Changes routing structure to no longer need to query the database for routing decisions.

This PR changes the layout of the data structure that's used to make routing decisions. This aims to simplify the overall routing decision by maintaining 3 separate lookup tables (aka, fast) using the 3 pieces of data we will need to use to make routing decisions:

- system "name" (namespace, name, and version, aka `str(system)`)
- system id
- instance id

We maintain a dictionary for each, where the values are simply the name of the garden that an operation intended for that "thing" should be routed to. This lets us make *fast* decisions about where to route things without needing to consult the database.

Data consistency is not really affected by this PR: we still have things that will update the routing table (new system, for example) do so directly, rather than trying to wait for an event handler to deal with it.

This PR does not address the fact that this data is going to be local to a specific entry point process. So right now the scheduler will not work properly with new systems, it would require a restart for the routing module to "pick up" the new route. This *does* need to be addressed, but worst case I think it would work to have an event handler in the routing module deal with this *as well as* have the system creation method updating the routing table directly. That would enable the other processes to pick up the new system *eventually* while also being causal enough to not cause problems during plugin standups. However, I do want to see if using some sort of shared memory would be performant enough - I feel like that would be cleaner.

Also, the "publish garden status" in the app module (called during startup and shutdown) was essentially the same thing as the garden sync functionality. So the app module just uses that now.

### DIFF
--- a/src/app/beer_garden/app.py
+++ b/src/app/beer_garden/app.py
@@ -6,9 +6,8 @@ from functools import partial
 from apscheduler.executors.pool import ThreadPoolExecutor as APThreadPoolExecutor
 from apscheduler.schedulers.background import BackgroundScheduler
 from brewtils import EasyClient
-from brewtils.models import Event, Events, Garden
+from brewtils.models import Event, Events
 from brewtils.stoppable_thread import StoppableThread
-from more_itertools import flatten
 from pytz import utc
 
 import beer_garden.api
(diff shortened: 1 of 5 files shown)


TheBurchLog (commented):
I don't have any problems with the code, just a thought.

What are your thoughts on calling a function in garden.py to query for the garden associated with the instance id, system id, or system name for the use case that it is not found. That would be the Hail Mary to determine the route, just in case the process never captured the routing information.

hazmat345:
@TheBurchLog definitely not opposed to something like that. This PR is fairly fragile for being such a core piece, and something like that would definitely help.

I think an interesting alternative would be to do a complete "refresh" of the routing state if we encounter something that's unroutable? My thinking is that we should always be able to route, and a failure to do so indicates that something is "wrong" with our routing state, so we should reload it.

PR #625, opened 2020-09-28
Motor

hazmat345:
Part of #611.

Transitions the instance update and heartbeat operations to use async database operations.

### DIFF
--- a/src/app/beer_garden/plugin.py
+++ b/src/app/beer_garden/plugin.py
@@ -336,11 +336,9 @@ async def update_async(
         update["instances.$.status"] = new_status
         update["instances.$.status_info.heartbeat"] = datetime.utcnow()
 
-    # TODO
-    # if metadata:
-    #     metadata_update = dict(instance.metadata)
-    #     metadata_update.update(metadata)
-    #     updates["set__instances__S__metadata"] = metadata_update
+    if metadata:
+        for k, v in metadata.items():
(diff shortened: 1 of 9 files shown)


src/app/beer_garden/plugin.py
```diff
@@ -302,6 +305,76 @@ def read_logs(
...
+        system: The System
+        new_status: The new status
+        metadata: New metadata
+
+    Returns:
+        The updated Instance
+    """
+    query = {"instances._id": ObjectId(instance_id)}
+    projection = {"instances.$": 1, "_id": 0}
+    update = {}
```
TheBurchLog, line 333: This is kind of an overarching comment on this file. I get it that we need to create these objects, but in my ideal world this stuff would be pushed down to the DB code some home. Because we are doing this for a select few of the commands, yeah we can't do it. But I still want us to think about how we could handle that going forward. If everything migrated over to motor, then I would not want to see this here.

TheBurchLog (approved):

src/app/beer_garden/plugin.py
hazmat345: Completely agree. I think the correct thing to do is to start to make use of our db "api". That's going to be involved... I'll make an issue to track it.

hazmat345: #634

PR #626, opened 2020-09-28
Monitor

hazmat345:
Part of #611, follow-on to #622.

This moves the plugin status monitor into the plugin module and modifies it to only be concerned with systems that are associated with the local garden.


TheBurchLog (commented):
This works as expected. A BG instance will only check the status of local plugins. When I shutdown the child garden, it does not run checks on those plugins to verify their status.

I did run into one weird issue. When I `kill -9` the process running a plugin on a child the child eventually says that it is unresponsive as expected. But that event is not propagated to the parent. I was expecting this when it runs the `update(` function.

hazmat345:
@TheBurchLog did you refresh or check the database on the parent directly? I've been seeing weirdness on the admin page that could be the same issue... I opened a ticket for it: https://github.com/beer-garden/beer-garden/issues/628

PR #631, opened 2020-09-29
Fixed validation error for choices

### LINKED ISSUE #630: Validation error when using choices decorator for parameter
I got a validation error for plugin using command in the choices decorator for parameter.

parameter sample: ```@parameter(
        key="message",
        type="String",
        description="Say what we want",
        optional=False,
        choices={
            "type": "command",
            "display": "select",
            "strict": True,
            "value": "get_choices_with_argument(key=${index})",
        },
    )```

Error: ```brewtils.errors.ModelValidationError: Could not find System named 'dynamic' matching version '3.0.0.dev0'```

west270:
Added namespace to ```choices_request``` to resolve ```brewtils.errors.ModelValidationError```
You can test using command ```say_specific_with_choices_argument``` from the dynamic plugin found in https://github.com/beer-garden/example-plugins

Fixes #630 #637

### DIFF
--- a/src/app/beer_garden/plugin.py
+++ b/src/app/beer_garden/plugin.py
@@ -76,9 +76,7 @@ def initialize(
 
 @publish_event(Events.INSTANCE_STARTED)
 def start(
-    instance_id: str = None,
-    instance: Instance = None,
-    system: System = None,
+    instance_id: str = None, instance: Instance = None, system: System = None
 ) -> Instance:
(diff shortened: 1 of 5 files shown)


hazmat345:
Rebased this onto v3 and combined a couple of commit. I moved the one fixture up into conftest since it should really be set for all tests - you shouldn't have had to worry about it.

PR #632, opened 2020-09-29
Issue/629

### LINKED ISSUE #629: Plugin system creation conflict error logging
Part of #611.

Scenario: Standing up multiple plugins with multiple instances where the systems don't already exist.

The current plugin startup procedure generates a *lot* of errors here. Basically, all instances of a "new" system will check to see if the system already exists. It doesn't, so they will all attempt to create it. This will fail for every plugin except the first that tries it.

Issue: This generates a large amount of error messages that aren't really errors. It also generates some additional network traffic, but I don't think it's significant.

This generates a *lot* of error messages and stacktraces, which is bad because this isn't an error condition. Because we don't have any control over the timing of when plugins come up this is actually expected behavior (evidenced by the fact that all the plugins successfully recover - if they get a `conflict` error they just re-query for the system).

The question is where to put the error handling for this. My first instinct is in the http entry point - we need to be able to set the HTTP response code for this, and that doesn't belong anywhere outside the entry point. On the other hand, that feels kind of gross since this isn't an error regardless of what entry point is being used. It'd be nice to only have to deal with this in once place.

hazmat345:
Fixes #629.

This PR makes system creation no longer fail if a system already exists - it turns the old "insert" operation into an "upsert".

I'm not excited about having to do this, but I don't see a better way at the moment. This is how things work in v2, and while I was hoping to improve on this for v3 it doesn't look like that's happening.

One thing that is different that v2: the HTTP response code is the same for both cases. In v2 we would return a 201 for a system that was actually created as a result of the POST and a 200 for a system that was just updated. In v3 we always return the same code. This doesn't *really* matter as the plugins don't care what the response code is as long as it's a success. It's just less REST-ideal.

### DIFF
--- a/src/app/beer_garden/db/mongo/api.py
+++ b/src/app/beer_garden/db/mongo/api.py
@@ -4,7 +4,7 @@ import logging
 from box import Box
 from brewtils.models import BaseModel
 from brewtils.schema_parser import SchemaParser
-from mongoengine import connect, register_connection, DoesNotExist
+from mongoengine import NotUniqueError, connect, register_connection, DoesNotExist
 from pymongo.errors import ConnectionFailure, ServerSelectionTimeoutError
 from typing import List, Optional, Type, Union, Tuple
 
@@ -18,6 +18,7 @@ from beer_garden.db.mongo.util import (
     ensure_roles,
     ensure_users,
 )
+from beer_garden.errors import NotUniqueException
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


TheBurchLog (commented):
@hazmat345 
In the scenario where you have 10 "instances" coming online and they all try to create that record. What will happen? Will the first one succeed, then seconds later the next one overwrite it because both of them checked and the system didn't exist prior.

hazmat345:
@TheBurchLog I think I see what you're saying. No, I don't think that's a possibility because we're using the modify API. Any changes that are made will only be to that single document and so they should happen atomically.

Issue #633, opened 2020-09-30
Asyncio


hazmat345:
Opening this as a place to track discussion about a potential switch to expanding our use of asyncio.

TheBurchLog:
To implement this, we need to roll out changes from https://github.com/beer-garden/beer-garden/pull/638

hazmat345:
Closing - if we move in any direction it should probably be *away* from Tornado / asyncio.

PR #638, opened 2020-10-01
Using executor for executing CREATE_REQUESTS in async context

### LINKED ISSUE #637: Command-based choices error
The switch away from using an executor didn't address the case with command-based choices validation. Need to fix.

Found during #630

hazmat345:
Fixes #637


TheBurchLog (approved):
When implementing #633 we need to address some of the things added in this ticket.

PR #642, opened 2020-10-02
Issue/635

### LINKED ISSUE #635: Incorrectly publishing system create error
Directly caused by #632 - the `publish_event` decorator on system create means that we'll publish an ERROR event when we shouldn't.

hazmat345:
This fixes #635.

### DIFF
--- a/src/app/beer_garden/systems.py
+++ b/src/app/beer_garden/systems.py
@@ -163,7 +163,7 @@ def upsert(system: System) -> System:
         The created / updated system
     """
     try:
-        return create_system(system)
+        return create_system(system, _publish_error=False)
     except NotUniqueException:
         logger.warning(f"Not unique, updating {system.name}")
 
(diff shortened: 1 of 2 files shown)


src/app/beer_garden/systems.py
```diff
@@ -163,10 +163,8 @@ def upsert(system: System) -> System:
         The created / updated system
     """
     try:
-        return create_system(system)
+        return create_system(system, _publish_error=False)
```
TheBurchLog, line 166: This is new to me. I did not know that you could pass in undefined variables like this. Everything works, just wanted to let you know that I didn't know about this.

TheBurchLog (approved):
:+1:

src/app/beer_garden/systems.py
hazmat345: It's not really undefined, it's just intended for the decorator "wrapper" function instead of the normal "wrapped" one. That's why I use `pop`:

```
 _publish_success = kwargs.pop("_publish_success", True)
```

 That way those kwargs are removed before the wrapped function call.

PR #644, opened 2020-10-02
Added Force Delete for Systems.

### LINKED ISSUE #526: Deleting System on Down Garden
Scenario:
Child BG turns on
Child BG turns off
Admin tries to delete system from Child on Parent
Receive error: `Something went wrong on the backend: Unknown target garden for SYSTEM_DELETE `

When we run through this scenario, we are unable to route the request down to the child, so it fails to delete. We need to decide if the downstream is available, should we try to run it locally if it is a non Request function?

TheBurchLog:
Added the functionality to force delete a System record from the database. This does not take into consideration RabbitMQ or shutting down the plugin. If the plugin is still operational it will throw errors, but this is the last resort to deleting systems if you get in a bad state. So we ask like 3 times if you are sure.

Fixes #526

### DIFF
--- a/src/app/beer_garden/api/http/handlers/v1/system.py
+++ b/src/app/beer_garden/api/http/handlers/v1/system.py
@@ -88,18 +88,15 @@ class SystemAPI(BaseHandler):
           - Systems
         """
 
-        if self.get_argument("force", default="").lower() == "true":
-            await self.client(
-                Operation(
-                    operation_type="SYSTEM_DELETE_FORCE",
-                    args=[system_id],
-                    kwargs={"force": True},
-                )
+        await self.client(
+            Operation(
+                operation_type="SYSTEM_DELETE",
+                args=[system_id],
+                kwargs={
+                    "force": self.get_argument("force", default="").lower() == "true"
+                },
             )
-        else:
-            await self.client(
-                Operation(operation_type="SYSTEM_DELETE", args=[system_id])
-            )
+        )
 
         self.set_status(204)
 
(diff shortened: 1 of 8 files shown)


hazmat345 (changes requested):
I think this is pretty good, couple of suggestions:

I feel like instead of just having the "force delete" call remove directly it should still call purge_system. Except now it passes the force flag as a parameter, and that parameter tells the purge operation to keep going if any errors are encountered. That way it's more of a true "force delete" and less of a "database delete." I'm thinking something like:

```python
def system_purge(..., force=False):
   try:
        stop()
    except:
        if force:
             # log something and don't raise
        else:
            raise
```

The other thing is I don't think we need the additional "are you sure" box after they click Force Delete again... there are enough warnings and the button is red enough that I think another confirm action is superfluous.

TheBurchLog:
@hazmat345 
I updated how the force delete function now works. 

Key changes
- If local, attempt to run the Purge function, but catch all exceptions to allow the delete.
- If managed by another Garden, don't route and only delete locally. Didn't want to add more logic in router class.
- UI displays different text to reflect the scenario they are in.

hazmat345 (approved):
I made a couple of changes regarding how the exceptions and routing are handled.

I'm not sure that the dialog will ever show up in the UI when trying to delete a child system. But I'm pretty sure that's because we don't actually return anything useful in the HTTP response for that case. Once the router determines that "forward" is the correct action it just puts the operation on the forward queue and the original action completes successfully. It's only when the forwarder goes to actually forward the operation that it encounters a problem (the child not running in my testing). This is a problem with our architecture, though, so I'm going to merge this.

PR #656, opened 2020-10-07
Changing tags in compose to 3.0.0a1

hazmat345:
This would change the default tag in the non-TLS compose file to be the released tag instead of unstable.

I'm not sure if this something we want to do... thoughts?


TheBurchLog:
*shrug* Not sure if this will buy us anything until we are happy with an Alpha release. Once we hit that and want to lock into a version, then I am good with this.

hazmat345:
@TheBurchLog yeah, we're already on alpha 2. I'm not going to bother keeping this updated.

PR #657, opened 2020-10-07
Pruner ResultSet no longer caches

### LINKED ISSUE #604: Pruner and Mongo Updates
Seeing an issue where the VM is hitting memory caps when running the Pruner Job if a ton of events are present.

Current example:
100,000 Records in Database
9,000 RabbitMQ Events in Database

Suggested to turn off pruner and let the database build up, then try running the pruner.

hazmat345:
Fixes #604 in version 3.

Originally developed in beer-garden/bartender#109


TheBurchLog (approved):
You know I fully support this one :)

PR #658, opened 2020-10-07
Bumping to fontawesome 5

hazmat345:
Updates the fontawesome dependency to version 5.


hazmat345:
@TheBurchLog found those icons are actually present in v4, so we don't need to upgrade (and the new beer icon looks worse :cry:)

PR #661, opened 2020-10-09
Removing everything related to thrift

hazmat345:
Removes the last vestiges of thrift.


TheBurchLog (approved):
Everything looks fine, but I was expecting to remove a removal of a dependency. Shouldn't thriftpy2 be somewhere in there? I looked and didn't see it, so I don't know if we removed that a while ago.

hazmat345:
I removed it a while ago, it was a cython thing and causes me enough issues that I removed it ASAP.

TheBurchLog:
@hazmat345 
The test cases were not removed.

Issue #590, opened 2020-09-01
Local Plugin Runner logging


hazmat345:
This is a mess, needs to be cleaned up.

hazmat345:
#664 will help with this. Still TODO:

- I'm pretty sure the `LockListener` (and all `QueueListener`s in general) will not exhaust their queue before stopping once their stop flag is set. This is almost certainly *not* what we want for the `PluginRunner.log_reader`. So that needs to be investigated.
- Less important, but still kind of important are the timestamps for the generated records. We aren't actually *logging* the messages until the log reader is started, which results in all the records having basically the same time. Adding locking at least groups the specific runner messages together, but the timestamps are still wrong. It would be nice if we could get this working better. Perhaps a way to construct `LogRecord`s as they come in, rather than just throwing the strings on the queue and logging later.

PR #664, opened 2020-10-10
Issue/590

hazmat345:
This helps with #590 but probably doesn't close it. This PR is ready to be reviewed, though.

This PR reworks how local plugin STDOUT and STDERR is handled when the Plugin inside the runner process fails to initialized. See the docstring for an in-depth discussion, but the short version is that all captured messages will be logged to both the application log and an plugin error file in the plugin working directory.

### DIFF
--- a/src/app/beer_garden/events/processors.py
+++ b/src/app/beer_garden/events/processors.py
@@ -53,6 +53,24 @@ class QueueListener(BaseProcessor):
                 pass
 
 
+class LockListener(QueueListener):
+    """Listens for items on a multiprocessing.Queue"""
+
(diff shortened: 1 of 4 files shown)


TheBurchLog (approved):
Approving if travis passes.

This is how I tested it:

I added `raise Exception` to a plugins `__init__` function to verify that an error was raised in the system logs. Then I added the exception to a command to ensure that it was raised in the plugin logger.

hazmat345:
@TheBurchLog travis should pass, but only because we're skipping all of these tests. I have a TODO to go back and fix them later.

PR #666, opened 2020-10-13
Grab the request object if the Job still exists for status

### LINKED ISSUE #665: Scheduler Counts not Updating
In V3, the scheduler is not updating the counts for success and failures.

TheBurchLog:
The scheduled was not getting the latest status from the request after our recent refactoring. This will grab the request object if it is required.

Fixes #665


hazmat345 (approved):
I wish there were a way to be able to specify the projection here... we really only need the status be we get everything including children :/

Oh well, that's a problem for another PR!

Issue #647, opened 2020-10-05
File Based Triggers


TheBurchLog:
Create new Trigger that executes against file changes on the local directory of host system for Beer Garden. This should operate similar to how local logger file is reloaded.

We should expose file parameters as valid input. Similar to how we inject variables for local plugin logging.

- `$file_name`
- `$src_path`
- `$file`

TheBurchLog:
This will probably be impacted by this other ticket: https://github.com/beer-garden/beer-garden/issues/368

hazmat345:
Hmm, not sure why this wasn't closed by the PR.

PR #670, opened 2020-10-14
Implementation of file-based triggers

SpecificallyGravity:
Adding file based triggers to UI and back-end.

Depends on: https://github.com/beer-garden/brewtils/pull/251

Fixes: #647


src/app/beer_garden/app.py
```diff
@@ -299,12 +304,21 @@ def _setup_scheduler():
...
+            "job_defaults": job_defaults,
+            "timezone": utc,
+        }
+
+        # return BackgroundScheduler(
+        #     jobstores=job_stores,
+        #     executors=executors,
+        #     job_defaults=job_defaults,
+        #     timezone=utc,
+        # )
```
TheBurchLog, line 319: Don't leave in comment blocks

src/app/beer_garden/scheduler.py
```diff
@@ -4,17 +4,369 @@
...
+    """
+    if isinstance(request, dict):
+        for k, v in request.items():
+            try:
+                request[k] = inject_values(v, dictionary)
+            except (ReferenceError, IndexError):
+                pass
+        return request
+
+    elif isinstance(request, str):
```
TheBurchLog, line 78: A way that we do this in other places is to use `six.string_types` instead of `str`. An example of this can be seen when we validate request parameters.

TheBurchLog (commented):
This is my quick read of the code comments. I will run the code later today to give you more feedback.

src/app/beer_garden/db/mongo/jobstore.py
```diff
@@ -24,6 +24,8 @@ def construct_trigger(trigger_type: str, bg_trigger) -> BaseTrigger:
         return IntervalTrigger(**bg_trigger.scheduler_kwargs)
     elif trigger_type == "cron":
         return CronTrigger(**bg_trigger.scheduler_kwargs)
+    elif trigger_type == "file":
+        return None
```
hazmat345, line 28: I think this is causing issues? When I try to create a job with a file trigger I get the "Oh Snap" error display with this error:

> {'trigger': ["Unable to use schema. Ensure there is a deserialization_schema_selector and that it returns a schema when the function is passed in {'pattern': ['.*'], 'path': '/home/mppatri/scratch/test_files', 'recursive': False, 'callbacks': {'on_created': True, 'on_modified': False, 'on_moved': False, 'on_deleted': False}}. This is the class I got. Make sure it is a schema: None"]}

This function is supposed to construct a `BaseTrigger` of some kind so this is probably a bug?

hazmat345 (commented):
This is really interesting, but I ran into a bug trying to test it out. I'll take a closer look tomorrow.

TheBurchLog:
I ran into a error when running a scheduled job.

```
2020-10-21 08:34:05,059 - beer_garden.scheduler - ERROR - Could not fetch job for parameter injection: 'None' is not a valid ObjectId, it must be a 12-byte input or a 24-character hex string
  File "/.pyenv/versions/beer-garden-v3/lib/python3.7/site-packages/mongoengine/base/fields.py", line 507, in to_mongo
    return ObjectId(str(value))
  File "/.pyenv/versions/beer-garden-v3/lib/python3.7/site-packages/bson/objectid.py", line 110, in __init__
    self.__validate(oid)
  File "/.pyenv/versions/beer-garden-v3/lib/python3.7/site-packages/bson/objectid.py", line 213, in __validate
    _raise_invalid_id(oid)
  File "/.pyenv/versions/beer-garden-v3/lib/python3.7/site-packages/bson/objectid.py", line 40, in _raise_invalid_id
    " or a 24-character hex string" % oid)
bson.errors.InvalidId: 'None' is not a valid ObjectId, it must be a 12-byte input or a 24-character hex string

During handling of the above exception, another exception occurred:

  File "/src/app/beer_garden/scheduler.py", line 397, in run_job
    db_job = db.query_unique(Job, id=job_id)
  File "/src/app/beer_garden/db/mongo/api.py", line 231, in query_unique
    query_set = _model_map[model_class].objects.get(**kwargs)
  File "/.pyenv/versions/beer-garden-v3/lib/python3.7/site-packages/mongoengine/queryset/base.py", line 260, in get
    result = next(queryset)
  File "/.pyenv/versions/beer-garden-v3/lib/python3.7/site-packages/mongoengine/queryset/base.py", line 1590, in __next__
    raw_doc = next(self._cursor)
  File "/.pyenv/versions/beer-garden-v3/lib/python3.7/site-packages/mongoengine/queryset/base.py", line 1666, in _cursor
    self._cursor_obj = self._collection.find(self._query, **self._cursor_args)
  File "/.pyenv/versions/beer-garden-v3/lib/python3.7/site-packages/mongoengine/queryset/base.py", line 1715, in _query
    self._mongo_query = self._query_obj.to_query(self._document)
  File "/.pyenv/versions/beer-garden-v3/lib/python3.7/site-packages/mongoengine/queryset/visitor.py", line 90, in to_query
    query = query.accept(QueryCompilerVisitor(document))
  File "/.pyenv/versions/beer-garden-v3/lib/python3.7/site-packages/mongoengine/queryset/visitor.py", line 185, in accept
    return visitor.visit_query(self)
  File "/.pyenv/versions/beer-garden-v3/lib/python3.7/site-packages/mongoengine/queryset/visitor.py", line 79, in visit_query
    return transform.query(self.document, **query.query)
  File "/.pyenv/versions/beer-garden-v3/lib/python3.7/site-packages/mongoengine/queryset/transform.py", line 120, in query
    value = field.prepare_query_value(op, value)
  File "/.pyenv/versions/beer-garden-v3/lib/python3.7/site-packages/mongoengine/base/fields.py", line 513, in prepare_query_value
    return self.to_mongo(value)
  File "/.pyenv/versions/beer-garden-v3/lib/python3.7/site-packages/mongoengine/base/fields.py", line 509, in to_mongo
    self.error(str(e))
  File "/.pyenv/versions/beer-garden-v3/lib/python3.7/site-packages/mongoengine/base/fields.py", line 171, in error
    raise ValidationError(message, errors=errors, field_name=field_name)
mongoengine.errors.ValidationError: 'None' is not a valid ObjectId, it must be a 12-byte input or a 24-character hex string
```

TheBurchLog:
@SpecificallyGravity 

Can you also take a look at when happens when a user doesn't enter a Pattern. When I tested it with "on_create" event nothing was triggered. So if a blank pattern doesn't do anything, then we shouldn't allow users to save it that way.

src/app/beer_garden/db/mongo/jobstore.py
SpecificallyGravity: I've changed this to now raise an error.  The MongoDB jobstore is linked to `APScheduler` which does not support asynchronous behavior (required for `FileTrigger`).  To schedule jobs, you should be going through the top-level `MixedScheduler` class, which will schedule jobs with file triggers separately.

SpecificallyGravity:
Valid injection syntax (only for FileFriggers right now): "Your text {x.y} here" where x is a source type, and y is a valid source field.

Current Source Types and their fields:
{event.*} : event_type, is_directory, src_path, dest_path* (only valid on file move events)
{trigger.*} : pattern, path, recursive, callbacks

hazmat345 (changes requested):
Couple of things:

- Is the pattern supposed to be a regex? If so, can we add that to the description?
- Creating a trigger for a directory that doesn't exist doesn't fail very gracefully.
- I'm still not able to get things working. I created a trigger for created files in a directory with the pattern `.*`. Then I created a file in that directory, and I expected that the job would have been fired, but it wasn't. Am I doing something incorrectly?

SpecificallyGravity:
@hazmat345 

- Patterns are specifically NOT regex (that's a separate `watchdog` handler this isn't implemented).  I can't find on all of the syntax supported by `watchdog`, but I believe it's just wildcards

- Adding a check for valid directories now

- I think I understand what you may be running into; if I remember correctly, when `watchdog` checks against your patterns it uses the `trigger.path` + `file` as input.  Practically speaking, this means that you may need to have a pattern similar to `*/.*` to achieve the functionality you're expecting.

hazmat345 (commented):
Apologies if this isn't ready for another review but I tested it out again today.

I'm seeing duplicate requests spawned when I don't think they should be. I made a FileTrigger based job and only selected the `on_created` callback, and when I created a file in the directory I had multiple requests spawn. The requests took a couple of seconds to actually complete, and they were executed sequentially, if that helps.

Originally I had both `on_created` and `on_modified` selected so I thought maybe both were firing, but I deleted that job and created the one with just `on_created` and still saw the duplicate requests.

Also, I took a look at the new scheduler code and it's kind of complex. I know a bunch of that is just APScheduler stuff, but could you give me a quick example of how I can exercise the replacement functionality? I think that would help me a bunch.

TheBurchLog:
@hazmat345 

How were you creating the file? Watchdog goes off OS level events. Some actions will kick off multiple OS level events. For example, if you are using `VIM` to create the file, I believe it is a CREATE when you initialize the file, then another CREATE when the file is saves. When I was testing I used `touch test.txt` as my example.

hazmat345:
I was using `touch`, should I have done something else?

This is kind of a big deal - is there any way we can mitigate this? Does watchdog support some sort of debounce? Or maybe a different observer?

SpecificallyGravity:
@hazmat345 

I have added logic to the custom pattern match class to help with multiple triggers.  

This will ignore triggers from the same source (e.g `/home/user/doc.txt`) and event type (e.g. `created`, `modified`, etc.) that occur within 0.5 seconds of one another **if the job created with it has the `coalesce` flag set**.  This solution won't handle operations that quickly trigger events of different types, just to err on the side of correctness.

As far as I can find, watchdog is the only package that does this sort of thing and it's pretty basic.  Any functionality like this would have to be implemented by us.

hazmat345 (approved):
LGTM

PR #671, opened 2020-10-14
Removing Login Feature

TheBurchLog:
Until we refactor how users and roles are handled in BG. We are just going to remove them completely from the UI.


hazmat345 (commented):
Changes are good here, but the commit message was not accurate and we didn't have a ticket / branch for this. This is especially important since we're going to want to undo these changes at some point and we'll need to be able to find them easily.

I reworded the commit and opened #678, going to close this.

PR #674, opened 2020-10-15
Removing Thrift Tests

TheBurchLog:
This was made to address pytests that were not removed in https://github.com/beer-garden/beer-garden/pull/661


hazmat345 (approved):
So mad I didn't remember to delete these myself. It would have been so cathartic :laughing:

PR #676, opened 2020-10-15
Translating empty string to garden name for SYSTEM_READ_ALL

### LINKED ISSUE #673: Default namespace in systems search API
Currently plugins that use a `SystemClient` are failing if the Plugin creation is not the first statement. The lack of namespace in the global config is causing the system query to fail.

We should probably default to the local garden name on the server side for this case.

hazmat345:
This fixes #673. It translate an empty string namespace filter param for the SYSTEM_READ_ALL operation to the local garden name.

### DIFF
--- a/src/app/beer_garden/router.py
+++ b/src/app/beer_garden/router.py
@@ -365,6 +365,10 @@ def _pre_route(operation: Operation) -> Operation:
         if operation.model.namespace is None:
             operation.model.namespace = config.get("garden.name")
 
+    elif operation.operation_type == "SYSTEM_READ_ALL":
+        if operation.kwargs.get("filter_params", {}).get("namespace") == "":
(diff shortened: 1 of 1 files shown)


TheBurchLog (commented):
Would there ever be a use case where we would expect a filter parameter to be a blank string? If not, we can just drop all empty filter parameters. Then this won't be required.

hazmat345:
@TheBurchLog the concern here is we need a way to distinguish between these cases:
- I want to search for a system with this name / versions across *this specific namespace* (namespace="foo")
- I want to search for a system with this name / version across *all namespaces* (no namespace parameter)
- I want to search for a system with this name / version in the *garden's default namespace* (namespace="")

The last one is what this PR adds, but we still need to be able to support the other two.

PR #679, opened 2020-10-15
Issue/590

### LINKED ISSUE #590: Local Plugin Runner logging

hazmat345:
I believe this fixes #590. It adds a config option that will allow for persisting plugins' STDOUT and STDERR to files in the plugin's working directory.

There are two issues still [outstanding](https://github.com/beer-garden/beer-garden/issues/590#issuecomment-706450395):
- The first is now OBE as the `LockListener` is removed entirely
- The second is addressed as we're no longer delaying logging at all - every messaged is logged as soon as it's received

### DIFF
--- a/src/app/beer_garden/config.py
+++ b/src/app/beer_garden/config.py
@@ -1032,6 +1032,17 @@ _PLUGIN_SPEC = {
                     "previous_names": ["plugins_directory", "plugin_directory"],
                     "alt_env_names": ["PLUGINS_DIRECTORY", "BG_PLUGIN_DIRECTORY"],
                 },
+                "logging": {
+                    "type": "dict",
+                    "items": {
+                        "stream_files": {
+                            "type": "bool",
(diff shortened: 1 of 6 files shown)


TheBurchLog (commented):
This works as expected. Setting the flag starts to push all log updates to those files. When I turned on file logging in `plugin-logging.yaml` both successfully wrote to the logs, when written by logger.

Unexpected outcome:
When running the sleep command,before returning the plugin prints to the logs that it is awake. This was only captured in the new `plugin.out` file and was not captured in the `default.log` file. But it worked perfect in a remote plugin, where the `plugin.out` file did not exist. This could be confusing for people who like to use print statements over the logger.

hazmat345:
@TheBurchLog right, that's a change from old behavior, but that's working exactly as intended. Using print statements almost always a Bad Idea for this exact reason. I think the fix here is to change our example plugins to not use print statements.

Issue #560, opened 2020-08-19
Persist display count selection for datatables instances


hazmat345:
Everywhere we use a datatables table we should persist the count selection in browser local storage.

We should treat each table as distinct. So setting the Requests Index page datatable to "100" should not affect any other page, but if the user navigates away and returns to that page "100" should still be selected.

PR #680, opened 2020-10-15
Storing request index length in local storage

hazmat345:
This starts addressing #560. Only the request index page is done, but the solution should be easy to expand to the other tables.


TheBurchLog (approved):
Works as long as you don't break the session.

hazmat345:
@TheBurchLog not sure what you mean here - what caused it to break?

TheBurchLog:
@hazmat345 
When I meant by breaking the session is going to a new tab or refreshing the page. Our cookies are session based, not browser based.

hazmat345:
@TheBurchLog ahh ok. I'm good with that :smile:

PR #683, opened 2020-10-19
Route all requests that are internally generated to the Router.

TheBurchLog:
This is required for any requests that are internally generated (i.e. Dynamic Command Validation). Now all requests will get sent through the Router class to ensure that the proper Garden is handling the request.


hazmat345 (approved):
LGTM

PR #684, opened 2020-10-20
Removing system_view partial and controller

hazmat345:
This PR removes the system view controller and partial as they're no longer used.


TheBurchLog (approved):
:+1:

PR #690, opened 2020-10-21
Allowing specific host environment variable expansions

### LINKED ISSUE #686: Allow resolving specific host environment vars for local plugins
We currently have no mechanism to allow host environment variable *values* to be injected into a local plugin process. This is prohibited in the general case for security reasons, but we should allow admins to specifically allow certain ones in the application config.

hazmat345:
This fixes #686 for v3, but I'm not super happy about the config name. Thoughts?

### DIFF
--- a/src/app/beer_garden/config.py
+++ b/src/app/beer_garden/config.py
@@ -955,13 +955,6 @@ _PLUGIN_SPEC = {
         "local": {
             "type": "dict",
             "items": {
-                "allowed_env_vars": {
-                    "type": "list",
-                    "items": {"env_var": {"type": "str"}},
-                    "default": [],
-                    "description": "Host environement variables that can be used in a "
-                    "beer.conf",
-                },
                 "auth": {
                     "type": "dict",
                     "items": {
@@ -985,6 +978,13 @@ _PLUGIN_SPEC = {
                     "required": False,
                     "previous_names": ["plugins_directory", "plugin_directory"],
                     "alt_env_names": ["PLUGINS_DIRECTORY", "BG_PLUGIN_DIRECTORY"],
+                },
+                "host_env_vars": {
+                    "type": "list",
+                    "items": {"env_var": {"type": "str"}},
+                    "default": [],
+                    "description": "Host environment variables that will be propagated "
+                    "to local plugin processes",
                 },
                 "logging": {
                     "type": "dict",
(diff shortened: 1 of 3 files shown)


TheBurchLog:
@hazmat345 

I am going back and forth on this one.

For what you wrote to work they both need to include it in the config.yaml AND the beer.conf. That feels like one time too many. Part of me thinks that if an Admin exposes a variable through the config.yaml it should be included in all plugins. BUT if a plugin developer wants to overwrite that variable through their beer.conf, then they should have that option. Or if they want to map an environment variable to something else, they have that option.

Example 1
config.yaml expose ENV FOO=1
Plugin has access to FOO=1

Example 2
config.yaml expose ENV FOO=1
beer.conf assigns ENV FOO=2
Plugin has access to FOO=2

Example 3
config.yaml expose ENV FOO=1
beer.conf assigns ENV BAR=$FOO
Plugin has access to FOO=1
Plugin has access to BAR=1

hazmat345:
@TheBurchLog you should have responded to my email!

My position from my email is - yes, this is a little weird. But I think it's preferable to populating host env vars directly. I really don't like the idea of spreading potentially sensitive info around to all plugins indiscriminately - I think that requiring the plugin developer to take the extra "opt-in" step of putting it in the beer.conf is better overall, even if it's slightly more work for the plugin that needs the value.

TheBurchLog:
@hazmat345 
I guess I should have responded to the email like that....

What you have written works as you described it. I feel like it is overkill, but I get it. Part of me thinks you should only be running Plugins that you trust. So if they want access to stuff, then let them. But I also know my future plans for where I want to go with BG, so we might not want to allow people full access. So I guess lets go with this and we can change it up later.

hazmat345:
@TheBurchLog I hope you know I was joking :smile:

TheBurchLog (commented):
I just gave this a try and it didn't work.

Steps:
Created a plugin with the following command

```Python
    @parameter(
        key="env",
        type="String",
        description="Echo Environment Variable",
        optional=False,
    )
    def env_echo(self, env):
        """Echos ENV Variable"""
        return os.environ.get(env, "")
```

Then ran 
`export TEST_ENV=ABC`

In the config.yaml I added:
`host_env_vars: ["TEST_ENV"]`

When I ran my command with `TEST_ENV` it didn't return `ABC`.

hazmat345:
@TheBurchLog how are you running beergarden? Can you try `logger.warning(os.environ.get("TEST_ENV")` in the application startup and seeing if it works there?

Try running with `TEST_ENV=ABC python -m beer_garden` and seeing how that works.

TheBurchLog:
@hazmat345 

I am running it through PyCharm. When I added it to the Environment Variables pass forward it worked.

PR #693, opened 2020-10-22
Separate logging config for local plugins

### LINKED ISSUE #636: File logging for plugins
We may need to move to two separate plugin logging configs. The rpm install creates a plugin logging config file that logs to a file in the `/opt/beer-garden/log` directory (to match v2 behavior).

This is a problem for remote plugins on another system that may not have permission to log to that directory.

hazmat345:
Fixes #636, depends on beer-garden/brewtils#255

### DIFF
--- a/src/app/beer_garden/api/http/handlers/v1/logging.py
+++ b/src/app/beer_garden/api/http/handlers/v1/logging.py
@@ -14,6 +14,13 @@ class LoggingAPI(BaseHandler):
         description: |
           Will return a Python logging configuration that can be used to configure
           plugin logging.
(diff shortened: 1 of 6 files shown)


TheBurchLog (commented):
We are monitoring the default plugin logger for file changes. (app.py around line 210). That way the don't have to reload Beer Garden to change the plugin logger. Should we include that functionality for this as well?

hazmat345:
@TheBurchLog oh, good point. I'll add that.

TheBurchLog (approved):
LGTM

PR #700, opened 2020-10-28
Issue/695 Improved Encapsulation for Stomp's Connection Class

### LINKED ISSUE #695: Stomp Connection
I think the organization of the `beer_garden.api.stomp.server.Connection` class can be improved a little. Right now it depends on the module-level `conn` and `bg_active` variables, which has a couple of issues:

- Having them be module-level signals that they're useful things that I can import and use somewhere else. However, I don't think this is the case... only `Connection` is used outside of the module.
- More importantly, it means that there can only ever be one instance of `Connection` created, since if I try to create another the values of `conn` and `bg_active` set by the first connection are going to be overwritten.

Having better encapsulation here would really help with code reuse. I think it should be possible to have the two classes (`OperationListener` and `Connection`) be the only things in the module scope. Assigning things to [instance attributes](https://dzone.com/articles/python-class-attributes-vs-instance-attributes) instead of module or class attributes should be helpful here.

west270:
I improved encapsulation for stomp's connection class and made some minor changes to the entry config for stomp to make more clear what is for ssl.

Fixes #695

### DIFF
--- a/docker/docker-compose/docker-compose.yml
+++ b/docker/docker-compose/docker-compose.yml
@@ -8,7 +8,7 @@ x-definitions: &env
     # IMPORTANT!!
     # Change this to support both local and remote plugins
     # Should be resolvable from inside and outside the docker network
-    BG_MQ_HOST: rabbitmq
+    BG_MQ_HOST: localhost
 
     # IMPORTANT!!
     # If you are connection to a Parent BG instance, toggle true
--- a/docker/docker-compose/docker-compose.yml
+++ b/docker/docker-compose/docker-compose.yml
@@ -8,12 +8,12 @@ x-definitions: &env
     # IMPORTANT!!
     # Change this to support both local and remote plugins
     # Should be resolvable from inside and outside the docker network
-    BG_MQ_HOST: localhost
+    BG_MQ_HOST: rabbitmq
 
     # IMPORTANT!!
     # If you are connection to a Parent BG instance, toggle true
     BG_PARENT_HTTP_ENABLE: "false"
-
+    BG_ENTRY_STOMP_ENABLED: "true"
     # IMPORTANT !!
     # If you are connecting to a Parent BG instance, change this the resolvable DNS entry
     BG_PARENT_HTTP_HOST: localhost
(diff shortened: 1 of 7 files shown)


src/app/beer_garden/api/stomp/__init__.py
```diff
@@ -22,11 +22,18 @@
 def run(ep_conn):
     global conn
```
TheBurchLog, line 23: Any way we can avoid global variables like this? That way it is easier to reuse.

src/app/beer_garden/api/stomp/stomp_manager.py
```diff
@@ -0,0 +1,74 @@
...
+logger = logging.getLogger(__name__)
+stop_thread = False
+
+
+def shutdown():
+    global stop_thread
+    stop_thread = True
+
+
+class StompManager:
```
hazmat345, line 20: This is the one area I think can be simplified even more. Having a thread that runs for the life of the application and needs a way to shutdown gracefully is a pattern we reuse a lot, so much so that we already have an implementation for it.

I think this `StompManager` is a good candidate to inherit from `brewtils.stoppable_thread.StoppableThread`. Doing that should allow you to remove the `stop_thread` global.

hazmat345 (approved):
This looks really good to me. Only one area where I think things could be more consistent with how the rest of the application works, but even without that I'm good with this.

I didn't retest this, so I'm assuming it all still works correctly :smile:

docker/docker-compose/docker-compose.yml
```diff
@@ -8,12 +8,12 @@ x-definitions: &env
     # IMPORTANT!!
     # Change this to support both local and remote plugins
     # Should be resolvable from inside and outside the docker network
-    BG_MQ_HOST: localhost
+    BG_MQ_HOST: rabbitmq
```
hazmat345, line 11: I know this is annoying, but we have to keep this as `localhost`. Changing this to `rabbitmq` means that local plugins will work "out of the box", but remote plugins will break. On the other hand, having it be `localhost` means that remote plugins will work but local plugin will break.

There's no great way to have both work out of the box, unfortunately. And since our docs / tutorials are written for remote plugins it's better to optimize for that case.

hazmat345 (approved):
One reversion, other than that LGTM :+1:

Issue #701, opened 2020-10-29
Entry Point early-stage logging


hazmat345:
We don't handle errors in very early-stage entry point startup well. Demonstrated by an `AttributeError` when running the stomp entry point:

```
beer-garden_1  | Exception in thread Application:
beer-garden_1  | Traceback (most recent call last):
beer-garden_1  |   File "/usr/local/lib/python3.7/threading.py", line 926, in _bootstrap_inner
beer-garden_1  |     self.run()
beer-garden_1  |   File "/usr/local/lib/python3.7/site-packages/beer_garden/app.py", line 119, in run
beer-garden_1  |     self._startup()
beer-garden_1  |   File "/usr/local/lib/python3.7/site-packages/beer_garden/app.py", line 208, in _startup
beer-garden_1  |     self.entry_manager.create_all()
beer-garden_1  |   File "/usr/local/lib/python3.7/site-packages/beer_garden/api/entry_point.py", line 231, in create_all
beer-garden_1  |     self.entry_points.append(self.create(entry_name))
beer-garden_1  |   File "/usr/local/lib/python3.7/site-packages/beer_garden/api/entry_point.py", line 234, in create
beer-garden_1  |     module = import_module(f"beer_garden.api.{module_name}")
beer-garden_1  |   File "/usr/local/lib/python3.7/importlib/__init__.py", line 127, in import_module
beer-garden_1  |     return _bootstrap._gcd_import(name[level:], package, level)
beer-garden_1  |   File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
beer-garden_1  |   File "<frozen importlib._bootstrap>", line 983, in _find_and_load
beer-garden_1  |   File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
beer-garden_1  |   File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
beer-garden_1  |   File "<frozen importlib._bootstrap_external>", line 728, in exec_module
beer-garden_1  |   File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
beer-garden_1  |   File "/usr/local/lib/python3.7/site-packages/beer_garden/api/stomp/__init__.py", line 12, in <module>
beer-garden_1  |     from beer_garden.api.stomp.server import Connection
beer-garden_1  |   File "/usr/local/lib/python3.7/site-packages/beer_garden/api/stomp/server.py", line 13, in <module>
[...]

PR #702, opened 2020-10-29
Better entry point isolation during creation

hazmat345:
Helps with #701, should at least result in the app starting even if one entry point fails during creation.


hazmat345:
@TheBurchLog I'm going to just merge this - it's pretty simple and I want to make sure it logs the current error we're seeing in docker.

PR #703, opened 2020-10-30
Issue/646

### LINKED ISSUE #646: System delete sleep
The system purge function currently uses a sleep to wait for instances to stop. This is not good.

hazmat345:
Fixes #646. Reorganizes the local plugin manager so key info is available in entry point processes as well as the main process.

Depends on beer-garden/brewtils#257.

### DIFF
--- a/src/app/beer_garden/api/entry_point.py
+++ b/src/app/beer_garden/api/entry_point.py
@@ -1,27 +1,26 @@
 # -*- coding: utf-8 -*-
+from importlib import import_module
+
 import logging
 import multiprocessing
 import signal
-from importlib import import_module
+from box import Box
+from brewtils.models import Event
 from multiprocessing.connection import Connection, Pipe
 from multiprocessing.context import SpawnContext
 from multiprocessing.queues import Queue
 from types import FrameType
-from typing import Any, Callable, TypeVar
-
-from box import Box
-from brewtils.models import Event
+from typing import Any, Callable
 
 import beer_garden
 import beer_garden.config
 import beer_garden.db.api as db
 import beer_garden.events
+import beer_garden.local_plugins.manager as lpm
 import beer_garden.queue.api as queue
 import beer_garden.router as router
 from beer_garden.events.processors import PipeListener, QueueListener
 from beer_garden.log import process_record
-
-T = TypeVar("T", bound="EntryPoint")
 
 logger = logging.getLogger(__name__)
 
@@ -92,6 +91,7 @@ class EntryPoint:
                 self._target,
                 self._signal_handler,
                 self._ep_conn,
+                lpm.lpm_proxy,
             ),
         )
         self._process.start()
@@ -120,7 +120,7 @@ class EntryPoint:
         self._process.join(timeout=timeout)
 
         if self._process.exitcode is None:
-            self._logger.warning(
+            logger.warning(
                 f"Process {self._process.name} is still running - sending SIGKILL"
             )
             self._process.kill()
@@ -143,6 +143,7 @@ class EntryPoint:
         target: Callable,
         signal_handler: Callable[[int, FrameType], None],
         ep_conn: Connection,
+        lpm_proxy,
     ) -> Any:
         """Helper method that sets up the process environment before calling `target`
 
@@ -161,6 +162,7 @@ class EntryPoint:
             target:
             signal_handler:
             ep_conn:
+            lpm_proxy:
 
         Returns:
             The result of the `target` function
@@ -178,6 +180,9 @@ class EntryPoint:
         try:
             # Also set up plugin logging
             beer_garden.log.load_plugin_log_config()
+
+            # Local plugin manager reference
+            lpm.lpm_proxy = lpm_proxy
 
             # Set up a database connection
             db.create_connection(db_config=beer_garden.config.get("db"))
@@ -234,7 +239,7 @@ class Manager:
                 except Exception as ex:
                     logger.exception(f"Error creating entry point {entry_name}: {ex}")
 
-    def create(self, module_name: str) -> T:
+    def create(self, module_name: str):
         module = import_module(f"beer_garden.api.{module_name}")
 
         return EntryPoint(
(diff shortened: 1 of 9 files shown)


src/app/beer_garden/local_plugins/manager.py
```diff
@@ -7,22 +7,24 @@
...
 
-import beer_garden
 import beer_garden.config as config
 from beer_garden.errors import PluginValidationError
 from beer_garden.local_plugins.env_help import expand_string
 from beer_garden.local_plugins.runner import ProcessRunner
 
+# This is ... complicated. See the PluginManager docstring
+lpm_proxy = None
+
```
TheBurchLog, line 27: I get why you are placing this in the Plugin Manager but it is only utilized by the App. Should we be placing this here just for a global processor variable?

src/app/beer_garden/api/entry_point.py
```diff
@@ -179,6 +181,9 @@ def _target_wrapper(
             # Also set up plugin logging
             beer_garden.log.load_plugin_log_config()
 
+            # Local plugin manager reference
+            lpm.lpm_proxy = lpm_proxy
+
```
TheBurchLog, line 186: What is the benefit of passing forward lmp_proxy instead of connecting to it?

```python
from multiprocessing.managers import BaseManager
lpm.lpm_proxy=  BaseManager(address=('', 5000), authkey='abc')
lpm.lpm_proxy.connect()
```

src/app/beer_garden/app.py
```diff
@@ -93,6 +95,22 @@ def initialize(self):
...
+                log_dir=config.get("plugin.local.log_directory"),
+                connection_info=config.get("entry.http"),
+                username=config.get("plugin.local.auth.username"),
+                password=config.get("plugin.local.auth.password"),
+            ),
+        )
+
+        data_manager = BaseManager()
+        data_manager.start()
+        beer_garden.local_plugins.manager.lpm_proxy = data_manager.PluginManager()
```
TheBurchLog, line 112: When initializing the Base Manager, why didn't you expose it with an address object? Because if you did, then we could have any where in our code initialize this class?

```python
data_manager = BaseManager(address=('', 50000)
```

src/app/beer_garden/app.py
```diff
@@ -133,7 +151,20 @@ def handle_event(event):
...
+            elif event.name == Events.INSTANCE_STARTED.name:
+                beer_garden.local_plugins.manager.lpm_proxy.handle_start(event)
+            elif event.name == Events.INSTANCE_STOPPED.name:
+                beer_garden.local_plugins.manager.lpm_proxy.handle_stop(event)
+            elif event.name == Events.SYSTEM_REMOVED.name:
+                beer_garden.local_plugins.manager.lpm_proxy.remove_system(event.payload)
+            elif event.name == Events.SYSTEM_RELOAD_REQUESTED.name:
+                beer_garden.local_plugins.manager.lpm_proxy.reload_system(event.payload)
+            elif event.name == Events.SYSTEM_RESCAN_REQUESTED.name:
+                beer_garden.local_plugins.manager.lpm_proxy.scan_path()
```
TheBurchLog, line 167: I don't know why this got pulled up. Why can't this stay with the Plugin Manager as a registered listener for events?

src/app/beer_garden/app.py
```diff
@@ -257,10 +288,10 @@ def _shutdown(self):
             self.scheduler.shutdown(wait=False)
 
         self.logger.debug("Stopping local plugin process monitoring")
-        PluginManager.instance().stop()
+        beer_garden.local_plugins.manager.lpm_proxy.stop()
 
         self.logger.debug("Stopping local plugins")
-        PluginManager.instance().stop_all()
+        beer_garden.local_plugins.manager.lpm_proxy.stop_all()
```
TheBurchLog, line 294: Again, only this class references lpm_proxy, so it feels more like only this class should use it.

TheBurchLog (commented):
Overall I think this is really cool. I have some ideas on how we could utilize this in other avenue, but those comments will be saved for later.

My big concern is that if we are going to use this, then let's not pass the object around in memory to the sub-processes. In the sub-process, just have them create a connection to the object. We just need to reserve a super high internal port for this like 50000. Then we can really use this class as it was intended.

hazmat345:
@TheBurchLog yeah, this isn't really useful yet, but I wanted you to take a look at it today. Most of your comments are things that I either explored earlier or that I wanted to add after I got the core functionality working. There's definitely still work that needs to be done here, so I'm going to hold off on addressing your comments until I have a better idea of the best way forward.

src/app/beer_garden/app.py
hazmat345: Because the Plugin Manager now lives in its own process, and that process doesn't have any plumbing to listen for events. But this is still a WIP, a lot of this logic is going to change now (the plugin manager shouldn't need to care about events for the most part anymore).

src/app/beer_garden/app.py
hazmat345: Yeah, I looked at that and it should work. I wasn't able to get it working yet, though.

src/app/beer_garden/api/entry_point.py
hazmat345: I couldn't get connecting to it working yet :upside_down_face:

src/app/beer_garden/local_plugins/manager.py
hazmat345: It's only used in the App because this PR is just the groundwork for the real work that I'm doing. The main purpose of this refactor is so that other modules can now get access to *something* that let's the interact with the local plugin state.

src/app/beer_garden/app.py
hazmat345: That's going to change

hazmat345:
I'm going to go ahead and merge this. I took another (quick) look at the server / client aspect of `BaseManager` but still can't get it working. As far as I can tell there's nothing wrong with passing the proxy instance around. We can revisit if that ends up not being the case.

PR #723, opened 2020-11-09
Fixed Scheduler for Dates

TheBurchLog:
The APScheduler was not getting passed the Trigger Object when the event was being created, so this now creates the trigger object and passes it forward. Now Date Triggers will display the proper next fire date and fire when expected.


hazmat345 (approved):
:+1:

Issue #727, opened 2020-11-10
Can Create Two Gardens with same name through the API


west270:
When working on issue #726, I was able to make two gardens with the same name.

The UI protects against creating a garden with a duplicate name, but there is currently no validation on the api side. We should add a unique constraint at the database model level, which should translate to the right HTTP status code getting returned when someone attempts to create a duplicate garden name.

scott-taubman:
I'll put this comment here, though maybe it deserves its own ticket...

Related to Garden naming, I've noticed that you can create garden names with non-word characters, like "-", but then you can never get them again via the API.  So I can create a Garden named "who-loves-hyphens", but then if I try to go to `/api/v1/gardens/who-loves-hyphens/` I get an immediate 404, presumably because that path doesn't match any of the defined routes, which look for word characters (`\w`) only.

Presumably we should either:
  * Fix the routing so that you can reach endpoints for all legal garden names
  * **or** validate that the garden name contains only word characters before we create it

The reason why this may be appropriate as one issue is because the broader problem appears to be garden name validation, which is maybe just non-existent today?  At a minimum, it needs some attention.

hazmat345:
@scott-taubman I think #940 is tracking that issue.

scott-taubman:
Looks like you're right. URL encoding works as it should. We'll address that in #940 then.

scott-taubman:
A possibly ignorant follow-up question...What's the theoretical downside to just being less strict in the routing regex?  Why only route stuff that matches `[\w%]+`, as opposed to routing whatever is supplied? Worst case should be that they get a 404 back right?  Granted, it's more work to do the lookup to return that 404, but the upside is that you don't have to codify the acceptable garden name format in another place.

PR #728, opened 2020-11-10
Enables creating a garden through ui issue #726

### LINKED ISSUE #726: Need ability to create garden through ui
Creating garden through ui is need for parent and child communication, since parent cannot get child garden information without having an active stomp listener.

west270:
Enables creating a garden through ui.
Found new issue #727 while testing.
Fixes #726

### DIFF
--- a/src/ui/src/js/controllers/admin_garden.js
+++ b/src/ui/src/js/controllers/admin_garden.js
@@ -19,13 +19,15 @@ export default function adminGardenController(
     GardenService,
     EventService) {
   $scope.setWindowTitle('gardens');
-
+  $scope.gardenCreateSchema = GardenService.CreateSCHEMA;
+   $scope.gardenCreateForm = GardenService.CreateFORM;
   $scope.successCallback = function(response) {
     $scope.response = response;
     $scope.data = response.data;
 
   };
-
+  $scope.garden_name = null;
+  $scope.createGardenFormHide = true;
   $scope.failureCallback = function(response) {
     $scope.response = response;
     $scope.data = [];
@@ -40,6 +42,19 @@ export default function adminGardenController(
 
   $scope.syncGardens = function() {
     GardenService.syncGardens()
+  }
+
+  $scope.createGarden = function() {
+      if ($scope.garden_name != "" & $scope.garden_name != null){
+        GardenService.createGarden({"name":$scope.garden_name, "status":"NOT_CONFIGURED"});
+        $state.go('base.garden_view',
+              {
+                'name': $scope.garden_name,
+              }
+            );
+      }
+      $scope.createGardenFormHide = true;
+
   }
 
   $scope.editGarden = function(garden) {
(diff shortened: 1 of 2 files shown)


src/ui/src/js/controllers/admin_garden.js
```diff
@@ -19,13 +19,15 @@ export default function adminGardenController(
...
-
+  $scope.gardenCreateSchema = GardenService.CreateSCHEMA;
+   $scope.gardenCreateForm = GardenService.CreateFORM;
   $scope.successCallback = function(response) {
     $scope.response = response;
     $scope.data = response.data;
 
   };
-
+  $scope.garden_name = null;
```
TheBurchLog, line 29: This is too generic for this page. The who page is to manage Gardens and having a field called `garden_name` gives no uniqueness to what the value is actually doing. Give this a more meaningful name.

src/ui/src/partials/admin_garden_index.html
```diff
@@ -1,14 +1,24 @@
...
   </span>
 </h1>
-  
+<div class="pull-right">
+  <div class="form-popup" id="myForm" ng-hide=createGardenFormHide>
+    <label><b>Garden Name</b></label>
+    <input type="text" placeholder="Enter Garden Name" name="garden_name" required ng-model="garden_name">
+
+    <button type="button" class="btn" ng-click="createGarden()">Create</button>
+    <button type="button" class="btn" ng-click="createGardenFormHide = true">Close</button>
```
TheBurchLog, line 16: On close, we should clear the value being set by the form. That way the user gets a clean form if they reopen it.

src/ui/src/js/controllers/admin_garden.js
```diff
@@ -42,6 +44,19 @@ export default function adminGardenController(
...
+  $scope.createGarden = function() {
+      if ($scope.garden_name != "" & $scope.garden_name != null){
+        GardenService.createGarden({"name":$scope.garden_name, "status":"NOT_CONFIGURED"});
+        $state.go('base.garden_view',
+              {
+                'name': $scope.garden_name,
+              }
+            );
+      }
+      $scope.createGardenFormHide = true;
```
TheBurchLog, line 56: Does this actually do anything? If this condition is hit, then the field was not populated, so isn't that an error condition? You should do an else statement and throw an validation alert that they didn't provide an Garden Name for the create function. Also, don't hide the form during the failure case.

TheBurchLog (commented):
Once you address me comments this should be good ot go

west270:
make lint:
```
flake8 beer_garden test
black --check beer_garden test
All done! ✨ 🍰 ✨
130 files would be left unchanged.
```


make test:
```
============================================================================================================================= warnings summary =============================================================================================================================
test/db/mongo/pruner_test.py: 12 warnings
  /bg_venv/venv/lib64/python3.7/site-packages/mongoengine/queryset/visitor.py:156: DeprecationWarning: 'empty' property is deprecated in favour of using 'not bool(filter)'
    warnings.warn(msg, DeprecationWarning)

-- Docs: https://docs.pytest.org/en/stable/warnings.html
============================================================================================================== 327 passed, 262 skipped, 12 warnings in 10.47s ==============================================================================================================
```

TheBurchLog (approved):

PR #746, opened 2020-11-16
Suppress duplicate Event processing for Requests

TheBurchLog:
When an event is sent to Beer Garden it is forwarded to all processors (Main App, REST, and STOMP). Each processor has an instance of the Requests library listening for transactions. This was causing a race condition where both processors attempted to create the record and one would fail. This does not cause any logistical errors for the user because the record is still created, but it does throw an error in the logs. So this will suppress that error message and provide some future proofing for events not showing up for Requests.

### DIFF
--- a/src/app/beer_garden/requests.py
+++ b/src/app/beer_garden/requests.py
@@ -17,7 +17,7 @@ import pika.spec
 import six
 import urllib3
 
-
+from beer_garden.errors import NotUniqueException
 from brewtils.choices import parse
 from brewtils.errors import ConflictError, ModelValidationError, RequestPublishException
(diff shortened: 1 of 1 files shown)


hazmat345 (commented):
So I think this kind of hints at a deeper underlying problem: our event handling is duplicated in places (processes, actually) where it shouldn't be.

I'm OK with a stopgap to get this error out of the logs as long as this doesn't close the issue.

Instead of doing another db query here can we just catch the `NotUniqueException` in the operation we're already doing?

hazmat345:
I'm going to close this as it's OBE by #911. The underlying issue is not fix and should remain open.

PR #755, opened 2020-11-19
Issue/555

### LINKED ISSUE #555: Use local storage for themes
Using local storage opens up a couple options with the UI that could consider. For example, keeping state on the Theme the user has selected. In this PR could you also add Theme tracking?

A more advance would be allowing users to favorite commands. We should chat about this feature and discuss if it is a good idea, and if so, make a new PR for it.

_Originally posted by @TheBurchLog in https://github.com/beer-garden/beer-garden/pull/547#issuecomment-672830025_

hazmat345:
This fixes #555. Was originally PR #558.

### DIFF
--- a/src/ui/src/js/run.js
+++ b/src/ui/src/js/run.js
@@ -91,8 +91,14 @@ export default function appRun(
         // coalescePermissions [0] is roles, [1] is permissions
         user.permissions = RoleService.coalescePermissions(user.roles)[1];
 
-        let theme = _.get(user, 'preferences.theme', 'default');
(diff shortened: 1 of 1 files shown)


TheBurchLog:
Converted this to draft because we need to figure out if this is the proper way to handle this.

hazmat345:
Just squashed and rebased this onto develop. Let's either merge this or kill the  branch.

hazmat345:
@scott-taubman did we make a decision on this?

west270 (approved):

PR #757, opened 2020-11-20
Stomp parent and child communication

### LINKED ISSUE #761: Stomp Communication Between Parent and Child
Add ability for parent and child to communicate using Stomp.

west270:
~~If the parent has stomp entry enabled, then the child can send to the parent subscription destination. 
If the parent does not have stomp entry enabled, then a new garden needs to be created through the ui and the parent will spawn a new entry point to listen for child messages. In both cases the connection needs to be configured through the ui.~~ 
The child cannot listen to the `send_destination` from the parents `entry.stomp.send_destination` because that will cause the child and parent to loop. Instead the child needs to listen to a different destination which will be configured in the parent gardens ui.
For testing I was creating the child garden using swagger, since currently there is no way to add a garden through ui yet.

Edit:
Changed code so it creates new stomp connections in the `stomp_manager.py`, and rolled back the `entry_point.py`. On start up beer garden creates a stomp entry but will not make any connections if stomp is not enabled in config. This allows connections to children over stomp without enabling stomp in the config.

Fixes #761


src/app/beer_garden/api/entry_point.py
```diff
@@ -118,9 +124,11 @@ def stop(self, timeout: int = None) -> None:
         # Then ensure the process is terminated
         self._process.terminate()
         self._process.join(timeout=timeout)
-
+        if closeCommunicationPipes:
+            self._ep_conn.close()
+            self._mp_conn.close()
         if self._process.exitcode is None:
-            logger.warning(
+            self._logger.warning(
```
hazmat345, line 131: I think this needs to be reverted.

src/app/beer_garden/config.py
```diff
@@ -942,7 +942,85 @@ def _parse_args(args: Sequence[str]) -> Tuple[YapconfSpec, dict]:
...
+                },
+                "port": {
+                    "type": "int",
+                    "default": 61613,
+                    "description": "Serve content on this port",
+                },
+                "host": {
+                    "type": "str",
+                    "default": "0.0.0.0",
+                    "description": "Host for the HTTP Server to bind to",
```
hazmat345, line 962: These descriptions aren't accurate.

src/app/beer_garden/api/entry_point.py
```diff
@@ -92,14 +97,15 @@ def start(self) -> None:
...
+                self.ep_config,
             ),
         )
         self._process.start()
 
         # And listen for events coming from it
         self._event_listener.start()
 
-    def stop(self, timeout: int = None) -> None:
+    def stop(self, timeout: int = None, closeCommunicationPipes=False) -> None:
```
hazmat345, line 108: Please don't use CamelCase for things. See https://stackoverflow.com/a/159745 (specifically https://www.python.org/dev/peps/pep-0008/#function-and-variable-names)

hazmat345 (changes requested):
I've make a couple comments on things, mostly surface-level stuff.

This PR is too complex to approve without a more detailed explanation or a conversation about what it's trying to do and how it works. I'm particularly concerned with the changes to entry point.

hazmat345:
@west270 I've been trying to get this to work for a while and haven't been successful. I'm trying to just create a request on a parent garden using the `StompBeerGardenClient` in the dev scripts repo but it doesn't seem to work anymore. Could you take a look?

hazmat345 (approved):
This looks good, but please run the formatter to fix the linting errors.

TheBurchLog (approved):
Overall, it worked as expected. Just a couple minor things to change. More for our quality of life and not functionality changes. Also, we need and example on how to set this up. I don't know where we want to put that, but we need it somewhere.

src/app/beer_garden/api/stomp/server.py
```diff
@@ -61,21 +76,24 @@ def on_error(self, headers, message):
         logger.warning("received an error:" + str(headers))
 
     def on_message(self, headers, message):
+        if headers.get("model_class") != "Operation":
+            return
```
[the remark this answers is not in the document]
west270: I removed the if statement since the try except should catch any error when an object received is not an operation, since we are assuming all messages received are operations.

Issue #770, opened 2020-11-26
rpm install bad shutdown


hazmat345:
Still seeing bad shutdowns (defunct processes) with the rpm install.

hazmat345:
Looks like attempting to publish a stop message that was rejected by the broker during shutdown caused this.

hazmat345:
Now it's looking like the HTTP entry point is shutting down before it's supposed to?

hazmat345:
I believe this is fixed in 3.0.1, closing

PR #781, opened 2020-12-05
Using SIGUSR1 for stopping entry points

hazmat345:
This is another part of #770. This PR prevents a premature termination of entry point processes when running under systemd.


TheBurchLog (approved):
:+1:

Issue #771, opened 2020-11-30
Force System Delete Failure


hazmat345:
Trying to delete a system from the admin page fails if the Stop message publishing fails. That's OK, but the force delete also fails with the same message.

PR #788, opened 2020-12-10
Issue/708

### LINKED ISSUE #708: System Admin view unassociated runners
Our local plugin management story is not good for the case where a plugin fails to associate. The v3 changes to the local plugin runner / manager make it impossible to know which system a particular runner is associated with until the plugin actually registers.

This means that if the plugin *never* registers then we won't know which system that runner is associated with, so we aren't able to present it in the system admin page. Also, the rescan functionality doesn't work since that directory is already registered, so the rescan won't attempt to start it again.

This will be fixed when it's possible to verify and control hung and dead plugins. It should be possible to see *something* that indicates the existence and status of local plugins like the following:

#### Hung Plugin
```python
def main():
    while True:
        pass
```

#### Dead Plugin
```python
def main():
    return
````

Users should be able to stop, start, reload, and remove *all* local plugins, *including the above*.

### PR DESCRIPTION
**Pull request by hazmat345**
Fixes #708. Depends on beer-garden/brewtils#271.

There's a lot here, so here are the highlights:

- There's now a separate API for runner-related interactions
- The system admin UI now has a new runner section

There have also been some factoring changes. Not all of them are complete, but these are the ideas:

- There are methods at the module level of `beer_garden.local_plugins.manager`. These are the new public interface for that module - you shouldn't be calling methods on the `lpm_proxy` anymore (this allows using the event decorators on those methods).
- The app no longer sends a Stop request to every local plugin when shutting down. Instead, the local plugin stop procedure will send a SIGTERM as the primary way of stopping local runners that aren't already stopped. However, this is mostly redundant, because:
- When shutting down, it's possible for the shutdown signal sent to the application process to be propagated to the local plugin processes as well (something something process groups ... this is a *very* deep rabbit hole). To handle this correctly we now check each time an Instance is updated. If the update is setting the status to "STOPPED" we will attempt to mark the corresponding runner as STOPPED as well (if the instance is from a remote plugin this just does nothing). Previously we did this in `plugin.stop` (so at the *start* of the stop procedure, rather than at the *end*)
- For event publishing, the location of the `try` has been moved out of the decorator. This allows `publish` to be called directly, which let's me handle situations where I need to publish an array of things.
- The `rescan` and `reload` functionality that was previously associated with Systems is now associated with Runners. This is a better logical fit as those operations / concepts only make sense in the context of local plugins (aka, runners).

Still not quite done:
- Still need to determine what makes the most sense to display in the UI to indicate the *status* of a runner
- Misc cleanup / more refactoring. This should all *work*, but I need to go though and clean it.

### DIFF
--- a/src/ui/src/partials/admin_system.html
+++ b/src/ui/src/partials/admin_system.html
@@ -23,6 +23,39 @@
   {{alert.msg}}
 </div>
 
+<h3>Unassociated Runners</h3>
+<div class="panel-container animate-if"
+     ng-if="responseState(response) === 'success'">
+
+  <div ng-repeat="runner in runners | filter:unassociatedRunners"
+      class="container-item panel panel-default"
+      style="display: block;">
+
+    <div class="panel-heading" style="font-size: 22px;">
+      <span>{{runner.runner_path}}</span>
+    </div>
+
+    <div class="panel-body" style="padding-bottom: 0px;">
+      <span>{{runner.runner_path}}</span>
+      <span>{{runner.runner_id}}</span>
+      <span>{{runner.instance_id}}</span>
+
+      <div class="btn-group btn-group-sm pull-right"
+            ng-show="hasPermission(user, 'bg-system-admin')">
+        <button type="button" class="btn btn-default fa fa-trash-o"
+                ng-click="deleteRunner(runner.runner_id)"
+                title="Delete Runner {{runner.runner_path}}">
+        </button>
+      </div>
+    </div>
+  </div>
+
+  <!-- This is ugly but empty panels ensure the bottom row is left-aligned -->
+  <div class="container-item empty"></div>
+  <div class="container-item empty"></div>
+</div>
+
+<h3>Systems</h3>
 <div class="panel-container animate-if"
      ng-if="responseState(response) === 'success'">
 
--- a/src/ui/src/partials/admin_system.html
+++ b/src/ui/src/partials/admin_system.html
@@ -7,6 +7,10 @@
             confirm="Are you sure you want to delete all Queues?">
       Clear All Queues
     </button>
+    <button type="button"
+            class="btn btn-primary"
+            ng-show="hasPermission(user, 'bg-system-admin')"
+            ng-click="rescan()">Rescan Plugin Directory</button>
   </span>
 </h1>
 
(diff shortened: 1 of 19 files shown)

### CODE REVIEW

File: src/ui/src/partials/admin_system.html [hunk 1]
```diff
@@ -23,8 +23,79 @@ <h1 class='page-header'>
...
+        style="display: block;">
+
+      <div class="panel-heading" style="font-size: 22px;">
+        <span>{{runners[0].path}}</span>
+
+        <div class="btn-toolbar pull-right">
+          <div class="btn-group btn-group-sm">
+            <button type="button" class="btn btn-default fa fa-play"
+                    ng-show="hasPermission(user, 'bg-system-admin')"
+                    ng-click="startSystem(system)"
```
TheBurchLog, line 45: If this works I will be surprised. I don't see the ng model for system here.

TheBurchLog (commented):
Make sure you try this with REST, UI, and STOMP.

src/ui/src/partials/admin_system.html
hazmat345: It won't. This is why it's in draft 😄

hazmat345:
To test, start with an existing local plugin. Remove it from the systems admin page or drop the systems collection in the database (aka make sure Beer Garden doesn't know about it already).

Then add an infinite loop to the `main`:

```python
def main():
    while True:
        pass

    plugin = Plugin(name="sleeper", version="1.0.0.dev0")
    plugin.client = SleeperClient()
    plugin.run()
```

Now start Beer Garden.

*Something* should be shown on the systems admin page that indicates the existence of this plugin.

Now, *without stopping Beer Garden*, remove the infinite loop (make it a correct plugin again).

You should be able to stop and start the plugin and see it register successfully. You should also be able to remove it and rescan the plugin directory and still see it register successfully.

TheBurchLog:
@hazmat345 I didn't like the UI and had an idea for it. I went ahead and made those changes.

TheBurchLog (approved):
I am not going to review the UI code because of my changes there, but the APP code looks good.

west270 (approved):
Everything in the UI works as I expected and functions  in ```router.py``` for runners work stomp messages also. 
I would suggest adding some way to indicate that there is no runner for a system because if the runner is deleted then the run button in the corresponding system will not work until plugins are rescanned.

TheBurchLog:
I forgot to add back into the UI a way to show runners for systems that 100% are unknown. So they now appear at the end.

hazmat345:
So I like the general idea, and I considered doing something like this. However, the issue here (and correct me if I'm wrong) is these changes make the assumption that the *runner name* will match the *system name*. This assumption holds for our test plugins, but it does not hold in practice where the typical plugin name will be `systemname-version`.

I tested this scenario and it looks like the runners don't show up properly. It's also going to be hard to know which runner goes with which instance if there are multiple versions of the same system.

To see what I mean, put two plugins in your local plugins directory:

```
sleeper-1.0.0
sleeper-2.0.0
```

Having the runners be a completely separate section sidesteps this issue. The runner name (the panel title) is going to be whatever the directory is named (sleeper-1.0.0) which is enough information for people to figure out what's going on.

west270:
@hazmat The ui is not showing runners. Here is what I see on the ui:
image

west270:
@hazmat nvm I realized what is doing now it only shows runners with the while loop error. LGTM

west270 (approved):
LGTM other than being unable to a delete a system that has the while loop as described above.

TheBurchLog (approved):
I don't have any changes. I verified that the Runner Tile only appears when there is an unknown runner, so that is good. It looks like it kills the runners correctly in that state.

hazmat345:
Hey @west270, when you say

> LGTM other than being unable to a delete a system that has the while loop as described above.

Are you trying to delete the actual system (something other than the runner in the red panel)? I just tried that and it fails, but I think that's a different issue (#771)

west270:
@hazmat345 I think it is related to the issue #771. I was trying to delete the actual system.

PR #791, opened 2020-12-17
Admin Systems Page Redesign 

### LINKED ISSUE #603: Update Systems Admin Page
The System Admin Page needs to be redesigned. If you have a system that is deployed on multiple namespaces or has a ton of instances it gets uneven on the page. So we should find a way to clean this up and present it better to the Admins.

west270:
~Changes admin system page to use column count instead of flex-boxes, so when a system has multiple instances the page will only have white space at the end of the columns.~
Redesigned system admin page into a table with collapsible rows. 
Fixes #603


hazmat345 (commented):
I don't know that this is better... I feel like ordering alphabetically by column is worse since "normal" reading is left-to-right, not top-to-bottom. It also doesn't solve the gaps issue:

image

I don't think this closes the linked issue - this change is a layout tweak, not a redesign.

west270:
Closing this pr since it will be looking at redesign for react ui

Issue #794, opened 2020-12-18
Remote Plugins Unable to Delete


TheBurchLog:
We are unable to delete a remote plugin when they are turned off

PR #795, opened 2020-12-18
Fixes delete of Remote Systems

TheBurchLog:
Had two options to check, one was check the LPM or check the Metadata. Opted to check the LPM just in case another Beer-Garden is managing the Runner and populated the Metadata field.

Fixes: #794


hazmat345 (approved):
LGTM

Issue #749, opened 2020-11-18
Delete System of Down Garden


TheBurchLog:
When the Garden is shutdown/unresponsive, we should be able to delete the Garden from the System Admin page.

PR #798, opened 2020-12-21
Handle Force Deletes for Child Systems

TheBurchLog:
Beer Garden will now wait on the response from Child Gardens. When we migrate to ASyncio this will improve performance. 

If we capture a router error during the forward of a Forced System Delete, then we will process it locally. That way at a minimum the System Admin can delete the record out of their database. 

fixes: #749

### DIFF
--- a/src/app/beer_garden/router.py
+++ b/src/app/beer_garden/router.py
@@ -271,10 +271,18 @@ def initiate_forward(operation: Operation):
 
     # TODO - Check to ensure garden conn_info is not 'local' before forwarding?
 
-    forward_processor.put(operation)
+    try:
+        response = forward(operation)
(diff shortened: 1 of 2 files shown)


hazmat345 (changes requested):
This logic is getting very circular, I really don't like how we are kicking things back to `execute_local`. I don't think the business logic for this case should live in the same place as the "normal" local executing locally logic.

Also, it seems like we aren't using `forward_processor` anymore... if that's the case then it should be removed completely.

TheBurchLog:
@hazmat345 I went ahead and removed the forward processor. Your additional changes look good.

hazmat345 (approved):
Cool, LGTM 👍🏼

PR #800, opened 2020-12-22
Entry point events

TheBurchLog:
This should fix the plugins from starting before the HTTP Entry point

### DIFF
--- a/src/app/beer_garden/api/http/__init__.py
+++ b/src/app/beer_garden/api/http/__init__.py
@@ -100,7 +100,9 @@ async def startup():
 
     beer_garden.api.http.logger.info("Http entry point is started. Hello!")
 
-    publish(Event(name=Events.ENTRY_STARTED.name, metadata={"entry_point_type": "HTTP"}))
+    publish(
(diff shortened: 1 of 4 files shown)


src/app/beer_garden/api/http/__init__.py
```diff
@@ -124,7 +126,9 @@ async def shutdown():
     # This will almost definitely not be published to the websocket, because it would
     # need to make it up to the main process and back down into this process. We just
     # publish this here in case the main process is looking for it.
-    publish(Event(name=Events.ENTRY_STOPPED.name))
+    publish(
+        Event(name=Events.ENTRY_STOPPED.name, metadata={"entry_point_type": "STOMP"})
```
hazmat345, line 130: This should be HTTP

hazmat345 (changes requested):
Do we not publish an entry stopped for the stomp entry point?

west270 (commented):
LGTM

TheBurchLog:
Just added the stop metadata for STOMP, didn't realize that was thrown in the Stomp Manager.

hazmat345 (approved):
👍🏼

PR #801, opened 2020-12-22
Fixed attribute error

west270:
If garden gets updated and has no configuration would get a attribute error for `event.payload.connection_type.casefold()` not having attribute `casefold()` for object type `None`. Added check to see if object is none


hazmat345 (approved):
👍🏼

PR #807, opened 2020-12-29
Updating system commands with empty list now works correctly

### LINKED ISSUE #806: Empty command list results in no changes when registering
This is kind of odd... if you change a plugin to have no commands that change won't be respected. As in, the old list of commands will still be there.

hazmat345:
Fixes #806, depends on beer-garden/brewtils#277


TheBurchLog (approved):
Works great when I remember to pull in brewtils

PR #808, opened 2020-12-30
Issue/805

### LINKED ISSUE #805: Unassociated Runner not displayed when there are no systems
Need to tweak the display logic

hazmat345:
Fixes #805

### DIFF
--- a/src/ui/src/js/controllers/admin_system.js
+++ b/src/ui/src/js/controllers/admin_system.js
@@ -2,6 +2,7 @@ import _ from 'lodash';
 import readLogs from '../../templates/read_logs.html';
 import adminQueue from '../../templates/admin_queue.html';
 import forceDelete from '../../templates/system_force_delete.html';
+import {responseState} from '../services/utility_service.js';
 
 adminSystemController.$inject = [
   '$scope',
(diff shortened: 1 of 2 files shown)


src/ui/src/js/controllers/admin_system.js
```diff
@@ -272,6 +273,11 @@ export default function adminSystemController(
     $scope.runnerResponse = response;
     $scope.runners = response.data;
 
+    // This is kind of messy, but oh well
+    if (responseState($scope.response) === 'empty') {
+      $scope.response = response;
+    }
+
```
TheBurchLog, line 280: So what is actually going on here? If nothing is returned, then set the `$scope.response` to an empty list?

TheBurchLog (commented):
If my comment is accurate, then I a guess it is approved :man_shrugging:

src/ui/src/js/controllers/admin_system.js
hazmat345: Yeah, I'm not super happy about this, but it's at least an improvement until we get the new system admin page.

So you know the big yellow display that says "Whoops, couldn't find that" when you go to the system admin page when there aren't any systems? That's based off of the return code of `$scope.response`. So even if there are unassociated runners we were still showing that big banner.

This change just hides the banner for that specific case.

The root cause of this is that we're combining two "queries" (systems and runners) into one logical grouping. Our frontend isn't that flexible so I did this 😄

hazmat345:
I'm going to merge this - it's not exactly clean, but it does improve the situation until the admin page redesign is done.

PR #812, opened 2021-01-04
Adding Checkers for Code Complexity and Security to PR Actions

SpecificallyGravity:
### Complexity Checker

This action compares the complexity results of the `base` reference to the `current` reference.  If any alterations/additions exceed the suggested [Cyclomatic Complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity) of 10, an error is thrown.

Cyclomatic Complexity is not an end-all metric, but can be considered a code smell.  If something is flagged, double check the code and consider refactoring.

### Security Checker
This action compares the security results of the `base` reference to the `current` reference.  The action uses a security-flavored linter (`bandit`) to analyze the code.

### DIFF
--- a/.github/workflows/pr-actions.yml
+++ b/.github/workflows/pr-actions.yml
@@ -156,7 +156,6 @@ jobs:
         uses: jitterbit/get-changed-files@v1
 
       - name: Compare complexities
-        working-directory: ./src/app
         run: |
-          wily build
+          wily build ./src/app/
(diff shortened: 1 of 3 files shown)


SpecificallyGravity:
Swapped from `Wily` to `flake8` because it didn't play nicely with the GitHub Checkout action for some reason.

hazmat345 (approved):
Interesting, I've never heard of bandit before. It'll be cool to see if it catches anything. It'll also be nice to see the results of the complexity analysis.

LGTM

Issue #813, opened 2021-01-04
Non-void self-closing tags need to be fixed


hazmat345:
This was originally noticed because child command names weren't visible on the request view page.

The issue was this:

```html
<div>
    <span class="fa fa-chevron-right" style="visibility: hidden" />
    <a class="child-link" ui-sref="base.request({requestId: child.id})">{{child.command}}</a>
</div>
```

I'm pretty sure that span tag was always "wrong," but the browsers handled it like a `</span>` tag immediately followed. But now with HTML5 they just ignore the slash, which results in the link being a part of the span.

I think this is actually the second time this has bitten us - I think the 3-column layout on the system admin page relied on this as well. We should verify that this isn't an issue in all of our partials.

hazmat345:
Haven't seen this come up again, and I don't think combing through all the HTML is worth the effort. Closing this and we'll deal with this if it comes up again.

PR #814, opened 2021-01-04
Self-closing non-void tags are not great in HTML5

hazmat345:
This fixes the most visible part of #813, but we should go through and verify this isn't an issue elsewhere.


hazmat345:
Going to go ahead and merge this.

PR #821, opened 2021-01-07
Issue/819

### LINKED ISSUE #819: Scheduler Not Rendering Trigger Form
Somewhere between the 3.0.0 and 3.0.4 releases a dependency was updated that has caused the scheduled page to no longer work. This can be seen in both our RPM and Docker artifacts. This wasn't noticed at first because no one ever runs `npm install`....

What we know:
The Scheduler renders two forms, Request and Job. When the HTML condition is sastifified for the Job form to be displayed, nothing is returned. We can see that the `$scope.jobModel` is not populated by looking at the preview box on the right. 

I have attempted to reload the Request Form in the Job's place in the HTML and I got the same result. So it appears to not be an issue with how we are formatting the sf-form values. 

Now we have to figure out which package changed to resolve this....

hazmat345:
Should fix #819. Breaks the job creation into separate states.


src/ui/src/js/configs/routes.js
```diff
@@ -62,16 +61,16 @@ export default function routeConfig(
     })
     .state('base.about', {
       url: 'about',
-      templateUrl: basePath + 'about.html',
+      templateUrl: 'about.html',
```
TheBurchLog, line 64: Just to make sure that I am tracking this properly, the removal of `basePath` is possible because of the change in the webpack? As long as this works when we publish the package I am good.

TheBurchLog (approved):
Everything looks good, pretty much a copy/paste into separate pages and adding the controllers for those pages. That was going to be my approach today if you hadn't done it.

This should make it easy for the future when we want to edit the Jobs because we can add a State Param for the original object and map that into the Request and Trigger pages.

src/ui/src/js/configs/routes.js
hazmat345: Yeah, that's correct. I didn't really take the time to understand that part of the webpack config when I created it, so it was kind of screwy and wouldn't work with child directories.

hazmat345:
@TheBurchLog did you pull this down and test it out? I want to make sure it works for everybody, not just me.

Issue #803, opened 2020-12-22
Log level not respected in plugin logging config


hazmat345:
I'm seeing an issue where changing the level of the root logger in the plugin logging config files is not making it into the plugins when they come up.

hazmat345:
I suspect this isn't a real problem. Going to investigate.

hazmat345:
Can't reproduce this, closing.

PR #826, opened 2021-01-08
Adding log message when logging config file is changed

hazmat345:
This doesn't really close out #803, but I'm guessing at this point that's going to be closed as not reproducible. And I think this is useful to have.


TheBurchLog (approved):
:+1:

PR #834, opened 2021-01-14
Issue/833

### LINKED ISSUE #833: Owner Collection
In the last release we added the Owner object as a metadata class to allow for easy cleanups of File Objects that are linked to Jobs and Requests. What we didn't realize is that this migrated all Request objects in the Mongo collection over to the Owner collection. This is not expected functionality.

We need to do the following actions:

- [x] Fix the Mongo objects to be mapped to the correct collections (Requests/Jobs instead of Owner)
- [x] Create migration path for fixing broken database
  - [x] Pre-3.0.2 requests still exist in a request collection. The migration shouldn't break them
  - [x] Post-3.0.2 requests that exist in the owner collection must be moved to the request collection. This must be tested on child requests.
  - [x] Post-3.0.2 jobs that exist in the owner collection must be moved to the owner collection.
  - [x] Files in the file collection must be updated to reflect the new structure

In addition, there's a brewtils component to this as well:
- [x] The structure of the models is changing. If there is functionality that will not be compatible pre-3.0.5 and post-3.0.5 that needs to fail gracefully:
  - [x] Attempting to create a request with a file parameter using pre-3.0.5 brewtils with post-3.0.5 beer-garden needs to either work or fail gracefully.
  - [x] Attempting to create a request with a file parameter using post-3.0.5 brewtils with pre-3.0.5 beer-garden needs to either work or fail gracefully.

SpecificallyGravity:
Hotfix for DB structure issues involving inheritance of a Mongo engine document.
Requires https://github.com/beer-garden/brewtils/pull/284

Fixes #833

### DIFF
--- a/src/app/beer_garden/db/mongo/models.py
+++ b/src/app/beer_garden/db/mongo/models.py
@@ -266,12 +266,7 @@ class Instance(MongoModel, EmbeddedDocument):
             )
 
 
-class Owner(Document):
-    # Stub to use for a reference field.
-    meta = {"allow_inheritance": True}
-
-
-class Request(MongoModel, Owner):
+class Request(MongoModel, Document):
     brewtils_model = brewtils.models.Request
 
     # These fields are duplicated for job types, changes to this field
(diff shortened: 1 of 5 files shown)


hazmat345 (changes requested):
Migration doesn't work for Requests with a parent - the MongoParser doesn't parse it correctly.

hazmat345:
Going to merge this, see my comments on the issue.

Issue #816, opened 2021-01-06
Actually run systemd service as beer-garden user


hazmat345:
~Will probably need an upgrade path for this as well~ - The application logging file is currently owned by root, but we're already wholesale changing ownership of the entire application directory in the after_install script, which fixes any outstanding ownership issues.

Another consideration: this change will make it impossible to run the HTTP entrypoint with a system port (<1024). We could look into doing some trickery like Apache or Nginx to drop permissions, but I don't know if that's worth it.

PR #844, opened 2021-01-25
Service now runs with 0002 umask

hazmat345:
This helps with #816. It makes the beer-garden group (not just user) have write permissions on created files in /opt/beer-garden.


TheBurchLog (approved):
:+1:

PR #846, opened 2021-01-26
Better database migration

hazmat345:
This addresses some serious flaws with the request migrations:

- Now works for child requests
- Doesn't change the updated_at timestamp

There is still work that needs to be done to bring these changes in line with other changes. Specifically, the File model `request` and `job` fields are not updated yet.


TheBurchLog:
The code you changed looks fine, but I am curious why we have all the PyTest errors. I don't see a brewtils requirement on this, so are you getting the same errors in your environment?

PR #853, opened 2021-02-01
Refactoring File Service

SpecificallyGravity:
Refactored some of the file code to increase code reuse.  

There were lots of places that used the `_unroll_object` function with the same set of ignore filters, so a new abstraction was made to simplify those code snippets.

Improvements made to the recursive parameter search to be more pythonic.

### DIFF
--- a/src/app/beer_garden/files.py
+++ b/src/app/beer_garden/files.py
@@ -578,11 +578,9 @@ def handle_event(event):
     """Handle events"""
     if event.garden == config.get("garden.name"):
         if event.name == Events.JOB_CREATED.name:
-            print(event.payload.request_template.parameters)
             for id in _check_file_ids(event.payload.request_template.parameters):
                 set_owner(id, owner_id=event.payload.id, owner_type="JOB")
 
         if event.name == Events.REQUEST_CREATED.name:
-            print(event.payload.parameters)
             for id in _check_file_ids(event.payload.parameters):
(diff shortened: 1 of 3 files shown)


SpecificallyGravity:
Looks like the complexity checker action could be improved; the new recursive search is now lower than the complexity threshold, but `diff` makes it fail because it's different.

src/app/beer_garden/files.py
```diff
@@ -610,9 +578,11 @@ def handle_event(event):
     """Handle events"""
     if event.garden == config.get("garden.name"):
         if event.name == Events.JOB_CREATED.name:
-            set_owner_for_files(
-                event.payload.id, "JOB", event.payload.request_template.parameters
-            )
+            print(event.payload.request_template.parameters)
```
hazmat345, line 581: Extra print

src/app/beer_garden/files.py
```diff
@@ -610,9 +578,11 @@ def handle_event(event):
...
+            for id in _check_file_ids(event.payload.request_template.parameters):
+                set_owner(id, owner_id=event.payload.id, owner_type="JOB")
 
         if event.name == Events.REQUEST_CREATED.name:
-            set_owner_for_files(event.payload.id, "REQUEST", event.payload.parameters)
+            print(event.payload.parameters)
```
hazmat345, line 586: Extra print

hazmat345 (approved):
Except for the extra print calls, LGTM.

PR #856, opened 2021-02-02
Improving the complexity action output 

SpecificallyGravity:
Trying to alter how the complexity checker action works so that only negative changes cause it to fail, not just changes in general.


hazmat345 (approved):
LGTM I guess?

PR #865, opened 2021-02-05
Changing placement of parent request popover

### LINKED ISSUE #862: Request Index page parent tooltip cut off
The tooltip shown when hovering over the up-arrow indicating a child request is shown to the left. This results in it being cut off.

hazmat345:
Fixes #862, realigns popover.


TheBurchLog (approved):
That looks 100% better because we can actually see it.

PR #864, opened 2021-02-05
Changing checkbox captions on request index page

hazmat345:
Fixes #861, changes captions.

PR #866, opened 2021-02-05
Issue/860

### LINKED ISSUE #860: Show hidden checkbox label on commands page
Clicking the label doesn't update the checkbox. You have to click the actual box.

hazmat345:
Fixes #860, corrects label

Merge after #864

### DIFF
--- a/src/ui/src/js/configs/dt_renderer.js
+++ b/src/ui/src/js/configs/dt_renderer.js
@@ -32,36 +32,36 @@ export default function runDTRenderer(DTRendererService) {
       }
 
       if (options && options.hiddenRequestContainer) {
-          let hiddenRequestContainer = $('<span>')
-            .attr('id', 'hiddenRequestContainer')
-            .css('margin-right', '20px')
-            .append(
-              $('<input>')
-                .attr('id', 'hiddenRequestCheck')
-                .attr('type', 'checkbox')
-                .css('margin-top', '-4px')
-                .change(() => { $('.dataTable').dataTable().fnUpdate(); })
-            )
-            .append(
-              $('<label>')
-                .attr('for', 'hiddenRequestCheck')
-                .css('padding-left', '4px')
-                .text('Show Hidden')
-            );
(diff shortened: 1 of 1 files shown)


src/ui/src/js/configs/dt_renderer.js
```diff
@@ -26,42 +26,42 @@ export default function runDTRenderer(DTRendererService) {
             $('<label>')
               .attr('for', 'childCheck')
               .css('padding-left', '4px')
-              .text('Include Children')
+              .text('Show Children')
```
TheBurchLog, line 29: You need to update the attribute here to 'filterHidden' to match what you did on line 61

src/ui/src/js/configs/dt_renderer.js
```diff
@@ -26,42 +26,42 @@ export default function runDTRenderer(DTRendererService) {
...
-                  .attr('for', 'childCheck')
-                  .css('padding-left', '4px')
-                  .text('Show Hidden')
-              )
+          .attr('id', 'hiddenList')
+          .css('margin-right', '20px')
+          .attr('style', 'list-style-type:none;margin-right:20px; display:inline; padding-left:0px;')
+          .append(
+            $('<label>')
+            .attr('for', 'filterHidden')
```
TheBurchLog, line 61: If you are going to change the attribute name, you also need to update request_index.js line 43 because it references this attribute.

hazmat345: This is ... confusing. I really wish there were a better way to do this. I even spent some time trying to rework this but failed :/

The request index page doesn't use this chunk, it uses the chunk above (`if (options && options.hiddenRequestContainer)`). This one is only used on the command index page.

src/ui/src/js/configs/dt_renderer.js
hazmat345: No, this label is `for` the input defined above it, and the id of that input is `childCheck`.

These blocks are all independent. I wish we could consolidate them into one so we wouldn't have to worry about this, but I haven't figure out how to do that.

PR #880, opened 2021-02-16
Issue/837 Adds expiration date to requests model

### LINKED ISSUE #837: Display information related to Request ageoff
We should update the Request View page:
- Right now we don't even display the command type (ACTION vs INFO). That would be a good start.
- It'd also be nice to see the actual time a request will be a candidate for deletion.
- Even better would be some sort of countdown timer. It would need to start off as unobtrusive, and get more visually apparent the closer it got to 0. Once it reaches 0 it should become something that says something to the effect of "this request may already be deleted, refresh this page at your own risk."

west270:
Added `expiration_date` to Request model. Also added ability to update the `expiration_date`, and the pruner to delete the requests if request is expired. There is also a migration script to migrate requests to have `expiration_date` when the Request document does not currently have the field. When updating the `expiration_date`, it will find the parent then update the parent's `expiration_date`.

Dependent on PR beer-garden/brewtils#292

brewtils: request_ttl

Fixes #837


src/app/beer_garden/db/mongo/models.py
```diff
@@ -312,6 +313,7 @@ class Request(MongoModel, Document):
             {"name": "command_index", "fields": ["command"]},
             {"name": "command_type_index", "fields": ["command_type"]},
             {"name": "system_index", "fields": ["system"]},
+            {"name": "expiration_date_index", "fields": ["expiration_date"]},
```
TheBurchLog, line 316: for optimization, you should have an index of both `expiration_date` and `status` and `has_parent` so we don't have to inspect every file while pruning

src/app/beer_garden/api/http/handlers/v1/request.py
```diff
@@ -108,6 +108,9 @@ async def patch(self, request_id):
                         raise ModelValidationError(
                             f"Unsupported status value '{op.value}'"
                         )
+                elif op.path == "/expiration_date":
+                    operation.operation_type = "REQUEST_UPDATE"
```
TheBurchLog, line 112: You either need a second flag for setting the new date or need a different function to route to.

TheBurchLog (commented):
The UI is looking good. When using it, it is not intuitive what the time actually is. I assumed it was going to use to local environment but it seems to be off. I am assuming that it resolves down to the actual date-time integer so we should be good?

src/app/beer_garden/requests.py
```diff
@@ -677,6 +678,36 @@ def start_request(request_id: str = None, request: Request = None) -> Request:
...
+        expiration_date = new_values.pop("expiration_date", None)
+        parent = find_parent(request)
+        parent.expiration_date = expiration_date
+        parent = db.update(parent)
+        if parent.id == request_id:
+            request = parent
+    for key in new_values.keys():
+        setattr(request, key, new_values[key])
+    if new_values:
+        request = db.update(request)
```
hazmat345, line 700: This seems pretty dangerous. Do we have a use-case in mind driving this change?

Historically we've only allowed updates to requests in certain specific cases, but this seems to expand that. The mongo model validation should prevent some things (changing the output after the request completes, for example). But this still seems dangerous to me - is it now possible to change the ID of a particular request?

src/app/beer_garden/requests.py
```diff
@@ -677,6 +678,36 @@ def start_request(request_id: str = None, request: Request = None) -> Request:
...
+    return request
+
+
+def find_parent(request):
```
hazmat345, line 704: This should be named something else. Returning the request passed should not be the behavior of a function called "find_parent."

hazmat345 (changes requested):
We don't need both of these in the logs, let's pick one:
```
2021-03-16 17:23:20,710 - beer_garden.db.mongo.util - WARNING - Found missing Request expiration date index, about to build it. This could take a while :)
2021-03-16 17:23:20,711 - beer_garden.db.mongo.util - WARNING - Found missing Request indexes, about to build them. This could take a while :)
```

The UI needs some changes:
- Can we make the change button look less buttony?
- The display shown when the button is clicked needs to be s popover or something that doesn't reflow the page.
- There needs to be another way to communicate that the request might have been deleted besides a big banner. Maybe turning the expiration date text red or something? And then the title of the "change expiration" popover could be "Request Expired"?
- This one is non-negotiable: When I tried to update the expiration date for a request that had been pruned nothing happened. There needs to be some sort of indication that the request is actually gone and that the expiration update did NOT work because the request was already gone.

src/app/beer_garden/db/mongo/pruner.py
```diff
@@ -116,6 +93,6 @@ def determine_tasks(**kwargs) -> Tuple[List[dict], int]:
 
         # Look at the various TTLs to determine how often to run
         real_ttls = [x for x in kwargs.values() if x > 0]
-        run_every = min(real_ttls) / 2 if real_ttls else None
+        run_every = min(real_ttls) / 2 if real_ttls else 15
```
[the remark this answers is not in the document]
west270: I can add a config option for it but the problem I have is when `run_every` is not in config or is set to -1 then the pruner would never run even though expiration dates can still be scheduled from config or the ui. I think we would still need a hardcoded value so the pruner can prune items with an expiration date.

src/app/beer_garden/requests.py
west270: I added a buffer function for updating `expiration_date` so the function can not be called by the rest api. I initially had it doing just `expiration_date` but changed the function able to work for any attribute to resolve TheBurchLog https://github.com/beer-garden/beer-garden/pull/880#discussion_r583767205 above.

PR #884, opened 2021-02-19
Issue/879

### LINKED ISSUE #878: Parent event publisher EasyClient is wrong
The EasyClient used to publish events to a parent is only given a host, port, and ssl_enabled.

### LINKED ISSUE #879: Parent HTTP configuration is wrong
The configuration section for specifying how to talk to a parent is missing some fields (like, ca_verify and client_cert) and has other fields that are not useful (ca_path, client_cert_verify, private and public key).

I suspect this relates to #878.

hazmat345:
Fixes #878, fixes #879

This corrects the HTTP parent spec and uses it when creating the EasyClient.

### DIFF
--- a/src/app/beer_garden/app.py
+++ b/src/app/beer_garden/app.py
@@ -30,7 +30,6 @@ import beer_garden.queue.api as queue
 import beer_garden.router
 from beer_garden.events.handlers import garden_callbacks
 from beer_garden.events.parent_procesors import HttpParentUpdater
-from beer_garden.scheduler import MixedScheduler
 from beer_garden.events.processors import (
     FanoutProcessor,
     QueueListener,
@@ -40,6 +39,7 @@ from beer_garden.log import load_plugin_log_config
 from beer_garden.metrics import PrometheusServer
 from beer_garden.monitor import MonitorFile
 from beer_garden.plugin import StatusMonitor
+from beer_garden.scheduler import MixedScheduler
 
 
 class Application(StoppableThread):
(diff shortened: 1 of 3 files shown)


TheBurchLog (commented):
There was not a brewtils branch noted in the PR, so I am assuming you are using develop. In develop, there is no migration path for the existing variables.

hazmat345:
I've reworked the spec to be explicit instead of copying from brewtils. This means that the `bg` prefixes are no longer there. I've also removed the reorganization where everything is pushed into a "connection" dictionary. I can not get the config migration to work, which suggests we have a bug. I'm not interested in going down that particular rabbit hole right now, so I backed that out.

This does remove some config options like `client_cert_verify`. However, the things that are removed don't make sense for the parent section. And they weren't being used, anyway.

TheBurchLog:
@hazmat345 
For every other SSL configuration the information is nested. If we want to be consistent, then we should keep that going in the Parent configuration. I know that it is easier to flatten it to make it match with brewtils, but maybe we need to update how Brewtils handles it instead? Make Brewtils nested to match what Beer-Garden does? Also, I don't see the migration scripts ever working for this because they Brewtils and Beer-Garden have different structures in their configs. Brewtils wouldn't know about `parent.http` .

hazmat345:
Ok, I'll back out the change where I flattened things.

Changing the brewtils spec would be a big lift. I'm not opposed but it'd have to be on a major release for sure.

hazmat345:
Ok, going with this nested ssl version for conformity. Merging.

PR #885, opened 2021-02-20
Adding % to garden url regex so encoded unicode will match

### LINKED ISSUE #875: Garden name with unicode characters
This doesn't seem to work.

hazmat345:
Fixes #875, add support for unicode garden names

@TheBurchLog you may want to do the same thing for the Users api?


TheBurchLog (approved):
yeah, I will look at this for the Users API as well.

Issue #897, opened 2021-02-26
Child garden connection parameters are ignored


hazmat345:
There are several parameters that can be specified when configuring a child garden but that aren't actually used when making the HTTP post. This is ... not good.

hazmat345:
This is "fixed" but the fix completely breaks things if there's no client cert provided.

PR #899, opened 2021-02-26
Use SSL configuration provided by UI Garden Admin Form

TheBurchLog:
Now we will actually use the SSL properties provided through the UI when connecting to Children

fixes: #897

### DIFF
--- a/src/app/beer_garden/router.py
+++ b/src/app/beer_garden/router.py
@@ -617,16 +617,15 @@ def _forward_http(operation: Operation, target_garden: Garden):
         conn_info.get("url_prefix", "/"),
     )
 
-    response = None
-
     try:
+        response = None
+
         if conn_info.get("ssl"):
-            http_config = config.get("entry.http")
             response = requests.post(
                 endpoint,
                 data=SchemaParser.serialize_operation(operation),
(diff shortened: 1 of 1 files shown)


src/app/beer_garden/router.py
```diff
@@ -617,16 +617,15 @@ def _forward_http(operation: Operation, target_garden: Garden):
...
+        response = None
+
         if conn_info.get("ssl"):
-            http_config = config.get("entry.http")
             response = requests.post(
                 endpoint,
                 data=SchemaParser.serialize_operation(operation),
-                cert=http_config.ssl.ca_cert,
-                verify=http_config.ssl.ca_path,
+                cert=conn_info.ca_cert,
```
hazmat345, line 627: This needs to be client cert

src/app/beer_garden/router.py
```diff
@@ -617,16 +617,15 @@ def _forward_http(operation: Operation, target_garden: Garden):
...
+                verify=conn_info.ca_path if conn_info.ca_verify else None,
```
hazmat345, line 628: This needs to be ca_cert

hazmat345 (changes requested):

Issue #888, opened 2021-02-23
HTTP API Client use Operation serialize_kwargs


TheBurchLog:
If we pass forward the `serialize_kwargs` within the `operation.kwargs`, then we should utilize that field for serializing the models

PR #900, opened 2021-03-01
Updated Client function to find serialization malforms

TheBurchLog:
Updated the Client function for the HTTP API. Now it expected a single argument for input and if the serialization is passed through the Operation Kwargs it is utilized, but first it throws an something into the Logger for us to go back and address that malform request.

fixes: #888

### DIFF
--- a/src/app/beer_garden/api/http/client.py
+++ b/src/app/beer_garden/api/http/client.py
@@ -14,8 +14,8 @@ logger = logging.getLogger(__name__)
 
 
 class SerializeHelper(object):
-    async def __call__(self, operation: Operation, serialize_kwargs=None, **kwargs):
-        result = beer_garden.router.route(operation, **kwargs)
+    async def __call__(self, operation: Operation, serialize_kwargs=None):
+        result = beer_garden.router.route(operation)
 
         # Await any coroutines
         if isawaitable(result):
--- a/src/app/beer_garden/api/http/client.py
+++ b/src/app/beer_garden/api/http/client.py
@@ -1,18 +1,21 @@
 # -*- coding: utf-8 -*-
 import json
+import logging
 from inspect import isawaitable
 
 import six
-from brewtils.models import BaseModel
+from brewtils.models import BaseModel, Operation
 from brewtils.schema_parser import SchemaParser
 
 import beer_garden.api
 import beer_garden.router
 
+logger = logging.getLogger(__name__)
+
 
 class SerializeHelper(object):
-    async def __call__(self, *args, serialize_kwargs=None, **kwargs):
-        result = beer_garden.router.route(*args, **kwargs)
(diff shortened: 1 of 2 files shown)


src/app/beer_garden/api/http/handlers/v1/system.py
```diff
@@ -312,7 +312,8 @@ async def get(self):
                     "exclude_fields": exclude_fields,
                     "dereference_nested": dereference_nested,
                 },
-            )
+            ),
+            serialize_kwargs=serialize_kwargs,
```
hazmat345, line 316: If you're adding this here, shouldn't you remove the `serialize_kwargs` from the operation kwargs?

src/app/beer_garden/api/http/client.py
```diff
@@ -1,23 +1,38 @@
...
+        # This is a helper function to find bad formatted Operations/Clients
+        if (
+            serialize_kwargs is None
+            and operation.kwargs is not None
+            and "serialize_kwargs" in operation.kwargs
+        ):
+            serialize_kwargs = operation.kwargs["serialize_kwargs"]
+            logger.info(
+                f"Serialized Kwargs was pass through Operation instead of Client "
+                f"function for {operation}, this was migrated but should be fixed"
```
hazmat345, line 33: I don't think this is a good idea. You're adding a non-trivial check to everything coming through the REST API at runtime to catch a programming error. If we want to *make sure* that this never happens then the right place to do that is in our tests.

I'm not convinced that's necessary, though. Unless we start seeing this happen all over the place it's probably not worth the effort. But I definitely don't think it's worth the cost of this check.

src/app/beer_garden/api/http/client.py
```diff
@@ -1,23 +1,38 @@
...
 import beer_garden.router
 
+logger = logging.getLogger(__name__)
+
 
 class SerializeHelper(object):
-    async def __call__(self, *args, serialize_kwargs=None, **kwargs):
-        result = beer_garden.router.route(*args, **kwargs)
+    async def __call__(self, operation: Operation, serialize_kwargs=None, **kwargs):
+        result = beer_garden.router.route(operation, **kwargs)
```
hazmat345, line 18: This is a good idea... we should always be invoking this with an operation.

hazmat345 (changes requested):
See comments

src/app/beer_garden/api/http/client.py
TheBurchLog: I agree, so I will remove it because we fixed it in the API

src/app/beer_garden/api/http/handlers/v1/system.py
TheBurchLog: You are right, that is redudent

hazmat345:
Rebased and consolidated this branch, added an additional commit. We should either merge this or close.

Issue #906, opened 2021-03-04
Remove Local Garden Option


TheBurchLog:
The UI should not have the option to define connect type as Local. If you set this, then it is impossible to change it back or delete it in the UI.

PR #909, opened 2021-03-04
Remove Local Garden Connection Type

TheBurchLog:
Removed the drop down for Local Connection Type on the Garden Edit page. This will prevent Admin's from accidentally bricking their Garden editor.

Fixes: #906


hazmat345 (approved):
👍🏼

PR #910, opened 2021-03-04
Fixed Bad Config Preventing Startup

### LINKED ISSUE #907: Bad Garden Config stops startup
If a Garden Connection Info is not populated for STOMP, but that is the set Connect Type, it will stop the Startup and not allow it to proceed. Need to better handle failed STOMP startups.

west270:
Fixed error that cause Beer Garden not to start when garden is configured to use stomp but is the field in config are empty. Also added logger warning message for connection failures to include a list of gardens affected so if the problem is with the connection config you can find which gardens to check in the logs.

Fixes #907

### DIFF
--- a/src/app/beer_garden/api/stomp/server.py
+++ b/src/app/beer_garden/api/stomp/server.py
@@ -156,8 +156,8 @@ class Connection:
                             logger.info("Stomp successfully " + connected_message)
 
                     except Exception as e:
-                        logger.warning(str(e))
                         logger.warning(
+                            f"Error connecting: {type(e).__name__}. "
                             f"Affected gardens are {[garden.get('name') for garden in gardens]}"
                         )
                         logger.warning(
--- a/src/app/beer_garden/api/stomp/server.py
+++ b/src/app/beer_garden/api/stomp/server.py
@@ -107,7 +107,7 @@ class Connection:
         username=None,
         password=None,
     ):
-        self.host_and_port = host_and_ports
+        self.host_and_ports = host_and_ports
         self.username = username
         self.password = password
         self.subscribe_destination = subscribe_destination
@@ -129,8 +129,8 @@ class Connection:
     def connect(self, connected_message=None, gardens=None):
         wait_time = 0.1
         if (
-            self.host_and_port[0][0]
-            and self.host_and_port[0][1]
+            self.host_and_ports[0][0]
+            and self.host_and_ports[0][1]
             and self.subscribe_destination
         ):
             while not self.conn.is_connected() and self.bg_active:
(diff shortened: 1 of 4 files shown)


TheBurchLog (commented):
The fact that it attempts to connect with missing configurations is bad. We need logic added in the UI that when the Connection Type of STOMP is selected they are required to populate the fields.

src/app/beer_garden/api/stomp/server.py
```diff
@@ -107,6 +107,7 @@ def __init__(
         username=None,
         password=None,
     ):
+        self.host_and_port = host_and_ports
```
hazmat345, line 110: Shouldn't these be named the same?

src/app/beer_garden/api/stomp/server.py
```diff
@@ -125,34 +126,44 @@ def __init__(
...
-                        destination=self.subscribe_destination,
-                        id=self.username,
-                        ack="auto",
-                        headers={
-                            "subscription-type": "MULTICAST",
-                            "durable-subscription-name": self.subscribe_destination,
-                        },
+        if (
+            self.host_and_port[0][0]
+            and self.host_and_port[0][1]
```
hazmat345, line 133: Can't `host_and_port[0]` be `None`? If that happens this is going to blow up.

hazmat345 (changes requested):

src/app/beer_garden/api/stomp/server.py
west270: @hazmat345 `host_and_port` defaults to `None` if nothing is passed in, but a value is always passed in. If the value of `host` and `port` are `None` then value passed in on init is ```host_and_port = [(None, None)]```, so `host_and_port` should not be `None`.

src/app/beer_garden/api/stomp/server.py
```diff
@@ -128,42 +128,43 @@ def __init__(
...
+                                    "durable-subscription-name": self.subscribe_destination,
+                                },
+                            )
+                        if connected_message is not None and self.conn.is_connected():
+                            logger.info("Stomp successfully " + connected_message)
+
+                    except Exception as e:
+                        logger.warning(str(e))
+                        logger.warning(
+                            f"Affected gardens are {[garden.get('name') for garden in gardens]}"
```
hazmat345, line 161: If you separate these into 2 calls there could be any number of lines in between them in the actual log file. It's safer to combine them into one call:

```python
logger.warning("Error connecting: {e}. Affected gardens are {[garden.get('name') for garden in gardens]}")
```

src/app/beer_garden/api/stomp/server.py
```diff
@@ -128,42 +128,43 @@ def __init__(
...
+                        )
+                        logger.warning(
```
hazmat345, line 163: This should be debug level.

hazmat345 (changes requested):

hazmat345 (approved):
LGTM, 👍🏼

PR #917, opened 2021-03-10
Stomp Reconnect Loop Break for Garden Change

### LINKED ISSUE #914: STOMP Reconnect Loop Break for Change
While a Garden Connection is in the back off loop for STOMP, it does not receive an interrupt when the Connection Type changes to HTTP.  So we need to add some logic to that loop to ensure that it is still a valid connection.

west270:
Moved stomp reconnect loop up to manager so stomp entry point can still process events.

Dependent on PR #910 

Fixes #914

### DIFF
--- a/src/app/beer_garden/api/stomp/server.py
+++ b/src/app/beer_garden/api/stomp/server.py
@@ -1,6 +1,5 @@
 import stomp
 import logging
-import time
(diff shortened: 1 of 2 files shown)


hazmat345 (approved):
So this looks good from what I can tell. I'm still not sure how to test this, but I'm good with merging it.

PR #920, opened 2021-03-15
Forward Event as Operation for HTTP

TheBurchLog:
To match the STOMP interface, HTTP events will be forwarded as Operations.

Depends on: https://github.com/beer-garden/brewtils/pull/301


src/app/beer_garden/events/parent_procesors.py
```diff
@@ -45,7 +45,10 @@ def process(self, event: Event):
 
         if event.name not in self._black_list:
             try:
-                self._ez_client.publish_event(event)
+                operation = Operation(
+                    operation_type="PUBLISH_EVENT", model=event, model_type="Event"
+                )
+                self._ez_client.post_forward(operation)
```
hazmat345, line 51: This should be `_ez_client.forward` now.

hazmat345 (changes requested):

PR #921, opened 2021-03-15
Adding back integration testing

TheBurchLog:
This is the test bed for integration testing.

So, this ended up being a lot of tests. So here is what is going on.

Every test spins up the full Beer Garden environment (Beer Garden, MongoDB, and RabbitMQ). We also build Beer-Garden Unstable for this test to make sure we are testing the latest version of the code.

# Tests
Local Plugins:
- Test the example plugins to ensure they turn on and can be interacted with.

Remote Plugins:
- With every combo of Brewtils/Python we support, connect a remote plugin and task it

Garden to Garden
- Connect a Child Garden
- Configure Child Connection
- Verify the Child Systems are on the Parent
- Task Child through Parent
- Task Child through Child
- Verify both requests are on Parent

Overall this takes around 20 minutes to run the full suite because we have to create the docker images.


.github/workflows/integration-actions.yml
```diff
@@ -0,0 +1,248 @@
...
+          path: ./tmp
+
+      - name: Move Test Plugins
+        run: |
+          cp -r ./tmp/echo ./docker/docker-compose/data/localplugins
+          cp -r ./tmp/complex ./docker/docker-compose/data/localplugins
+          cp -r ./tmp/dynamic ./docker/docker-compose/data/localplugins
+          cp -r ./tmp/sleeper ./docker/docker-compose/data/localplugins
+          cp -r ./tmp/echo-sleeper ./docker/docker-compose/data/localplugins
+          cp -r ./tmp/error ./docker/docker-compose/data/localplugins
```
hazmat345, line 126: Can we just check these out directly to the tmp directory? With this if we add a new plugin we'll have to update this list too.

docker/docker-compose/docker-compose.yml
```diff
@@ -32,6 +33,10 @@ services:
         environment:
             <<: *env
 
+            # If you are running in Ubuntu, then you need to set the Timezone
+            # We are actively investigating why this is
```
hazmat345, line 37: Well, we have a ticket 😄

test/conf/docker-compose.yml
```diff
@@ -0,0 +1,212 @@
...
+        image: grafana/grafana:5.2.3
+        ports:
+            - "3000:3000"
+        volumes:
+            - ./data/grafana/provisioning:/etc/grafana/provisioning
+            - ./data/grafana/dashboards:/var/lib/grafana/dashboards
+        networks:
+            - bg-network
+        depends_on:
+            - prometheus
```
hazmat345, line 91: Do we need prometheus and grafana? If not can we remove them? Less moving parts in the test is always good.

test/integration/local_plugins/conftest.py
```diff
@@ -1,6 +1,9 @@
 import pytest
 
-from helper import setup_easy_client, RequestGenerator
+try:
+    from helper import setup_easy_client, RequestGenerator
+except:
+    from ..helper import setup_easy_client, RequestGenerator
```
hazmat345, line 6: I saw this in another place as well... why is this necessary? It it because the parent and child gardens are in different "levels?" could we move the parent down a level so this wouldn't be needed?

test/integration/remote_plugins/plugins/system_registration_test.py
```diff
@@ -76,14 +77,14 @@ def test_system_register_different_versions(self):
...
-        with pytest.raises(ConflictError):
-            self.easy_client.create_system(new_plugin.system)
+    # def test_system_register_same_display_name(self):
+    #     plugin = create_plugin("test", "1.0.0", TestPluginV1, display_name="TEST")
+    #     start_plugin(plugin, self.easy_client)
+    #     assert_system_running(self.easy_client, "test", "1.0.0")
+    #
+    #     new_plugin = create_plugin("new_test", "1.0.0", TestPluginV1, display_name="TEST")
+    #     with pytest.raises(ConflictError):
+    #         self.easy_client.create_system(new_plugin.system)
```
hazmat345, line 87: What's up with this? Should we just delete it?

hazmat345 (approved):
This is super cool. A couple of questions / comments for my own curiosity, but LGTM.

.github/workflows/integration-actions.yml
TheBurchLog: I first started off with checking out the plugins directly. It technically works, but I didn't like how Child plugin was throwing errors when the downstream Beer Garden connected. I guess that is a fine issue to to have. Also, for the example plugins it works best if we have integration tests for them. So originally I only added those.

test/conf/docker-compose.yml
TheBurchLog: Yeah, I can remove them.

test/integration/local_plugins/conftest.py
TheBurchLog: Honestly, I didn't try having the `..helper` in the tests in github. I just moved the folder and hoped for the best. I can try not moving the folder and seeing what happens.

test/integration/remote_plugins/plugins/system_registration_test.py
TheBurchLog: I actually meant to circle back with you on this one and forgot. Is this a valid test still? I didn't see anything in our code preventing two Systems to have the same display name.

docker/docker-compose/docker-compose.yml
TheBurchLog: A ticket is better than nothing :)

test/integration/local_plugins/conftest.py
TheBurchLog: So, I went back and tested this out in a couple different combos. If I want to be able to run the tests in PyCharm I need `..helper`. If I run them through the Make file, then `helper` is all that is needed. So this is more a quality of life thing for people using an IDE.

PR #930, opened 2021-03-22
Fixing HTTP Linting Error

TheBurchLog:
This is just a test for the API code that was failing the lint checker. It was failing because the values were not declared.... So I declared them as None and it works. Not sure why it doesn't also hit `server`, `tornado_app`, or any of the other ones.


hazmat345 (approved):
LGTM!

PR #933, opened 2021-03-25
Using EasyClient for HTTP forwarding

hazmat345:
Uses EasyClient for HTTP forwarding. Depends on beer-garden/brewtils#301.


TheBurchLog (approved):
Again, how can I critique  such beautifully written code that I wrote :)

PR #935, opened 2021-03-27
Stomp updates

### LINKED ISSUE #922: Full CPU utilization
I'm seeing an issue where one of the processes is resulting in ~100% CPU utilization. This install is an rpm running 3.1.2rc3 (not sure exactly what commit that is).

The process is

> /opt/beer-garden/bin/python3.7 -c from multiprocessing.spawn import spawn_main; spawn_main(tracker_fd=11, pipe_handle=25) --multiprocessing-fork

I believe that means it's one of the entry point processes. When I accidentally killed the *other* process the REST api stopped working and the problem persisted. When I killed the correct one the utilization went down, but it still takes a strangely long time to log in or switch users.

So I'm guessing this is something with the stomp entry point?

hazmat345:
This fixes #922.

Depends on #917

Stomp updates. Opening PR so I have a place to jot down notes.

- Line breaks
- Type hinting and docstrings
- Using 'key' and 'value' as iterator targets for non-trivial loops
- Error checking strategies
  - Failing as early as possible
  - Runtime problems vs programmer errors
  - Failing in the best way
- Passing arguments just for error logging
- Consistent log levels
- **eval / exec**


src/app/beer_garden/api/stomp/manager.py
```diff
@@ -1,68 +1,98 @@
...
-                    "conn": self.connect(
-                        stomp_config, [{"name": name, "main": is_main}]
-                    ),
-                    "gardens": [{"name": name, "main": is_main}],
-                    "headers_list": headers,
-                }
+        self.ep_conn = ep_conn
+
+    def add_connection(self, stomp_config=None, name=None, is_main=False):
+        host_and_ports = [(stomp_config.get("host"), stomp_config.get("port"))]
```
west270, line 55: Need to add some logic to check if `stomp_config` has `subscribe_destination` and is not `None`. Since the child can be configured in the UI to set `subscribe_destination` to `None`, which means the child needs to send to a topic the entry point is already listening to.

src/app/beer_garden/api/stomp/transport.py
```diff
@@ -0,0 +1,212 @@
...
+def send(
+    body: Any,
+    garden_headers: dict = None,
+    conn: stomp.Connection = None,
+    send_destination: str = None,
+    request_headers: dict = None,
+):
+    message, model_headers = process(body)
+
+    headers = consolidate_headers(model_headers, request_headers, garden_headers)
```
west270, line 66: Should change order so the result response sends the correct `model_class` in the header, since a child garden will send a request object as a response from receiving a request create operation.

src/app/beer_garden/api/stomp/transport.py
```diff
@@ -0,0 +1,212 @@
...
+        Returns:
+            None
+        """
+        try:
+            operation = SchemaParser.parse_operation(message, from_string=True)
+
+            if hasattr(operation, "kwargs"):
+                operation.kwargs.pop("wait_timeout", None)
+
+            result = beer_garden.router.route(operation)
```
west270, line 105: Should either check if `model_class` from headers is an operation or that `operation` is not none, since can receive request objects from child gardens.

west270 (changes requested):

src/app/beer_garden/api/stomp/transport.py
```diff
@@ -0,0 +1,212 @@
...
+        username: str = None,
+        password: str = None,
+    ):
+        self.host = host
+        self.port = port
+        self.username = username
+        self.password = password
+        self.subscribe_destination = subscribe_destination
+        self.send_destination = send_destination
+        self.bg_active = True
```
west270, line 154: `bg_active` can be removed it was used by a while loop I removed from the code.

TheBurchLog: `bg_active` as added back in to be used on line 199 in `transport.py` to set to `False`, then the manager uses it to determine if a re-connection should occur.

src/app/beer_garden/api/stomp/transport.py
```diff
@@ -0,0 +1,212 @@
...
+    message, model_headers = process(body)
+
+    headers = consolidate_headers(model_headers, request_headers, garden_headers)
+
+    if conn.is_connected() and send_destination:
+        destination = send_destination
+        if request_headers and "reply-to" in request_headers:
+            destination = request_headers["reply-to"]
+
+        conn.send(body=message, headers=headers, destination=destination)
```
TheBurchLog, line 73: The IF statement should also take into account of the `request_headers` has `reply-to` set. Then we will publish to that topic regardless.

If only send_destination is set, then only forward `message` objects of type `Operation`. Because that is the only communication supported between Beer Gardens.

src/app/beer_garden/api/stomp/transport.py
```diff
@@ -0,0 +1,212 @@
...
+
+        Args:
+            headers: Message header dict
+            message: The message body
+
+        Returns:
+            None
+        """
+        try:
+            operation = SchemaParser.parse_operation(message, from_string=True)
```
TheBurchLog, line 100: After walking through this with west270, this made me question have the SchemaParser works. When a child responses to the Create Request Operation with a Request object in the message, the `parse_operation` function returns back `None`. Why does that not raise an error instead? The example we walked through didn't throw an error until line 105, when the error really occurred on line 100.

src/app/beer_garden/api/stomp/transport.py
```diff
@@ -0,0 +1,212 @@
...
+        )
+
+        if ssl and ssl.get("use_ssl"):
+            self.conn.set_ssl(
+                for_hosts=[(self.host, self.port)],
+                key_file=ssl.get("private_key"),
+                cert_file=ssl.get("cert_file"),
+            )
+
+    def connect(self) -> True:
```
TheBurchLog, line 166: Kinda bold to type hint that it will always be `True`, instead of boolean

src/app/beer_garden/api/stomp/transport.py
hazmat345: Yes, we discussed that as well. I don't really have an answer, the attempt to construct the Operation object should have failed when we tried to pass arguments it couldn't handle (parameters, for example).

Opened issue #942 for this.

src/app/beer_garden/api/stomp/transport.py
hazmat345: Gotta be confident!

src/app/beer_garden/api/stomp/manager.py
hazmat345: I'm still not 100% sure why this is the case. Can you add a commit with your preferred way of fixing this?

src/app/beer_garden/api/stomp/transport.py
west270: I agree that it can be removed since @hazmat345 changed how STOMP attempts to reconnect. Prior to the change `bg_active` was used to breakout of a reconnecting loop when system is shutdown.

src/app/beer_garden/api/stomp/manager.py
hazmat345: Sweet, thank you.

PR #936, opened 2021-03-29
Adding Stomp Integration Testing

TheBurchLog:
Adding some testing to verify STOMP is in a functional state. This supports both Garden to Garden over STOMP and direct Garden communication over STOMP.

Also, nested the jobs to inherit from a master list of versions. That way it is easier to support future releases.

On 4/8 the Stomp test is failing due to known issues and the remote plugins using Brewtils 3.2.0 is failing.


TheBurchLog:
We need to decide if these failing tests are acceptable. I think that the STOMP test is acceptable because that should be fixed in the future. The Remote Plugin one is where I have questions.

PR #947, opened 2021-04-04
Issue/946

### LINKED ISSUE #946: Stomp connections
Still having issues making stomp connections.

hazmat345:
This fixes #946 stomp TLS connections.

Here's how to prove that this works. Use the activemq services from our docker-compose files and a locally installed beer-garden.

### Non-TLS
- Start the non-tls activemq container
- Set entry.stomp.enabled: true in the beer-garden config
- Start beer-garden. Stomp connection should connect.
- Stop beer-garden. Set entry.stomp.ssl.use_ssl: true. Start beer-garden. Stomp connection should NOT connect. It will probably hang for a while.
- Stop both beer-garden and the activemq container

### TLS
- Start the TLS activemq container
- Set entry.stomp.ssl.use_ssl: false in the config. Start beer-garden. Stomp connection should *NOT* connect. There should be a connection error logged.
- Change entry.stomp.ssl.use_ssl: true in the config. Start beer-garden. Stomp connection should *NOT* connect. There should be a connection error logged.
- Set the following config items, changing the paths to match your paths. Start beer-garden. Stomp connection should connect.

```yaml
stomp:
  ssl:
    ca_cert: /home/matt/Projects/beergarden/beer-garden/docker/docker-compose/data/certs/ca_certificate.pem
    client_cert: /home/matt/Projects/beergarden/beer-garden/docker/docker-compose/data/certs/client_certificate.pem
    client_key: /home/matt/Projects/beergarden/beer-garden/docker/docker-compose/data/certs/client_key.pem
```


src/app/beer_garden/router.py
```diff
@@ -334,23 +337,20 @@ def forward(operation: Operation):
...
 
-    return conn
+    Args:
+        garden: The garden specifying
+
+    Returns:
+        The created connection wrapper
+
+    """
+    return Connection(**format_connection_params(garden.connection_params))
```
west270, line 353: The input for `Connection` should have `subscription_destination` set `None`, since the entry_point should handle the listening. Having `subscription_destination` set `None` will setup the connection to only be able to send

west270 (changes requested):
https://github.com/beer-garden/beer-garden/blob/58179da69498ddbb889fc5b52118b863a95ba590/src/app/beer_garden/api/stomp/manager.py#L58-L60
When configuring a child garden `ssl` is being returned as `None` when not set in the UI and is causing an attribute error.

src/app/beer_garden/router.py
```diff
@@ -574,22 +576,18 @@ def _determine_target_garden(operation: Operation) -> str:
...
         if not stomp_garden_connections[target_garden.name].is_connected():
-            stomp_garden_connections[target_garden.name].connect(
-                username=target_garden.connection_params["stomp_username"],
-                passcode=target_garden.connection_params["stomp_password"],
-                wait=True,
-                headers={
-                    "client-id": target_garden.connection_params["stomp_username"]
-                },
-            )
+            if not stomp_garden_connections[target_garden.name].connect():
```
TheBurchLog, line 579: Assuming that this returns back a True/False, this looks good. If it returns anything else then we should look more into it.

TheBurchLog (commented):
I didn't scrub through the garden_service.js because it all looked like formatting stuff. Overall, it looks good. I did not run this. My environment was not in a good status for testing it.

src/app/beer_garden/router.py
hazmat345: Yes, part of this PR is that the router (specifically `stomp_garden_connections`) uses our Connection class instead of the stomp.py one. That method returns a boolean.

hazmat345: The UI stuff is not purely formatting. The `ssl` section has been changed from:

* use_ssl
* cert_file
* private_key
* verify_host
* verify_hostname

To:

* use_ssl
* ca_cert
* verify_host
* verify_hostname
* client_cert
* client_key

src/app/beer_garden/api/stomp/transport.py
```diff
@@ -207,7 +209,7 @@ def connect(self) -> bool:
 
                 self.conn.subscribe(
                     destination=self.subscribe_destination,
-                    id=self.username,
+                    id="".join([choice(ascii_letters) for _ in range(10)]),
```
TheBurchLog, line 212: These changes are making the assumption that the topic being connected to is not durable. If the topic is durable, every connection will appear as a new user. So any messages queued up for the previous ID will just sit there. Are we making that decision that this should be our standard operating procedure?

src/app/beer_garden/config.py
```diff
@@ -845,30 +845,27 @@ def _parse_args(args: Sequence[str]) -> Tuple[YapconfSpec, dict]:
             "items": {
                 "use_ssl": {
                     "type": "bool",
-                    "description": "Use SSL for connection",
+                    "description": "Use SSL when connecting to the message broker",
                     "default": False,
                 },
-                "private_key": {
+                "client_key": {
```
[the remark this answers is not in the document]
hazmat345: 👍🏼

PR #951, opened 2021-04-06
Issue/950

### LINKED ISSUE #950: UI Image remove ipv6 default
The UI image default server block currently listens on both ipv4 and ipv6. This breaks in environments that don't support ipv6.

hazmat345:
This fixes #950, stops using ipv6 in the UI image.


src/ui/default.conf.template
```diff
@@ -1,6 +1,5 @@
 server {
     listen       80;
-    listen  [::]:80;
```
TheBurchLog, line 3: Is that what this line did, was IPv6? If so, approved

TheBurchLog (approved):

PR #954, opened 2021-04-06
Namespace not required for Jobs

TheBurchLog:
When creating a Job, the namespace field is not required to be populated.

### DIFF
--- a/src/app/beer_garden/db/mongo/models.py
+++ b/src/app/beer_garden/db/mongo/models.py
@@ -275,7 +275,6 @@ class Request(MongoModel, Document):
         "system": {"field": StringField, "kwargs": {"required": True}},
         "system_version": {"field": StringField, "kwargs": {"required": True}},
         "instance_name": {"field": StringField, "kwargs": {"required": True}},
-        "namespace": {"field": StringField, "kwargs": {"required": True}},
         "command": {"field": StringField, "kwargs": {"required": True}},
         "command_type": {"field": StringField, "kwargs": {}},
         "parameters": {"field": DictField, "kwargs": {}},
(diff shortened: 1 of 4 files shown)


TheBurchLog:
To test this manually, I used this code:
```python
server = EasyClient(bg_host="localhost", bg_port=2337, ssl_enabled=False)
job_name = "no-namespace-test"
template = RequestTemplate(
    system='echo',
    system_version='3.0.0.dev0',
    instance_name='default',
    command='say',
    parameters={
        'message': "hello",
        'loud': False
    },
    comment=job_name + ' Job'
)

trigger = IntervalTrigger(minutes=45)
trigger.reschedule_on_finish=True

job = Job(
    name=job_name,
    trigger_type='interval',
    trigger=trigger,
    request_template=template,
    status="RUNNING",
    coalesce=True,
    max_instances=1
)

job_response = server.create_job(job)
print(job_response)
```

hazmat345 (changes requested):
Couple things:

I added a commit putting the namespace back in `TEMPLATE_FIELDS`. The subsequent definition should overwrite it when creating the Request class. That way we don't have to subtract one in that test.

I also checked the brewtils schemas / models, it doesn't look like we need to make any change there, which is nice.

However, I think it's is broken. These changes make it possible to create a job with a request template with no namespace. However, when I actually did that the scheduler broke. There's no easy way to do this from the UI, so I modified scheduler.create_job to manually set it to None:

```python
def create_job(job: Job) -> Job:
    job.request_template.namespace = None

    # Save first so we have an ID to pass to the scheduler
    job = db.create(job)

    return job
```

I think it's failing because the scheduler `run_job` just assumes that requests will be targeting the local garden:

```python
    request = process_request(request_template, wait_event=wait_event)
```

So it's not actually run through the router where we have the logic to figure out a missing namespace.

TheBurchLog:
@hazmat345 
Updated to route through the router. We should be good now.

hazmat345 (approved):
LGTM

PR #960, opened 2021-04-11
Database layer serializes to string when converting to brewtils

### LINKED ISSUE #959: Database layer holding weak references
Trying to change a nested dictionary that's been pulled out of mongo is giving me an error about changing a weak reference that's been removed.

hazmat345:
Fixes #959 - an issue with the database layer incompletely converting to brewtils models.


TheBurchLog (approved):
Can't find a way to break it, so it should all work. The QuerySet acts just like a list. So it shouldn't hit any issues down the road.

PR #962, opened 2021-04-12
Issue/912

hazmat345:
Fixes #912, groups garden connection parameters.

There's currently no migration for existing garden connections. You'll have to re-enter any prior connection information you had saved for a child garden. I don't really think it's worth the hassle of trying to write a migration.

Should be merged after #961.

PR #970, opened 2021-04-13
Issue/969

### LINKED ISSUE #969: Stomp Headers
Stomp headers are not getting sent when forwarding to child.

hazmat345:
Fixes #969, stomp headers.

### DIFF
--- a/src/app/beer_garden/router.py
+++ b/src/app/beer_garden/router.py
@@ -563,7 +563,8 @@ def _forward_stomp(operation: Operation, target_garden: Garden):
         message, response_headers = process(operation)
 
         response_headers = consolidate_headers(
-            response_headers, target_garden.connection_params.get("stomp_headers")
+            response_headers,
+            target_garden.connection_params.get("stomp", {}).get("headers", {})
         )
 
         stomp_garden_connections[target_garden.name].send(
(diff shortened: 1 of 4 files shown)


src/app/beer_garden/router.py
```diff
@@ -556,20 +562,18 @@ def _determine_target_garden(operation: Operation) -> str:
...
 
-        message, response_headers = process(operation)
+        if not conn.is_connected() and not conn.connect():
+            raise ConnectFailedException()
 
-        response_headers = consolidate_headers(
-            response_headers, target_garden.connection_params.get("stomp_headers")
-        )
+        header_list = target_garden.connection_params["stomp"].get("headers", {})
+        conn_headers = parse_header_list(header_list)
```
TheBurchLog, line 571: Was the issue that we were not parsing the headers? Previously the headers were grabbed from the connection parameters. I am curious what was it trying to jam into the headers field since it didn't know what the Key/Value properties were.

TheBurchLog (approved):

src/app/beer_garden/router.py
hazmat345: I'm not sure, but I think this problem was probably caused by #962, where I missed this part when switching to the nested connection parameters.

Before this PR headers taken from the config were strings, so they needed to be parsed. Headers taken from the database (aka, this section of code) weren't strings, so they didn't need to be parsed the same way. I'm not sure if they were being set correctly, though.

PR #971, opened 2021-04-13
Issue/949

### LINKED ISSUE #949: STOMP entry point unable to route to newly created Systems
When deploying to a clean environment, the Router for STOMP is not being populated.

Things observed:
- The router relies on the System.py to update the routing table. This does not scale across processes
- If the router is not updated in the System.py, then there is a race condition with plugins where the plugin fails to initialize.
- Attempted to have the router `handle_event` class handle it regardless, and that didn't work.
- If you reboot Beer Garden it works as expected
- If you shutdown Beer Garden and delete the database, then start back up Beer Garden, it works as expected.
- The error message is not helpful to let the end user know that it was a routing error.
  - `2021-04-05 17:38:01,080 - beer_garden.api.stomp.transport - WARNING - Error parsing and routing message: 'docker:echo-3.0.0.dev0'`
- The STOMP entry point is receiving events from the HTTP entry point because in the Unit Tests we can publish over HTTP and capture the `REQUEST_CREATED` and `REQUEST_COMPLETED` events on the STOMP topic.

The only way I could get this to work is to add `setup_routing()` in the router.py class prior to looking up the system mapping for a `REQUEST_CREATE` (around line 550).

hazmat345:
Fixes #949.

Doesn't improve error handling. That's going to require extensive changes that are going to have to wait until after the auth changes.

### DIFF
--- a/src/app/beer_garden/api/http/__init__.py
+++ b/src/app/beer_garden/api/http/__init__.py
@@ -369,7 +369,6 @@ def _event_callback(event):
     websocket_publish(event)
 
     # And also register handlers that the entry point needs to care about
-    # As of now that's only the routing subsystem
     for handler in [
         beer_garden.router.handle_event,
         beer_garden.log.handle_event,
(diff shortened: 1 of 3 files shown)


TheBurchLog (approved):
Routing now works for STOMP for new Plugins added to BG

PR #977, opened 2021-04-14
Wrapping all stomp handler calls in a try

hazmat345:
This wraps the event handling for the stomp entry point in a try so it doesn't cause problems if there's ever an exception.


TheBurchLog (approved):
Looks good to me

Issue #978, opened 2021-04-14
Events mangling event objects


hazmat345:
We should pass a copy of the event to each event handler, rather than the original. We're seeing issues with the garden connection params being modified in one place affecting the event in another.

scott-taubman:
If you look at [beer_garden/events/handlers.py](https://github.com/beer-garden/beer-garden/blob/develop/src/app/beer_garden/events/handlers.py#L48), we pass the original event in to a bunch of different handlers.  Those handlers sometimes modify the event, which means that what gets passed into any subsequent handler is no longer the original event.  Based on this issue it seems that that is not the intention.  Making a copy of the event first and passing that in to each handler should fix this.

**However**, I don't know that we can say with any confidence that the handlers today haven't come to depend on the fact that some earlier handler modified the event in some way. With the state of the testing, I highly doubt such a dependency would be revealed via some test failure.  We should therefore at least do a cursory examination of each of the handlers that gets called to at least do a base level assessment of whether we think each handler could work in isolation of all of the others.

Additionally, there are similar handler loops in the following:

* beer_garden/api/http/__init__.py
* beer_garden/api/stomp/manager.py

We should update those as well and do the same analysis on any of the handlers than are being called.

Jake-Ross:
The handlers present in `handlers.py`, on a surface level, seem to not have anything that might rely on information that has changed from other handlers. In fact, reversing the order the handlers are run doesn't seem to affect anything (unit tests all pass, making jobs/requests/other things in the UI run normally), so I'd say this is a safe change. The same can be said for the other two files that have been tagged.

PR #979, opened 2021-04-14
Deep copying connection parameters before modifying them

hazmat345:
This should help with #978. It only addresses one specific case though, so it doesn't fix the issue.


TheBurchLog (approved):
This looks good, but I had an idea... What if we did a deep copy on every Event prior to it being executed? So add this to the Processors?

hazmat345:
> This looks good, but I had an idea... What if we did a deep copy on every Event prior to it being executed? So add this to the Processors?

That's what I was thinking as well

PR #985, opened 2021-04-16
Child request view starts as collapsed on request view page

### LINKED ISSUE #984: Don't expand child requests by default
If you have a ton of child requests then expanding the view in the UI causes the browser to freeze. Since the view is expanded by default this is a big problem.

This ticket is just for changing the default to collapsed.

hazmat345:
Fixed #984. Child request view now starts collapsed.


TheBurchLog (approved):
That's the change I would have done

PR #996, opened 2021-04-24
Issue/991

### LINKED ISSUE #991: Bytes-type parameters
We currently support file parameters. Should explore what bytes parameters would look like.

hazmat345:
Fixes #991.

This moves the old files REST API back into vbeta - it shouldn't have been in v1. This API has also been renamed to "chunks", so its new location is /api/vbeta/chunks. This will still be used by `base64` parameters.

There is a new, simpler api at /api/vbeta/file. This is used by `bytes` parameters.

The long-term plan is to combine these into a single unified api.

### DIFF
--- a/src/app/beer_garden/api/http/handlers/vbeta/file.py
+++ b/src/app/beer_garden/api/http/handlers/vbeta/file.py
@@ -95,7 +95,9 @@ class RawFileListAPI(BaseHandler):
         db_file.file.put(io.BytesIO(self.request.body))
         db_file.save()
 
-        resolvable = Resolvable(storage="gridfs", details={"id": str(db_file.id)})
+        resolvable = Resolvable(
+            type="bytes", storage="gridfs", details={"id": str(db_file.id)}
+        )
         response = SchemaParser.serialize(resolvable, to_string=True)
 
         self.set_header("Content-Type", "application/json; charset=UTF-8")
--- a/src/app/beer_garden/api/http/handlers/vbeta/file.py
+++ b/src/app/beer_garden/api/http/handlers/vbeta/file.py
@@ -17,7 +17,7 @@ class RawFileAPI(BaseHandler):
         summary: Retrieve a File
         parameters:
           - name: file_id
-            in: body
+            in: path
             required: true
             description: The file ID
             type: string
--- a/src/app/beer_garden/api/http/handlers/vbeta/file.py
+++ b/src/app/beer_garden/api/http/handlers/vbeta/file.py
@@ -91,9 +91,8 @@ class RawFileListAPI(BaseHandler):
         tags:
           - Files
         """
-        name = self.get_argument("name", default="")
-        db_file = RawFile(name=self.get_argument("name", default=""))
-        db_file.file.put(io.BytesIO(self.request.body), filename=name)
+        db_file = RawFile()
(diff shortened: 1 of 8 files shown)


hazmat345:
To exercise this:

```
curl --upload-file ./binary http://localhost:2337/api/vbeta/file\?name\=unique_name
curl http://localhost:2337/api/vbeta/file\?name\=unique_name
curl -X DELETE http://localhost:2337/api/vbeta/file\?name\=unique_name
```

src/app/beer_garden/api/http/handlers/vbeta/file.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+class RawFileAPI(BaseHandler):
+    @authenticated(permissions=[Permissions.READ])
+    async def get(self):
+        """
+        ---
+        summary: Retrieve a File
+        parameters:
+          - name: file_id
```
loganasherjones, line 18: I'm assuming this will become part of the URL `/api/vbeta/file/<id>`

src/app/beer_garden/api/http/handlers/vbeta/file.py
```diff
@@ -0,0 +1,102 @@
...
+          400:
+            $ref: '#/definitions/400Error'
+          50x:
+            $ref: '#/definitions/50xError'
+        tags:
+          - Files
+        """
+        name = self.get_argument("name", default="")
+        db_file = RawFile(name=self.get_argument("name", default=""))
+        db_file.file.put(io.BytesIO(self.request.body), filename=name)
```
loganasherjones, line 70: This will work for small files (<100mB).

In order to make something work we'd have to do some kind of chunked uploads for large files.

src/app/beer_garden/api/http/handlers/vbeta/file.py
```diff
@@ -0,0 +1,102 @@
...
+          50x:
+            $ref: '#/definitions/50xError'
+        tags:
+          - Files
+        """
+        db_file = RawFile.objects.get(id=self.get_argument("file_id", default=""))
+        file = db_file.file.read()
+
+        self.set_header("Content-Type", "application/octet-stream")
+        self.write(file)
```
loganasherjones, line 39: Again, this could be a problem for large files, it'd be nice to have another API for "streaming" chunks back. I think GridFS already stores thinks in chunks (at every 4MB) so maybe that's our chunk size?

src/app/beer_garden/db/mongo/models.py
```diff
@@ -757,3 +757,8 @@ class FileChunk(MongoModel, Document):
     data = StringField(required=True)
     # Delete Rule (2) = CASCADE; This causes this document to be deleted when the owner doc is.
     owner = LazyReferenceField(File, required=False, reverse_delete_rule=CASCADE)
+
+
+class RawFile(Document):
+    name = StringField()
```
loganasherjones, line 763: Weird question, do we actually need a name field?

As of right now it adds no value.

src/app/beer_garden/api/http/handlers/vbeta/file.py
hazmat345: 👍🏼

PR #1039, opened 2021-06-08
Better check for when requests need to be stored in GridFS

### LINKED ISSUE #1038: Mongo Request model gridfs storage
**Original Description saved for posterity:**
The `save` method looks off - I don't think this is working as intended.

**Updated Description:**
The handling of large requests and in particular the way we're using gridfs is lamentable.  Even so, it certainly should be possible to fail less often with relatively minor changes.

The big problem right now (which the linked PR attempts, but fails, to fix) is that if the parameters are what push the document over the 16MB limit, we don't handle that correctly. The current code only even looks at the parameters if there is output, meaning that initial creation  of the request (before there is output) is doomed to fail if the parameters are large.  

The other problem is the use of `sys.getsizeof` which is not going to return the correct value, particularly for the parameters, which are a dict-like object.  I would suggest the following:

1) Put a 5MB cap on `parameters` and `output` before switching over to storing them in gridfs.  Don't bother checking the total document size.  If neither one of them is allowed to be over 5MB, worst case scenario is that the two together, if they are just under the cap, are just under 10MB.  That leaves 6MB of headroom for the rest of the document (comments, etc).

2) Instead of `sys.getsizeof`, just use `len`.  For the parameters that will mean converting to a string first, so `len(str(self.parameters))`.  That should be a simple, accurate enough check.

3) Write unit tests.  The PR that's out there now doesn't work and unit tests would have made that obvious.

This isn't a perfect solution by any means, but it should lead to failure much less often with much lower risk than a broader rewrite of the large request handling.

hazmat345:
This fixes #1038 - cleans up the large request GridFS failover.

### DIFF
--- a/src/app/beer_garden/db/mongo/models.py
+++ b/src/app/beer_garden/db/mongo/models.py
@@ -396,38 +396,46 @@ class Request(MongoModel, Document):
     logger = logging.getLogger(__name__)
 
     def pre_serialize(self):
+        """Pull any fields out of GridFS"""
         encoding = "utf-8"
-        """If string output was over 16MB it was spilled over to the GridFS storage solution"""
+
         if self.output_gridfs:
-            self.logger.debug("~~~Retrieving output from gridfs")
+            self.logger.debug("Retrieving output from GridFS")
             self.output = self.output_gridfs.read().decode(encoding)
             self.output_gridfs = None
 
         if self.parameters_gridfs:
-            self.logger.debug("~~~Retrieving parameters from gridfs")
+            self.logger.debug("Retrieving parameters from GridFS")
             self.parameters = json.loads(self.parameters_gridfs.read().decode(encoding))
             self.parameters_gridfs = None
 
     def save(self, *args, **kwargs):
-        self.updated_at = datetime.datetime.utcnow()
-        max_size = 15 * 1_000_000
-        encoding = "utf-8"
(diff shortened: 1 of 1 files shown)


src/app/beer_garden/db/mongo/models.py
```diff
@@ -396,38 +396,46 @@ class Request(MongoModel, Document):
...
+        request_size = 0
         encoding = "utf-8"
-        parameter_size = sys.getsizeof(self.parameters)
 
-        if self.output:
-            parameter_size = sys.getsizeof(self.parameters)
-            if parameter_size > max_size:
-                self.logger.info("~~~Parameter size too big, storing in gridfs")
+        if self.parameters:
+            request_size += sys.getsizeof(self.parameters)
```
scott-taubman, line 426: This won't work.  getsizeof doesn't actually account for the size of the values in the dict.  See [this random blog](https://nedbatchelder.com/blog/202002/sysgetsizeof_is_not_what_you_want.html) for a decent explanation. The end result is that the check below will basically always return False, even if request_size is huge.

scott-taubman:
Closing this PR as it doesn't quite address the issue.  I've outlined another possible approach in the issue description.

jlrpnbbngtn:
> Closing this PR as it doesn't quite address the issue. I've outlined another possible approach in the issue description.

Link to issue/1038 404's.

PR #1068, opened 2021-07-09
Republishing a modified event when child request changes

### LINKED ISSUE #1057: Child Request status updates in UI
The frontend is having issues updating child request status changes. Pretty sure it's because we rip off the parent ID before sending the request to the child garden, and then we're publishing the event generated *by* that child garden, which obviously won't include that parent ID.

Then the frontend looks for the parent ID:

```javascript
    else if (_.get(event, 'payload.parent.id') == $stateParams.requestId) {
        <this is a child request, update the page>
    }
```

hazmat345:
Fixes #1057 - child requests on a downstream garden should now be properly updated in the UI.

## Test Instructions
- Set up parent and child gardens named 'parent' and 'child'
- Run the `echo` example plugin on the parent garden
- Run the `sleeper` example plugin on the child garden
- Modify the `echo-sleeper` plugin so that the SystemClient targets the sleeper plugin on the child garden:
```python
        self.sleeper_client = SystemClient(system_namespace="child", system_name="sleeper")
```
- Use the parent garden UI to create a `say_sleep` request
- On the request view page click the request to expand the child request view

Before this PR the status of the child request targeted at the parent garden would appear and update normally. However, the child request targeted at the child garden would either fail to appear or not update as the status changes.

This PR corrects this - now both child requests should appear and update normally.

### DIFF
--- a/src/app/beer_garden/requests.py
+++ b/src/app/beer_garden/requests.py
@@ -754,6 +754,12 @@ def invalid_request(request: Request = None):
     return request
 
 
+@publish_event(Events.REQUEST_UPDATED)
+def update_request(request: Request):
+    db.update(request)
+    return request
+
+
 def process_wait(request: Request, timeout: float) -> Request:
     """Helper to process a request and wait for completion using a threading.Event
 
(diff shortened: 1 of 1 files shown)


scott-taubman (approved):
A few concerns regarding the websocket event publishing have been raised and will be discussed further within the team.  These changes highlight the need for some of that discussion, but don't exacerbate the issues any, so it seems okay to merge this.

PR #1070, opened 2021-07-14
Issue/870

### LINKED ISSUE #870: _read_logs requests are not fully-formed
This was an issue during the 3.1 database migration.

_read_logs requests are missing some fields, most notably (for the DB migration) the `has_parent` field. This is ... concerning.

hazmat345:
This fixes #870, fixes a deficiency with log read requests.

One of the steps in creating a request is validation. There are a couple of things that we need to validate, things like "does a system with this name and version actually exist?"

The request validator has been around since the very beginning, and one of my pet peeves with it is that it modifies the request object. One of the steps involved in validation is "does this request have a parent request that's already complete?" A side-effect of this validation was setting the `has_parent` field of the request.

Because the log read request is an admin-type request we don't to subject it to the same validation as a "regular" request. This is normally fine since it's hard-coded ahead of time - there's nothing dynamic to validate.

These two things combined to cause the issue we were seeing. To correct I've moved the `has_parent` determination to a more general location in the database model and out of the request validator.

## Test Instructions
- Create a log read request
- Look at the request that was just created, it should now have a `has_parent` field

### DIFF
--- a/src/app/beer_garden/db/mongo/models.py
+++ b/src/app/beer_garden/db/mongo/models.py
@@ -460,15 +460,10 @@ class Request(MongoModel, Document):
         # Deal with has_parent
         if self.has_parent is None:
             self.has_parent = bool(self.parent)
-        elif self.has_parent is False and self.parent:
+        elif self.has_parent != bool(self.parent):
             raise ModelValidationError(
-                f"Can not save Request {self}: Request has a parent field but "
-                f"has_parent is False"
-            )
-        elif self.has_parent is True and not self.parent:
-            raise ModelValidationError(
-                f"Can not save Request {self}: Request does not have a parent field "
-                f"but has_parent is True"
+                f"Cannot save Request {self}: parent value of {self.parent!r} is not "
+                f"consistent with has_parent value of {self.has_parent}"
             )
 
     def clean_update(self):
--- a/src/app/beer_garden/db/mongo/models.py
+++ b/src/app/beer_garden/db/mongo/models.py
@@ -409,6 +409,10 @@ class Request(MongoModel, Document):
             self.parameters_gridfs = None
 
     def save(self, *args, **kwargs):
+        # Set has_parent if not already set
+        if self.has_parent is None:
+            self.has_parent = bool(self.parent)
+
         self.updated_at = datetime.datetime.utcnow()
         max_size = 15 * 1_000_000
         encoding = "utf-8"
--- a/src/app/beer_garden/db/mongo/models.py
+++ b/src/app/beer_garden/db/mongo/models.py
@@ -409,10 +409,6 @@ class Request(MongoModel, Document):
             self.parameters_gridfs = None
 
     def save(self, *args, **kwargs):
-        # Set has_parent if not already set
-        if self.has_parent is None:
-            self.has_parent = bool(self.parent)
-
         self.updated_at = datetime.datetime.utcnow()
         max_size = 15 * 1_000_000
         encoding = "utf-8"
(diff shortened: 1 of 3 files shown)


src/app/beer_garden/db/mongo/models.py
```diff
@@ -409,6 +409,10 @@ def pre_serialize(self):
             self.parameters_gridfs = None
 
     def save(self, *args, **kwargs):
+        # Set has_parent if not already set
+        if self.has_parent is None:
```
scott-taubman, line 413: Does it even make sense that `has_parent` is a thing that you can set on the `Request`? As is, and seemingly unchanged here, one could instantiate a Request that sets a `parent`, but then explicitly sets `has_parent` to `False`.  Is that a valid / desired state for the Request to be in?

Would it make more sense to just have the `Request.__init__` set `has_parent` based on the value of `parent`? Or maybe not even write down `has_parent` and just have it be a computed value? There are things like this elsewhere in the code:

```python
# Clear parent before forwarding so the child doesn't freak out about an                                                                                                                                   
# unknown request                                                                                                                                                                                          
operation.model.parent = None                                                                                                                                                                              
operation.model.has_parent = False
```

It feels a bit like having `has_parent` as its own value is just a recipe for it not getting set at some point when its supposed to and thus being out of sync with reality.

scott-taubman (changes requested):
Had one question about the nature of `has_parent` in general. Plus, as always, I'll advocate the creation of unit tests for such changes.

src/app/beer_garden/db/mongo/models.py
hazmat345: Yeah, this is pretty gross. The only reason `has_parent` exists at all is because we need to construct some compound indexes in mongo to prevent our request index page from being too slow to be usable. And you can't construct an index from a reference field, or at least you couldn't when we first needed this.

So yeah, there are weird states you can get in that don't make sense (for example, the one this fixes). But I think we need to keep `has_parent` as a "real" database field, unfortunately.

scott-taubman: Those compound indexes you're referring to look like they could use some love. I suspect that there's some work that could be done there that would obviate the need for this whole `has_parent` thing, but in the interest of not blowing up the scope of your work here, let's carry on under the presumption that the field needs to be written down.

It still seems like we should protect ourselves from setting this incorrectly.  Is the validation that you refer to that happens on non-Admin requests (the `RequestValidator` stuff you altered here) unrelated to the mongoengine document model validation that happens?  It looked like it might be.  If that's the case, I'd suggest:

* This sort of validation and modification seems to belong in the [clean](https://docs.mongoengine.org/guide/validation.html#custom-validation) rather than the `save` method.
* You should raise a `ModelValidationError` if `has_parent` is `False` while `parent` is actually set.

scott-taubman (changes requested):
Replied with some suggestions.  Also, how do you feel about writing some tests?

src/app/beer_garden/db/mongo/models.py
hazmat345: Moved into `clean`.

Yep, those indexes are ugly. But they're the only way I could find to have any sort of reasonable performance on the request index page once you start filtering and sorting. We need to support ~750k requests and if mongo has to resort to a COLSCAN it'll take 10s of seconds.

src/app/beer_garden/db/mongo/models.py
```diff
@@ -457,6 +457,20 @@ def clean(self):
                 f"Can not save Request {self}: Invalid output type '{self.output_type}'"
             )
 
+        # Deal with has_parent
+        if self.has_parent is None:
+            self.has_parent = bool(self.parent)
+        elif self.has_parent is False and self.parent:
```
scott-taubman, line 463: You could combine the two else blocks here no?

```python
elif self.has_parent != bool(self.parent):
    raise ModelValidationError(
        f"Cannot save Request {self}: parent value of {self.parent} is not consistent with "
        f"has_parent value of {self.has_parent}"
    )
```

Something like that.

hazmat345: Sure, that's fine.

scott-taubman (approved):

PR #1071, opened 2021-07-14
#1063 - Stomp integration tests are failing

### LINKED ISSUE #1063: Stomp integration tests are failing
The Stomp portion of the integration tests has been failing for some time. See an example in the github actions that run nightly [here](https://github.com/beer-garden/beer-garden/runs/3004160469?check_suite_focus=true).

The problem is twofold:

* The "stomp" test module name conflicts with the name of the actual "stomp" module we're using.  This is easily remedied by renaming the test module (something like "_stomp" would do the trick).
* Once the name conflict is fixed, the unit tests themselves will still actually fail.  We'll want to track down the issue and get them working again.

jlrpnbbngtn:
Closes #1063 

These changes address the failure of the stomp integration tests run by GitHub actions nightly. Items addressed:

* Renamed `stomp` test module to `stomp_` in order to avoid collision with the actual `stomp.py` package
* Changed base class of listener to match what Stomp expects
* Added property to listener, replacing class variable
* Added guard code to listener for the case where its method is called with a bad value
* Added needed decorator to test class
* Refactored ordering of test method so the logic is clearer
* Added files created by the *act* runner to `.gitignore`
* General code hygiene (comments refactored, imports cleaned up, dead code removed, type hints added)

## Test Instructions
* Sadly, because these changes seek to address the test runs on GitHub, the only *true* way to verify they are correct and satisfy the requirement is to wait for them to run on GitHub...
* That aside, the updated integration tests were verified to no longer fail when run on the local system and also on a simulated GitHub Actions environment via `act` (alas, with pinned versions of everything, so no matrix of versions).


jlrpnbbngtn:
Closing because I neglected to rebase back to dev :(

PR #1072, opened 2021-07-14
 Stomp integration tests are failing

### LINKED ISSUE #1063: Stomp integration tests are failing

jlrpnbbngtn:
Closes #1063

These changes address the failure of the stomp integration tests run by GitHub actions nightly. Items addressed:

* Renamed stomp test module to stomp_ in order to avoid collision with the actual stomp.py package
* Changed base class of listener to match what Stomp expects
* Added property to listener, replacing class variable
* Added guard code to listener for the case where its method is called with a bad value
* Added needed decorator to test class
* Refactored ordering of test method so the logic is clearer
* Added files created by the act runner to .gitignore
* General code hygiene (comments refactored, imports cleaned up, dead code removed, type hints added)

## Test Instructions
*  Sadly, because these changes seek to address the test runs on GitHub, the only true way to verify they are correct and satisfy the requirement is to wait for them to run on GitHub...
* That aside, the updated integration tests were verified to no longer fail when run on the local system and also on a simulated GitHub Actions environment via act (alas, with pinned versions of everything, so no matrix of versions).


test/integration/stomp_/publisher/create_request_test.py
```diff
@@ -1,36 +1,14 @@
...
+from brewtils.rest.easy_client import EasyClient
 from brewtils.schema_parser import SchemaParser
-import pytest
-import time
-import json
-
-from brewtils.errors import ValidationError
-
-try:
-    from helper import delete_plugins
```
jlrpnbbngtn, line 13: None of these were in use. But even if they had been used, the syntax appears to be wrong. `helper` cannot be imported directly, as far as I can tell. The relative import statement is the correct one (if `helper` code was needed in this file).

test/integration/stomp_/publisher/create_request_test.py
```diff
@@ -43,47 +21,71 @@ def system_spec():
     }
 
 
-class MessageListener(object):
-    create_event_captured = False
+class MessageListener(ConnectionListener):
```
jlrpnbbngtn, line 24: Since this is being passed to stomp, it makes sense that it should inherit from the type that stomp expects.

test/integration/stomp_/publisher/create_request_test.py
```diff
@@ -43,47 +21,71 @@ def system_spec():
...
-                        print(message)
-                        self.create_event_captured = True
+        Args:
+            headers: a dictionary containing all headers sent by the server
+            as key/value pairs.
+
+            message: the frame's payload - the message body.
+        """
+        try:
+            if "model_class" in headers:
```
jlrpnbbngtn, line 47: Though it's probably a fair assumption that this will never be called with an incorrect argument, this allows a more meaningful message to be produced if the passed dictionary doesn't have the key we're looking for.

test/integration/stomp_/publisher/create_request_test.py
```diff
@@ -118,8 +112,23 @@ def test_listen_create_request(self):
                 "durable-subscription-name": "events",
             },
         )
+        assert stomp_connection.get_listener("bg_listener") == listener
```
jlrpnbbngtn, line 115: Reordered so that variable creation and use are close to each other. Also added some additional asserts so if this fails because of a step before the last, it will be clear where the failure happened.

test/integration/stomp_/publisher/create_request_test.py
```diff
@@ -130,7 +139,7 @@ def test_listen_create_request(self):
 
     @pytest.mark.usefixtures("easy_client", "request_generator")
     def test_publish_create_request(self):
-        """Published the Request over STOMP and verifies of HTTP"""
+        """Publish a Request over STOMP and verify it via HTTP."""
```
jlrpnbbngtn, line 142: I know this is very nitpicky, but the pythonic way to write doc strings is for the first line to begin with a verb in the imperative mood. `pydocstyle` enforces this. We're not using it, but maybe we should (?)

scott-taubman: I'm generally in favor of abdicating all styling decisions to some tool that will enforce it for us.  So I'm down with `pydocstyle`, though at first glance theres...a lot of work ahead of us to become compliant.

test/integration/stomp_/publisher/create_request_test.py
```diff
@@ -43,47 +21,71 @@ def system_spec():
...
                     else:
-                        print(parsed.model.name)
-            elif headers["model_class"] == "error_message":
-                print("Error Message Returned:", message)
-        except:
-            print("Error: unable to parse message:", message)
+                        print("ERROR: no parsed model found")
+                elif headers["model_class"] == "error_message":
+                    print(f"ERROR: Message returned: {message!r}")
+                else:
```
scott-taubman, line 60: I think this else statement and the corresponding print message are in one level too far based on your above stated goal.  That is, this will print if "model_class" is present, but isn't one of the value you're checking for.

Also, could you have achieved the same thing by just outputting the exception message in the Exception block?  That would give you the key error when "model_class" isn't present, but then also be useful if some other, unexpected exception is actually what gets thrown.

test/integration/stomp_/publisher/create_request_test.py
```diff
@@ -43,47 +21,71 @@ def system_spec():
...
+                    print(f"ERROR: 'model_class' not in headers, message={message}")
+        except Exception:
+            print(f"ERROR: unable to parse, message={message}")
 
+    @property
+    def create_event_captured(self):
+        return self._create_event_captured
 
+    @create_event_captured.setter
```
scott-taubman, line 69: I'm not going to pretend I care all that much, but is there a reason for this?  It doesn't look like anything outside of this class tries to set `_create_event_captured` and inside the class it should be fair game to just set it directly.

scott-taubman (changes requested):
Mostly LGTM.  I even created a branch from this and changed the run condition on the action workflow just to verify that a full run of the integration tests passes, which [it does](https://github.com/beer-garden/beer-garden/actions/runs/1034010488).

The only one of the comments I'm guessing you'll want to address for sure is the misplaced print statement.  The setter one I'll leave to your discression.

test/integration/stomp_/publisher/create_request_test.py
```diff
@@ -43,47 +21,71 @@ def system_spec():
...
+        try:
+            if "model_class" in headers:
+                if headers["model_class"] == "Operation":
+                    parsed = SchemaParser.parse_operation(message, from_string=True)
+
+                    if parsed.model and parsed.model.name:
+                        if parsed.model.name.startswith("REQUEST"):
+                            self.create_event_captured = True
+                        else:
+                            print(f"ERROR: no 'REQUEST' is parsed model")
```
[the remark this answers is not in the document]
jlrpnbbngtn: Considering how short the ignore cases are in the setup.cfg, it's surprising that this error in particular was chosen for omission.

test/integration/stomp_/publisher/create_request_test.py
jlrpnbbngtn: Duly noted on the print.

As for you other point, I would have not written it this way, but I tried to preserve the intent and form of the original and add some extra guards I think are appropriate for the form it's in.

test/integration/stomp_/publisher/create_request_test.py
jlrpnbbngtn: The reason is that this is the idiom I'm accustomed to and prefer.

scott-taubman (approved):
:+1:

PR #1075, opened 2021-07-21
Issue/958 - Added config option for changing commands of non-dev systems

### LINKED ISSUE #958: Disable Disallow System Command Changes
The current behavior of not allowing people to change the commands of a non-dev version system is ... questionable. We should at least provide a config option to disable this behavior.

Jake-Ross:
Fixes #958. I'm unsure if `allow_update_systems` might be too vague of a config option. The description in `config.py` suffices, though.

Tested with brewtils' `update_system`.

### DIFF
--- a/src/app/beer_garden/config.py
+++ b/src/app/beer_garden/config.py
@@ -1261,10 +1261,16 @@ _PLUGIN_SPEC = {
                 },
             },
         },
-        "allow_update_systems": {
+        "allow_system_command_updates": {
             "type": "bool",
             "default": False,
-            "description": "Allow updating the commands of a a non-dev system",
+            "description": "Allow commands of non-dev systems to be updated",
+            "long_description": "When False, this prevents changes to the command "
+            "definitions of a registered version of a system. This means that the "
+            "system will fail to start if the commands do not match what is on "
+            "record for that version of the system. When True, the system will be "
+            "allowed to start and the commands on record will be updated accordingly. "
+            "NOTE: System versions containing 'dev' are exempt from this check.",
         },
         "status_heartbeat": {
             "type": "int",
--- a/src/app/beer_garden/config.py
+++ b/src/app/beer_garden/config.py
@@ -1261,7 +1261,7 @@ _PLUGIN_SPEC = {
                 },
             },
         },
-        "allow_update_system": {
+        "allow_update_systems": {
             "type": "bool",
             "default": False,
             "description": "Allow updating the commands of a a non-dev system",
--- a/src/app/beer_garden/config.py
+++ b/src/app/beer_garden/config.py
@@ -1261,6 +1261,11 @@ _PLUGIN_SPEC = {
                 },
             },
         },
+        "allow_update_system": {
+            "type": "bool",
+            "default": "false",
+            "description": "Allow updating the commands of a a non-dev system", 
+        },
         "status_heartbeat": {
             "type": "int",
             "default": 10,
--- a/src/app/beer_garden/config.py
+++ b/src/app/beer_garden/config.py
@@ -1261,7 +1261,7 @@ _PLUGIN_SPEC = {
                 },
             },
         },
-        "allow_system_command_updates": {
+        "allow_command_updates": {
             "type": "bool",
             "default": False,
             "description": "Allow commands of non-dev systems to be updated",
--- a/src/app/beer_garden/config.py
+++ b/src/app/beer_garden/config.py
@@ -1263,8 +1263,8 @@ _PLUGIN_SPEC = {
         },
         "allow_update_system": {
             "type": "bool",
-            "default": "false",
-            "description": "Allow updating the commands of a a non-dev system", 
+            "default": False,
+            "description": "Allow updating the commands of a a non-dev system",
         },
         "status_heartbeat": {
             "type": "int",
--- a/src/app/test/systems_test.py
+++ b/src/app/test/systems_test.py
@@ -1,0 +1,44 @@
+# -*- coding: utf-8 -*-
+import pytest
+from mock import Mock
+from mongoengine import connect
+from brewtils.models import System, Command
+from brewtils.errors import ModelValidationError
+
+import beer_garden
+import beer_garden.router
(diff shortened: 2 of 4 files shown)


src/app/beer_garden/config.py
```diff
@@ -1261,6 +1261,11 @@ def _parse_args(args: Sequence[str]) -> Tuple[YapconfSpec, dict]:
                 },
             },
         },
+        "allow_update_systems": {
```
scott-taubman, line 1264: I would call this `allow_system_command_updates`, since it's really just the commands that are prevented from changing.  There are other properties of a system that can be updated regardless of this setting.

Also, I would add a `long_description` here. Something like:
> When False, this prevents changes to the command definitions of a registered version of a system. This means that the system will fail to start if the commands do not match what is on record for that version of the system.  When True, the system will be allowed to start and the commands on record will be updated accordingly. NOTE: System versions containing "dev" are exempted from this check.

src/app/beer_garden/config.py
```diff
@@ -1261,6 +1261,11 @@ def _parse_args(args: Sequence[str]) -> Tuple[YapconfSpec, dict]:
...
+            "type": "bool",
+            "default": False,
+            "description": "Allow updating the commands of a a non-dev system",
```
scott-taubman, line 1267: You've got a double "a" here:  `a a non-dev` system.  I might just reword this as:

> Allow commands of non-dev systems to be updated

scott-taubman (changes requested):
Made a few minor comments related to naming and wording. Although it's a simple change, we should add a test for the expected behavior.  It doesn't look like there are existing unit tests for the systems code, which makes it a little more upfront work to write the tests than it ought to be, but it's still a good exercise worth going through.

Also, I figured out how to functionally test this, but see the developer SOP for an example of the type of test instructions that should be present on PRs. Effectively, describe to someone how they would validate, through use of the application, that the change you made does what is intended.

src/app/beer_garden/config.py
Jake-Ross: That's a much better config name. I tried to keep it succinct to fit with the rest of the config, but really it just makes its use less obvious.

Jake-Ross:
I drafted up two unit tests, but they make somewhat liberal use of patching. Considering the complexity of the `update_system` function, I figured it was unavoidable unless I wanted to split that function up into smaller pieces, but I didn't want to break something by doing that. I'd be glad to get feedback on how to write a less-hackey unit test.

Also, for the sake of documenting it on the PR, test instructions are as follows:
* Ensure `allow_system_command_updates` is `false` in your `config.yaml`.
* Use brewtils to send a system to your garden. I used an `EasyClient`, but I believe other methods are available.
* Call an `update_system` with new commands (they can be just a name) to that system, ensure you get a `ModelValidationError`.
* Repeat with `allow_system_command_updates=True` and ensure there no error is raised.
* You can also try `allow_system_command_updates=False` while your System version contains `dev`. Expected behavior for that is no model error.

hazmat345 (changes requested):
Can we change the name to just `allow_command_updates`? The enclosing spec names become a part of the way you specify the config value, so as-is it would be `BG_PLUGIN_ALLOW_SYSTEM_COMMAND_UPDATES`, which is a little verbose.

Otherwise looks good to me!

scott-taubman:
> Can we change the name to just `allow_command_updates`? The enclosing spec names become a part of the way you specify the config value, so as-is it would be `BG_PLUGIN_ALLOW_SYSTEM_COMMAND_UPDATES`, which is a little verbose.
> 
> Otherwise looks good to me!

I can live with `allow_command_updates`.  The practical distinction between a "Plugin" a "System" remains unclear to me, so possibly having both in the config name is redundant.

src/app/test/systems_test.py
```diff
@@ -0,0 +1,47 @@
...
+        name="original",
+        version="v0.0.1",
+        namespace="beer_garden",
+        commands=[Command(name="original")],
+    )
+    create_system(system)
+    monkeypatch.setattr(beer_garden.db.api, "modify", Mock())
+    monkeypatch.setattr(beer_garden.router, "add_routing_system", Mock())
+    try:
+        update_system(system=system, new_commands=[Command(name="changed_command")])
```
scott-taubman, line 29: For testing exceptions, you can do something like:

```python
       with pytest.raises(ModelValidationError):
            update_system(system=system, new_commands=[Command(name="changed_command")])
```

Rather than this try / except block.

src/app/test/systems_test.py
```diff
@@ -0,0 +1,47 @@
...
+        version="v0.0.1",
+        namespace="beer_garden",
+        commands=[Command(name="original")],
+    )
+
+    create_system(system)
+    monkeypatch.setattr(beer_garden.config, "get", lambda x: True)
+    monkeypatch.setattr(beer_garden.db.api, "modify", Mock())
+    monkeypatch.setattr(beer_garden.router, "add_routing_system", Mock())
+    update_system(system=system, new_commands=[Command(name="changed_command")])
```
scott-taubman, line 47: Seeing a test without any assertion always gives me pause.  Even if it's the case that everything executing successfully implies a successful test, it's not always clear to someone reading it that that was the intention.  Instead, make a specific assertion that makes the intent of the test explicit.

src/app/test/systems_test.py
```diff
@@ -0,0 +1,47 @@
...
+    monkeypatch.setattr(beer_garden.db.api, "modify", Mock())
+    monkeypatch.setattr(beer_garden.router, "add_routing_system", Mock())
+    try:
+        update_system(system=system, new_commands=[Command(name="changed_command")])
+        raise AssertionError()
+    except ModelValidationError:
+        pass
+
+
+def test_allow_system_command_updates(mongo_conn, monkeypatch):
```
scott-taubman, line 35: We should have docstrings that explain what we're testing, just for added clarity.

scott-taubman (changes requested):
Made some comments on the unit tests. We should talk about it a bit more.

PR #1077, opened 2021-07-22
Overriding operation target garden if necessary

### LINKED ISSUE #1076: Two-level garden routing
This was found during investigations for #1060.

Trying to create a request on a grandchild garden is broken. The routing procedure will not attempt to determine the correct garden to route an operation to if the operation already has a target garden.

Due to the way that we handle garden hierarchies this doesn't work. Gardens only know about their direct children gardens, so when the originating garden ("parent") determines that a request needs to be routed to a system on the grandchild garden ("grandchild") it *thinks* the target garden is actually it's direct child garden ("child").

This is half correct - the request does need to be sent to the child garden. However, that isn't the final destination, and marking it as such causes the child garden routing to fail since it's operating on incorrect information.

hazmat345:
This fixes #1076, multi-level garden routing.

Prior to this PR the routing process would not attempt to determine the target garden for an operation if one already existed on the model. Due to how we handle garden visibility, this meant that requests would never be able to be routed more than one garden away.

This PR tweaks that process to always attempt to determine the target garden.

## Test Instructions
- Set up a parent -> child -> grandchild garden structure
- From the parent, attempt to create a request for a system on the grandchild garden. It should work as normal

### DIFF
--- a/src/app/beer_garden/router.py
+++ b/src/app/beer_garden/router.py
@@ -43,6 +43,7 @@ import beer_garden.systems
 from beer_garden.api.stomp.transport import Connection, consolidate_headers, process
 from beer_garden.errors import (
     ForwardException,
+    RoutingException,
     RoutingRequestException,
     UnknownGardenException,
 )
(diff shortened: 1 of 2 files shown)


src/app/beer_garden/router.py
```diff
@@ -176,14 +176,28 @@ def route(operation: Operation):
         )
 
     # Determine which garden the operation is targeting
-    if not operation.target_garden_name:
-        operation.target_garden_name = _determine_target_garden(operation)
+    target_garden = _determine_target_garden(operation)
 
-    if not operation.target_garden_name:
+    if not target_garden and not operation.target_garden_name:
```
scott-taubman, line 181: I can't help but feel that there's a lot of work that comes after the `_determine_target_garden` call that looks very much related to determining the target garden.  Is there a reason why most of this logic shouldn't live there?

scott-taubman (changes requested):
Had one general question / comment, but also this seems like a thing that should be unit testable.  We should strive to add test coverage for these sorts of things.

src/app/beer_garden/router.py
hazmat345: Yeah, it should probably be renamed / organized better at this point.

scott-taubman (approved):

PR #1082, opened 2021-07-29
#940 - Change garden routing to allow any garden name

### LINKED ISSUE #940: Change garden endpoint routing to be less restrictive
We're hitting an error when a garden name contains a '-'. This is the same issue as https://github.com/beer-garden/beer-garden/pull/885, but with special characters instead of unicode.

I think always urlencoding the garden name before hitting the rest api should fix this.

Jake-Ross:
Closes #940 

This PR changes the routing regex for tornado to be less restrictive, allowing all characters. A further PR will be created to always urlencode garden names before routing so that users won't be able to create invalid garden names.

One minor issue is that the `(.*)` group will capture the ending slash of a url, if present. For example: 
In `.../api/v1/gardens/garden-name/`, the captured group will be `garden-name/` instead of `garden-name`.

This hasn't affected my ability to access the garden in any way, but I can't figure out what that group is used for. If it's an issue, maybe `([^/]+)/?` will be a better regex here, to include all characters except for `/`.

To test: 
- Create a garden with special characters in the name, such as `-`, `&`, or `%`.

### DIFF
--- a/src/app/beer_garden/api/http/__init__.py
+++ b/src/app/beer_garden/api/http/__init__.py
@@ -199,7 +199,7 @@ def _setup_tornado_app() -> Application:
         (rf"{prefix}api/v1/tokens/(\w+)/?", v1.token.TokenAPI),
         (rf"{prefix}api/v1/jobs/(\w+)/?", v1.job.JobAPI),
         (rf"{prefix}api/v1/logging/?", v1.logging.LoggingAPI),
-        (rf"{prefix}api/v1/gardens/([\w%]+)/?", v1.garden.GardenAPI),
+        (rf"{prefix}api/v1/gardens/(.*)/?", v1.garden.GardenAPI),
         # Beta
         (rf"{prefix}api/vbeta/events/?", vbeta.event.EventPublisherAPI),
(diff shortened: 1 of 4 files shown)


Jake-Ross:
Latest push fixes being able to create two gardens with the same name due to bad validity checking. Names are now decoded when displayed on the garden admin page. This also removes the 'feature' that prevents deleting gardens with invalid names.

To test:
- On the `Admin>Gardens` tab, ensure you can create/delete/edit gardens with special characters including, but not limited to `/ \ @ # $ % ^ ₩`
- Ensure you cannot create two gardens with the same name

hazmat345:
> This hasn't affected my ability to access the garden in any way, but I can't figure out what that group is used for. If it's an issue, maybe `([^/]+)/?` will be a better regex here, to include all characters except for `/`.

Groups in those regexes supply positional args to the handler that processes the request. So for example, the `api.http.handlers.v1.garden.GardenAPI` implements a `get` method with a `garden_name` parameter. The result of the capture group will be supplied as that parameter.

jlrpnbbngtn:
Tested with all the usual ASCII symbols, but also with a mix of ASCII symbols, ASCII alphanumeric and Unicode characters. No problems found. Code looks good.

PR #1086, opened 2021-08-03
Issue/1085

### LINKED ISSUE #1085: Command-based choices error indication
Currently the error indication is not displaying if a command-based dynamic choices parameter encounters an error.

hazmat345:
Fixes #1085. Fixes the error indicator when a command-based choices request fails.

Depends on https://github.com/beer-garden/angular-schema-form-addons/pull/30.

## Test instructions
Run the `dynamic` plugin, and open the UI page to execute the `say_specific_from_command_fully_specified` command. This command's choices parameter is targeting the `default` instance, which no longer exists. So there are no choices available for the `message` parameter but also no indication that anything has gone wrong. If you check the browser console you can see an error, which is another indication that something isn't working correctly.

Using this branch and the linked PR, go to the page again. There should now be a warning indicator next to the `message` parameter title.

### DIFF
--- a/src/ui/src/js/controllers/command_view.js
+++ b/src/ui/src/js/controllers/command_view.js
@@ -73,11 +73,7 @@ export default function commandViewController(
       };
     }
 
-    return RequestService.createRequestWait(request).then(
-      function(response) {
-        return response.data;
-      }
-    );
+    return RequestService.createRequestWait(request);
   };
 
   $scope.checkInstance = function() {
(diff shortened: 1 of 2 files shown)


scott-taubman (approved):
LGTM and tested okay.  Do we need to update the package.json for the new version of the form addons that this requires?

hazmat345:
@scott-taubman yes, we'll need to bump the versions when we release.

PR #1088, opened 2021-08-07
Waiting for a request to complete will now time out

### LINKED ISSUE #1087: Command-based choices frontend timeout
Command-based choices has a bad failure condition if the instance the choices request is targeting isn't running. The request will be in a CREATED state forever, and while the request is in that state the frontend will query that status every half a second. Forever.

Putting a hard limit on the amount of time we'll wait for the request to complete is probably the best way to go here, and I'm thinking 30 seconds. It would be better to have the timeout be a configurable parameter on the choices definition itself, but I think that would be more trouble than it's worth - a command-based choice source that takes more than 30s to execute is going to require a user to sit on the command page for that long before they can do anything. I don't really think that's something we should be worried about supporting.

I think this might have been the cause of #1047.

hazmat345:
Fixes #1087 - command-based choices will time out if the request takes longer than 30 seconds to complete.

## Test Instructions

1. Modify the `dynamic` plugin `get_choices` function to sleep for 31 seconds
2. Navigate to the command page for `say_specific_from_command` in a browser
3. Wait 30 seconds. An indication should appear with a tooltip indicating a timeout expired.

### DIFF
--- a/src/ui/src/js/controllers/command_view.js
+++ b/src/ui/src/js/controllers/command_view.js
@@ -73,7 +73,7 @@ export default function commandViewController(
       };
     }
 
-    return RequestService.createRequestWait(request);
+    return RequestService.createRequest(request, true);
   };
 
   $scope.checkInstance = function() {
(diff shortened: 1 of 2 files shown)


Jake-Ross:
Works well for me, although I couldn't tell the first few times I tried since the indication circle is so small.

Jake-Ross (approved):

PR #1090, opened 2021-08-09
Issue/1067

### LINKED ISSUE #1067: Add api endpoints for importing and exporting scheduled jobs
We need two new api endpoints:

* /api/v1/export/jobs
* /api/v1/import/jobs

The export endpoint takes a body like:
```json
{
   "ids": ["string"]
}
```
An empty body would indicate that everything should be exported.  The result of the api call should be a json that is in the format that the import endpoint would use to load the jobs back in (a list of serialized jobs).

For the import endpoint, it accepts json as the input in the format output by the exporter (again, a list of serialized jobs) and will load all of those jobs in.

For the initial version of this, if it's easy enough to do, let's reject the entire import set if there are problems with any of them, rather than doing a partial import.  That may or may not prove to be feasible.

jlrpnbbngtn:
Closes #1067 

## Test Instructions
1. There are integration tests that exercise the added functionality. Check the code for the tests at `test/integration/scheduler/jobs/job_import_export_test.py` and see the successful run of those tests [here](https://github.com/beer-garden/beer-garden/actions/runs/1113384087).
2. Alternately, the functionality can be tested through `curl` with the following:

* If none exist on the system, create a scheduled job through the BG UI in the normal way. To find the IDs of the jobs on the system, the easiest way seems to be to use Swagger's `/api/v1/jobs` section, which will list all the jobs. Note the ID of a least one job.
* From the command line, type

    ```bash
    curl -X 'POST'  'http://localhost:8080/api/v1/export/jobs' \
    -H 'accept: application/json' -H 'Content-Type: application/json' \
    -d '{
    "ids": [
    <the job ID you noted including the double quotes>
     ]
    }'
    ```

* Verify the returned JSON is accurate for the job exported. Save the returned JSON in a convenient location for the next test.
* Type at the command line:

    ```bash
    curl -X 'POST' 'http://localhost:8080/api/v1/import/jobs' \
    -H 'accept: application/json' \
    -H 'Content-Type: application/json' \
    -d'<the body of the JSON returned from the last step including the enclosing braces>'
    ```

* Verify that the job has been created through the "Requests" page of the BG UI and that it is a copy of the exported job. Note that the "Next Run Time" field will be blank until the scheduler picks up the new job.

NB: The job export functionality can also be checked using the Swagger UI. However, due to what appears to be a bug in this version of Swagger, the JSON returned in the web page is wrong. Swagger will not accept the date-time objects unless they are enclosed in quotes, but BG will then reject them.


jlrpnbbngtn:
Note that 2 of the github actions are failing because this branch depends on updates to brewtils that have not been merged yet.

src/app/beer_garden/api/http/handlers/v1/job.py
```diff
@@ -210,11 +217,145 @@ async def post(self):
...
+        response = await self.client(
+            Operation(operation_type="JOB_CREATE_MULTI", args=[cleaned_job_dfn_list])
+        )
+
+        self.set_status(201)
+        self.set_header("Content-Type", "application/json; charset=UTF-8")
+        self.write(response)
+
+    @classmethod
+    def _clean_job(cls, job: brewtils.models.Job) -> brewtils.models.Job:
```
scott-taubman, line 289: I made a related comment on the brewtils side, but I think rather than doing this whole cleanup thing, you can just use a schema that is specific to the import and export process which doesn't include this fields.

src/app/beer_garden/api/http/handlers/v1/job.py
```diff
@@ -210,11 +217,145 @@ async def post(self):
...
+            $ref: '#/definitions/400Error'
+          50x:
+            $ref: '#/definitions/50xError'
+        tags:
+          - Jobs
+        """
+        filter_params_dict = {}
+
+        # we're focused on the "ids" list, but if arguments are passed, allow
+        # them through; that is, provide maximum flexibility to the user
```
scott-taubman, line 335: This sounds like we're providing a secret, undocumented API.  Meaning the schema definition tells you that you can use "ids", but we'll secretly take other arguments too.  I'm not sure that's a thing we want to do.

src/app/beer_garden/scheduler.py
```diff
@@ -510,10 +521,25 @@ def get_job(job_id: str) -> Job:
     return db.query_unique(Job, id=job_id)
 
 
-def get_jobs(filter_params: Dict = None) -> List[Job]:
+def get_jobs(filter_params: Optional[Dict] = None) -> List[Job]:
     return db.query(Job, filter_params=filter_params)
 
 
+def get_some_jobs(filter_params: Optional[Dict] = None) -> JobDefinitionList:
```
scott-taubman, line 528: I think once `JobDefitionList` goes away you lose the need for this method.  `get_jobs` although mapped to the "JOB_READ_ALL" operation actually allows you to get "some" jobs by providing a filter.

src/app/beer_garden/scheduler.py
```diff
@@ -530,6 +556,48 @@ def create_job(job: Job) -> Job:
     return job
 
 
+def create_jobs(jobs: JobDefinitionList) -> JobIDList:
```
scott-taubman, line 559: Probably obvious by this point, but this should just take a list of `Job` objects and return a list of string ids.

src/app/beer_garden/scheduler.py
```diff
@@ -530,6 +556,48 @@ def create_job(job: Job) -> Job:
...
+    the_connection: pymongo.MongoClient = mongoengine.get_connection()
+    the_db: pymongo.database.Database = mongoengine.get_db()
+
+    session: ClientSession
+    with the_connection.start_session() as session:
+        job_collection: Collection = the_db.get_collection(job_collection_name)
+
+        for job in jobs_list:
+            mongo_obj: mongoengine.Document = from_brewtils(job)  # noqa
+            job_as_son: bson.BSON = mongo_obj.to_mongo()
```
scott-taubman, line 585: was this intended to be job_as_**b**son?

src/app/beer_garden/api/http/handlers/v1/job.py
```diff
@@ -210,11 +217,145 @@ async def post(self):
...
             )
         )
 
         self.set_status(201)
         self.set_header("Content-Type", "application/json; charset=UTF-8")
         self.write(response)
+
+
+class JobImportAPI(BaseHandler):
+    def data_received(self, chunk: bytes) -> Optional[Awaitable[None]]:
```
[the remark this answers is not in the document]
jlrpnbbngtn: That was retained in error. The IDE erroneously marked that as being an abstract method and then erroneously produced  an implementation that proves that it's not abstract. smh...

src/app/beer_garden/scheduler.py
jlrpnbbngtn: Surprisingly, no it was not. That's how the library authors chose to name their functions.

src/app/beer_garden/api/http/handlers/v1/job.py
jlrpnbbngtn: Removed.

src/app/beer_garden/scheduler.py
jlrpnbbngtn: Done.

src/app/beer_garden/scheduler.py
jlrpnbbngtn: Changed.

src/app/beer_garden/api/http/handlers/v1/job.py
jlrpnbbngtn: Done.

src/app/beer_garden/api/http/handlers/v1/job.py
```diff
@@ -210,11 +217,145 @@ async def post(self):
...
+          50x:
+            $ref: '#/definitions/50xError'
+        tags:
+          - Jobs
+        """
+        empty_body_regex = re.compile(r"\s*\{\s*\}\s*")  # noqa
+        decoded_body: str = self.request.body.decode("utf-8")
+
+        if not len(decoded_body) or empty_body_regex.match(decoded_body):
+            raise ValueError("Body was empty")
```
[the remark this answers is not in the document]
jlrpnbbngtn: Check moved into base handler.

PR #1094, opened 2021-08-12
Issue/1007

### LINKED ISSUE #1007: Router failures
I'm seeing issues with routing after beergarden has been running for a while. Trying to use a garden with no local plugins, but a bunch of plugins on a downstream garden. Things work correctly initially, but after a while attempts to create a request that needs to be routed to the child garden will fail. It looks like the routing is determining that the request needs to be published to the broker (as if the plugin was on the local garden) instead of forwarded to the correct child garden.

hazmat345:
This fixes #1007 by correcting two issues that were causing the routing to become out of sync.

1. The routing module now ignores garden sync events from grandchild gardens.
2. Whenever the garden module sees a downstream sync event it publishes system update events. These events now include the correct downstream garden so the router module doesn't think they originated from the local garden.

## Test instructions
This is pretty nebulous as I think the problem this PR fixes is kind of a race condition. To demonstrate the issue use the develop branch and set up three gardens in the following configuration, each with plugins running locally:

parent -> child -> grandchild

The problem manifests as the parent thinking a system that exists on a downstream garden actually exists locally. So an attempt to create a request for a grandchild system **from the parent** will generate an error about how the request couldn't be published to the message queue.

The best way to make this happen is to restart the child and/or grandchild gardens some number of times without restarting the parent. Restarting the parent will "reset" the routing using the systems currently in the database, and this is always correct.

Using this branch it shouldn't be possible to make this behavior occur. Any restarts and syncs directed at any of the gardens shouldn't matter - a request originating from the parent garden should be routed correctly regardless of the destination.

### DIFF
--- a/src/app/beer_garden/garden.py
+++ b/src/app/beer_garden/garden.py
@@ -293,6 +293,7 @@ def handle_event(event):
                     publish(
                         Event(
                             name=Events.SYSTEM_UPDATED.name,
+                            garden=event.garden,
                             payload_type="System",
                             payload=system,
                         )
--- a/src/app/beer_garden/router.py
+++ b/src/app/beer_garden/router.py
@@ -416,16 +416,20 @@ def handle_event(event):
     elif event.name == Events.SYSTEM_REMOVED.name:
         remove_routing_system(system=event.payload)
 
-    # Handle downstream events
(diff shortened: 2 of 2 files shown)


scott-taubman (approved):
LGTM and tested okay.

PR #1099, opened 2021-08-13
Choices no longer use "default" instance by default

### LINKED ISSUE #1096: Choices shouldn't use "default" instance
The UI shouldn't attempt to create a choices request against the "default" instance on page load. In the case of multiple instances we should be waiting for one to be selected before attempting the request.

hazmat345:
Fixes #1096. Command-based choices won't populate on initial page load until an instance is selected (instead of just trying to use the "default" instance).

## Test Instructions
1. Use the dynamic plugin
2. Create instances `d1` and `default`
3. Stop the `default` instance

Now go to the `say_specific_from_command` page. Previously, the request will spin for the 30 seconds and then time out. Now there should be an info icon telling the user to select an instance.

Selecting the `d1` instance should cause the choices population to work normally.

Selecting the `default` instance will still cause the request to spin for 30 seconds and then time out. However, this is now the correct behavior - the user has intentionally tried to populate choices using an instance that isn't running.

### DIFF
--- a/src/ui/src/js/controllers/command_view.js
+++ b/src/ui/src/js/controllers/command_view.js
@@ -7,6 +7,7 @@ commandViewController.$inject = [
   '$state',
   '$stateParams',
   '$sce',
+  '$q',
   'RequestService',
   'SFBuilderService',
   'system',
@@ -20,6 +21,7 @@ commandViewController.$inject = [
  * @param  {Object} $state            Angular's $state object.
  * @param  {Object} $stateParams      Angular's $stateParams object.
  * @param  {Object} $sce              Angular's $sce object.
+ * @param  {Object} $q                Angular's $q object.
  * @param  {Object} RequestService    Beer-Garden's request service object.
  * @param  {Object} SFBuilderService  Beer-Garden's schema-form builder service object.
  */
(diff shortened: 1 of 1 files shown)


src/ui/src/js/controllers/command_view.js
```diff
@@ -60,10 +63,19 @@ export default function commandViewController(
...
+      instance_name: requestPrototype['instance_name'] || $scope.model['instance_name']
     };
 
+    // If a system has more than one instance this will be undefined on initial page
+    // load. We could just let this continue as normal, but the backend would return
+    // an error like "Could not find instance with name 'None' in system". That's not
+    // the best, so this special handling makes the error message a little nicer.
+    if (!request.instance_name) {
+      let deferred = $q.defer();
+      deferred.reject("Please select an instance");
```
Jake-Ross, line 75: Instead of this error, I'm getting an error in the small circle tooltip that says `Couldn't populate choices from Undefined`. Selecting a running instance does not remove this error, nor correctly populate choices with a new request.

scott-taubman: I'm guessing this related to the the changes in the form-addons repo seen [here](https://github.com/beer-garden/angular-schema-form-addons/pull/37/files#diff-bc6b134f359dc185bece7f778ad5d71d88b95764684f1c5d2522d42a26468207R329).  I tested this PR against the form-addons from that one and it seems to work as expected.

scott-taubman (approved):

PR #1103, opened 2021-08-16
Fix double URL encoding on UI

### LINKED ISSUE #1100: Garden creation url encoding issues
We recently added url encoding in various spots to ensure that the frontend could make correct calls to the backend to retrieve gardens with non-alphanumeric characters in their name.

Unfortunately we appear to have inadvertantly added the url encoding to the garden creation call, which results in the POST data being encoded.  This results in the actual garden name being encoded, i.e. "g%40rden" rather than "g@rden".

Somewhat related, I also notice that the URL that you get directed to after creating a garden ends up being double-encoded.  Meaning that if I create a garden named "g@rden", I should be redirected to:

* `/admin/gardens/g%40rden/`

but instead get directed to:

* `/admin/gardens/g%2540rden/`

The `%` of the encoded `%40` is getting encoded again.

Jake-Ross:
Closes #1100

This PR also removes all URL decoding from the UI - what the user sees as a garden name should be what's read directly from Beergarden's DB.

## To Test
- Create gardens in the Beergarden UI with special characters (`@ # $ % ^ & *`)
- Ensure that no garden names double-encode
- Ensure that the URL for the 'edit config' page of a garden is properly URLencoded

Note that the issue linked to this PR says:

>Somewhat related, I also notice that the URL that you get directed to after creating a garden ends up being double-encoded. Meaning that if I create a garden named "g@rden", I should be redirected to:
>`/admin/gardens/g%40rden/`

... although, this PR does not encode the `@` symbol. This doesn't seem to cause an issue with routing, however.

### DIFF
--- a/src/app/beer_garden/api/http/handlers/v1/garden.py
+++ b/src/app/beer_garden/api/http/handlers/v1/garden.py
@@ -4,6 +4,7 @@ from beer_garden.api.http.base_handler import BaseHandler
 from brewtils.errors import ModelValidationError
 from brewtils.models import Operation
 from brewtils.schema_parser import SchemaParser
+from requests.utils import unquote
 
 
 class GardenAPI(BaseHandler):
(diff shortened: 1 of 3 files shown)


scott-taubman (approved):
Tested and behaves as expected now.

PR #1107, opened 2021-08-20
#1074 - Lint integration tests during PRs, ignore fewer flake8 errors

### LINKED ISSUE #1074: Expand linting that gets run on pull requests
The set of tests that run against PRs does not include running the linting against the integration tests.  This has lead to the integration test code not meeting our general coding standards, despite the fact that it is python just like the application code.  We should fix this so that linting always runs against the integration test code.

Also, the flake8 config in `src/app/setup.cfg` ignores some pretty common errors that we actually want to catch.  We should remove those from the config, clean up the code accordingly so that the flake8 checks pass, and then use that config for the integration test linting in addition to the app code linting.

scott-taubman:
Closes #1074 

This does a few things:

* Adds linting of the integration tests to the PR github action workflow
* Removes a couple of errors that we had been ignoring during the flake8 check (line length being the important one).
* Updates the code in `src/app` to conform to the 88 char line length limit
* Updates the code in `test/integration` to pass the flake8 check
* Removes code that was labeled "old_v2_stuff" rather than updating it to pass the linting.

The majority of the changes are related to reducing line length down to < 88 characters, and not using an f-string in cases where there is no substitution required. This was not done by hand; black took care of it all for me. So between that and the fact that the tests still pass, you probably don't want to kill yourself scrutinizing every change.

# Test Instructions
* This one is really about the checks in this PR still passing.  If you look at the [details of Linting check](https://github.com/beer-garden/beer-garden/pull/1107/checks?check_run_id=3385045929#step:6:23) you'll notice that there's now a 2nd linting section that's specific to the integration tests, which is mostly the whole point of this PR.


jlrpnbbngtn (approved):
LGTM, no obvious deficiencies. Merging.

Issue #1108, opened 2021-08-23
SchemaForm coupling


hazmat345:
Currently we expose a capability that's strongly coupled to the UI AngularSchemaForm implementation in Brewtils. This was almost definitely a bad idea and should probably be deprecated to prepare for the new UI.

## Background
The "normal" way things work is this: brewtils uses info from various sources (mostly the `@command` and `@parameter` decorators) to construct a Command object. This is stored in the database and retrieved by the frontend UI. The UI then passes this Command definition to the angular-schema-form-builder, which constructs an ASF `schema` and an ASF `form`. ASF then uses the schema and form to render the UI widgets, do validation, and manage the models.

If you turn on `ui.debug_mode` in the Beergarden config, you can see the schema and form definitions on the "make a request" page.

## The problem
We currently allow you to specify an alternate schema and/or form for a Command. So if you understand the structure of what ASF is expecting, you can customize the "make a request" page in ways that aren't possible the "normal" way:

```python
@command(form=[{"key": "parameters.message", "readonly": True}])
@parameter(key="message", type="String", optional=False, nullable=False)
def echo_message_custom_form_as_list(self, message="Can't change me! Hahaha!"):
    return message
```

This results in a disabled control where the value cannot be changed.

## Discussion
Generally, it's nice to let power users customize the frontend if they're willing to take on the burden of figuring out how to make it happen. This gives them the tools they need to do what they want without the need to support every possible customization ASF supports inside of Beergarden.

However, this obviously ties us to the way that ASF builds and interprets these structures, which means that in order for things to continue to work with a new UI we would need to translate ASF-specific things into whatever way the new UI expects things to work. This isn't really feasible as the amount of effort required wouldn't be worth it.

I think the best way forward is to deprecate the `schema` and `form` parameter attributes. They can then be removed when the new UI is ready, removing the need to worry about supporting them past their useful life.

scott-taubman:
@scott-taubman We should notify people that this will be deprecated at the same time we notify about python 2 support coming to an end.

PR #1109, opened 2021-08-23
#1104 - Explicitly write default Garden configs upon Garden creation

### LINKED ISSUE #1104: Make default garden configuration more explicit
Currently it is the case that garden configuration is loaded from a merging of settings that you have explicitly configured and the default values as defined in the config schema.  This behavior is and has the potential to be the source of great confusion.  Two examples:

1) The UI reads only the explicitly defined configuration on a garden when loading the garden configuration page.  This means that if, for instance, you have set hostname and port on the garden, but not defined "ssl_enabled" at all, the UI will render the "ssl_enabled" checkbox as unchecked, when in reality the default value for that (which is the value that will get used in the absence of explicit configuration) is True.

2) If a user had not defined a setting on their garden because the default was acceptable, and then for some reason we decide to change the default value, the user's garden configuration would then change through no fault or desire of their own.  We'll again use "ssl_enabled" as the example.  Let's say the default value were False and users had their gardens configured and working using that default.  If we then change the default to True to encourage better practices on the part of our users, all existing garden configurations that were relying on that default would break.  This seems likely to confuse and annoy people.

One approach that might solve both ends of this confusion is to write down the default connection_param configuration at garden creation time.  That is, load the default configs and then save them as explicit configuration for the garden.  From that point forward, the defaults would not need to get merged with the explicitly set values because all of the values would be explicitly set (to those defaults).  It would be nice, though not required, if this was the behavior any time configuration was saved, such that if new configuration was added the defaults for that new configuration would be recorded in the garden config the next time it got saved.

Worth noting is that the configuration that was causing the problem described in 1) above is actually on the brewtils side of things (shown [here](https://github.com/beer-garden/brewtils/blob/develop/brewtils/specification.py#L60)).  I'm not sure that the source is important, except that it probably makes it less of a good idea to just change that default value to False to clear up the immediate source of confusion.

Jake-Ross:
Closes #1104 

The idea behind this change is to use Yapconf to specify default connection parameters for each newly created garden, then overwrite these defaults with the user-specified config.

Worth noting is the difference in specification between `_ENTRY_SPEC` in `config.py` and the form created for the UI in `garden_service.js`. As far as I understand, these should match each other, so I changed the UI end to reflect how `_ENTRY_SPEC` looks. 

## To Test
- Create a new garden in the UI
- Ensure Garden edit page is populated with default values (defined in `config.py`)
- Create a new garden using the API, specifying at least one config option under the garden's `connection_params` (e.g. specify a non-default port).
- Ensure the new garden reflects your changes in the Garden edit page.

### DIFF
--- a/src/app/beer_garden/garden.py
+++ b/src/app/beer_garden/garden.py
@@ -15,6 +15,7 @@ from typing import List
 
 from brewtils.errors import PluginError
 from brewtils.models import Event, Events, Garden, Operation, System
+from brewtils.specification import _CONNECTION_SPEC
 from mongoengine import DoesNotExist
 
 import beer_garden.config as config
@@ -22,6 +23,8 @@ import beer_garden.db.api as db
 from beer_garden.events import publish, publish_event
 from beer_garden.namespace import get_namespaces
 from beer_garden.systems import get_systems, remove_system
+
+from yapconf import YapconfSpec
 
 logger = logging.getLogger(__name__)
 
@@ -177,6 +180,34 @@ def create_garden(garden: Garden) -> Garden:
         The created Garden
 
     """
+    # Explicitly load default config options into garden params
+    if garden.connection_params is None:
+        garden.connection_params = {}
+
+    connection_params = getattr(garden, "connection_params", {})
+    # bg_host is required by brewtils garden spec
+    connection_params.setdefault('bg_host', '')
+
+    spec = YapconfSpec(_CONNECTION_SPEC)
+    connection_params = spec.load_config(connection_params)
+
+    key_dict = {key: key for key, _ in connection_params.items()}
+    config_map = {
+        "bg_host": "host",
+        "bg_port": "port",
+        "ssl_enabled": "ssl",
+        "bg_url_prefix": "url_prefix",
+        "ca_cert": "ca_cert",
+        "ca_verify": "ca_verify",
+        "client_cert": "client_cert",
+    }
+    key_dict.update(config_map)
+    garden.connection_params['http'] = {
+        key_dict[key]: value for key, value in connection_params.items()
+    }
+    for key in config_map:
+        garden.connection_params.pop(key, None)
+
     garden.status_info["heartbeat"] = datetime.utcnow()
 
     return db.create(garden)
--- a/src/app/beer_garden/garden.py
+++ b/src/app/beer_garden/garden.py
@@ -197,10 +197,10 @@ def create_garden(garden: Garden) -> Garden:
 
     if garden.connection_params is None:
         garden.connection_params = {}
-    garden.connection_params.setdefault('http', {})
+    garden.connection_params.setdefault("http", {})
 
     for key in config_map:
-        garden.connection_params['http'].setdefault(config_map[key], defaults[key])
+        garden.connection_params["http"].setdefault(config_map[key], defaults[key])
 
     garden.status_info["heartbeat"] = datetime.utcnow()
 
--- a/src/app/beer_garden/garden.py
+++ b/src/app/beer_garden/garden.py
@@ -186,7 +186,7 @@ def create_garden(garden: Garden) -> Garden:
 
     connection_params = getattr(garden, "connection_params", {})
     # bg_host is required by brewtils garden spec
-    connection_params.setdefault('bg_host', '')
+    connection_params.setdefault("bg_host", "")
 
     spec = YapconfSpec(_CONNECTION_SPEC)
     connection_params = spec.load_config(connection_params)
@@ -202,7 +202,7 @@ def create_garden(garden: Garden) -> Garden:
         "client_cert": "client_cert",
     }
     key_dict.update(config_map)
-    garden.connection_params['http'] = {
+    garden.connection_params["http"] = {
         key_dict[key]: value for key, value in connection_params.items()
     }
     for key in config_map:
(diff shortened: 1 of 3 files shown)


Jake-Ross:
To be more specific about the issue with specifications:

[Here](https://github.com/beer-garden/beer-garden/blob/883dce0365991b9bcadb06ad2e5438ea81c7245c/src/app/beer_garden/config.py#L720-L899) is the location of the specification that the Beergarden backend has for what a garden's `connection_params` should look like. This is the form I used to populate default values for new Gardens. 

[Here](https://github.com/beer-garden/beer-garden/blob/883dce0365991b9bcadb06ad2e5438ea81c7245c/src/ui/src/js/services/garden_service.js#L90-L280) is the specification for the form on the UI side that dictates what a user can specify for a Garden. 

These seem independent of each other, meaning that if changes are made on the backend side, changes will have to be manually reflected here. I believe it uses the same AngularSchemaForm system described in #1108, which we are trying to deprecate.

src/app/beer_garden/garden.py
```diff
@@ -167,6 +169,11 @@ def create_garden(garden: Garden) -> Garden:
         The created Garden
 
     """
+    spec = YapconfSpec(config._SPECIFICATION)
+    garden.connection_params = spec.load_config({"entry": garden.connection_params})[
```
scott-taubman, line 173: It's ultimately no different, but you can actually load the `_ENTRY_SPEC` directly to avoid the slightly awkward way this reads with the `["entry"]` bit.

```python
spec = YapconfSpec({"entry": config._ENTRY_SPEC})
garden.connection_params = spec.load_config(garden.connection_params)
```

I don't love that we're referencing the specification constants here, as the leading `_` in the name implies that those are intended to be for internal use in config.py only.  I think it's worth having a conversation about this, but given what you're trying to accomplish I think what you are doing is probably the most correct option available to you.

src/app/example_configs/config.yaml
```diff
@@ -22,12 +22,12 @@ entry:
     host: 0.0.0.0
     port: 2337
     ssl:
-      ca_cert: null
-      ca_path: null
+      ca_cert: ''
```
scott-taubman, line 25: I'm not sure these need to change either.

scott-taubman (commented):
I made some mostly minor comments.  In addition to those, we should add a unit test that confirms that the `create_garden` function does indeed merge the defined settings with the default settings.  Ideally that test would mock the default settings so that the test isn't dependent on what those defaults are at any given time.

src/app/beer_garden/garden.py
Jake-Ross: I actually tried using `_ENTRY_SPEC`, but Yapconf has a strange requirement that dict that you pass in has a dict for all its key-value pairs. It definitely makes sense to just use `_ENTRY_SPEC`, but Yapconf doesn't allow it, so some bypassing has to be done with this.

Also, I understand the concern with `_`. The only place I could find a reliable source for Garden config defaults was here, though. There may be something that presents this spec info publicly, though - I will BOL for that.

src/app/example_configs/config.yaml
Jake-Ross: I didn't change these manually - these actually change when defaults in `_SPECIFICATION` are updated, which is a neat function.

Jake-Ross:
Recent pushes address all previous comments, as well as some offline conversations. Some notes:

- `bg_host` is a required parameter (which transforms into `http.host`) from the brewtils specification, meaning it has to be supplied a default value upon Garden creation. Right now, the default is an empty string.
- The problem with misaligned UI and actual config values persists, as even if defaults are sent to the UI upon garden creation, the UI doesn't 'load' these values.

Jake-Ross:
> The problem with misaligned UI and actual config values persists, as even if defaults are sent to the UI upon garden creation, the UI doesn't 'load' these values. This can likely be dealt with within this PR.

As it turns out, this problem only persisted because of a JS syntax error - should be fixed now, and PR seems ready for review.

src/app/beer_garden/garden.py
```diff
@@ -177,6 +180,34 @@ def create_garden(garden: Garden) -> Garden:
...
+    # Explicitly load default config options into garden params
+    if garden.connection_params is None:
+        garden.connection_params = {}
+
+    connection_params = getattr(garden, "connection_params", {})
+    # bg_host is required by brewtils garden spec
+    connection_params.setdefault("bg_host", "")
+
+    spec = YapconfSpec(_CONNECTION_SPEC)
+    connection_params = spec.load_config(connection_params)
```
scott-taubman, line 192: I ran into a problem during testing that the `connection_params` I was supplying weren't getting used at all. One problem I think you might have here is that because you are calling `load_config` before you've done the config_map ceonversion stuff, the input `connection_params` have to use the `_CONNECTION_SPEC` config names in order to get used. That's not the intended behavior.  The expected input should be "host", "port", etc. rather than "bg_host", "bg_port", and the like.  Otherwise we're effectively changing the API, which isn't our intention.

scott-taubman (changes requested):
Made a few comments, mostly related to what you're expecting the input to look like.  Let me know if you have questions or if I misinterpreted anything.

src/app/beer_garden/garden.py
Jake-Ross: You could tell from the unit tests that, for a reason that I can't figure out now, I was expecting `_CONNECTION_SPEC` to have the args that `create_garden` was expecting. Looking at a test POST request confirmed I had it backwards. I've pushed changes to fix this.

scott-taubman (approved):

PR #1111, opened 2021-08-25
#1110 - Return 404 from Garden API in appropriate situations

### LINKED ISSUE #1110: Garden REST endpoints should return proper response codes
It is currently the case that:

* GET /api/v1/gardens/{garden_name} for a garden that does not exist returns a 200 with "null" as the body rather than a 404.
* DELETE /api/v1/gardens/{garden_name} for a garden that does not exist throws a 500, again instead of a 404.

This is a problem for more than just the Garden endpoints, but in service of [brewtils/#348](https://github.com/beer-garden/brewtils/issues/348) we should fix the problem at least for the Garden ones.

scott-taubman:
Closes #1110 

# Test Instructions
* Unfortunately writing unit tests to verify the API responses isn't viable at the moment.
* To manually test, try doing a GET and a DELETE against /api/v1/gardens/notarealgarden.  You should get a 404 in each case (as opposed to a 200 and a 500 respectively).

### DIFF
--- a/src/app/beer_garden/garden.py
+++ b/src/app/beer_garden/garden.py
@@ -14,11 +14,11 @@ from datetime import datetime
 from typing import List
 
 from brewtils.errors import PluginError
-from brewtils.models import Events, Garden, Operation, System, Event
+from brewtils.models import Event, Events, Garden, Operation, System
 
 import beer_garden.config as config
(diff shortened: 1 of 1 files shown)


Jake-Ross (approved):
Simple fix; works as intended.

Issue #1079, opened 2021-07-28
More explicitly surface garden information on models


scott-taubman:
Currently the distinction between namespace and garden isn't as clear as it could be.  Also, the garden information isn't really explicit and surfaced in the places that it would be really helpful to have.  We should update the models, ui, etc to properly expose the garden that systems belong to wherever we can.

scott-taubman:
If we want to use an actual ReferenceField for this (which seems preferable, but its unclear if that's actually wise or not with the current model setup), we need to tackle #1092 first.

scott-taubman:
@jlrpnbbngtn So after starting to go down this path a little bit, I'm starting to lean toward it not being worth adding an actual "garden_name" field to things.  There is a lot of stuff throughout the code that very much treats namespace as if it is the garden name and I'm not entirely convinced that using the namespace in any other way is a thing that would work as desired or really make sense.  Basically, I'm suggesting that it's less that "the distinction between namespace and garden isn't as clear as it could be" and more that there is no distinction.  Namespace should just be treated as the garden name.

For this issue then, I'm thinking that I might just add namespace to a couple of places where it is missing and then have the authorization stuff proceed under the assumption that namespace is the garden name.  Then one day (probably the mythical 4.0) bite the bullet and rename the field.

Thoughts?

scott-taubman:
I had a conversation with some people internally and it seems that everyone has just accepted that, regardless of what the original desire / intent was, namespace and garden name have become too conflated to separate at this point.  We're going to proceed treating namespace as the garden name, which may make this ticket OBE. Job still needs namespace info, but that should always be accessible through the RequestTemplate.  If that doesn't work out, we can come back to this and add it.  If it proves unnecessary, we'll close this.

scott-taubman:
Since deciding to treat namespace as the garden name, the need for this hasn't surfaced.  I'm going to close it for now.

PR #1114, opened 2021-08-26
#1092 - Store local garden info in the database

### LINKED ISSUE #727: Can Create Two Gardens with same name through the API

### LINKED ISSUE #1092: Eliminate the distinction between local and remote gardens
We currently treat "local" and "remote" gardens differently, such that it becomes necessary to do things like:

```python
def get_gardens(include_local: bool = True) -> List[Garden]:
    gardens = db.query(Garden)

    if include_local:
        gardens += [local_garden()]

    return gardens
```

This special handling of the local garden becomes especially problematic when we want to add authorization checks. We don't want to have two separate checks with potentially different logic sprinkled through the system.

Instead we should just figure out how to have the local garden info live in the DB alongside the remote gardens such that we can just get all of them with a single query.  It may be be necessary to keep local_garden() around for compatibility purposes and just change it to return the local garden info from the DB, but if we can eliminate local_garden() calls altogether that would be a nice bonus.

scott-taubman:
Closes #1092
Closes #727

This takes the garden name as defined in the config file and creates a Garden entry for it in the database.  It then updates the appropriate places to pull said entry from the db when appropriate.

Things still exist in a slight in-between land in this PR, meaning there are still some distinctions between local and remote gardens that have to be accounted for.  Specifically, local gardens don't get the list of embedded System documents that remote gardens have, and instead that data is assembled in real-time when requested.  This is not ideal, but it's the easiest way to maintain compatibility as we make the incremental model changes necessary for all of the auth stuff.  There are comments in the code that explain why certain things had to hang around.

I wrote a bunch of unit tests, since there were none and I wanted some sort of sanity check that my changes weren't going to break things.  I'm open to writing more if there are any obvious gaps.

# Test Instructions
* The new unit tests I wrote ought to be passing. They should be scrutinized though for usefulness and relative completeness.
* The rest of the testing is really just firing up beer garden and making sure that your usual interactions with the local garden work as you'd expect.  The goal here is that there should be no discernible difference.

### DIFF
--- a/src/app/beer_garden/garden.py
+++ b/src/app/beer_garden/garden.py
@@ -52,11 +52,12 @@ def get_gardens(include_local: bool = True) -> List[Garden]:
         All known gardens
 
     """
-    filter_params = {}
-    if not include_local:
-        filter_params["connection_type__ne"] = "LOCAL"
+    # This is necessary for as long as local_garden is still needed. See the notes
+    # there for more detail.
+    gardens = db.query(Garden, filter_params={"connection_type__ne": "LOCAL"})
 
-    gardens = db.query(Garden, filter_params=filter_params)
+    if include_local:
+        gardens += [local_garden()]
 
     return gardens
 
@@ -70,6 +71,12 @@ def local_garden(all_systems: bool = False) -> Garden:
     Returns:
         The local Garden
     """
+    # This function is still necessary because there are various things that expect
+    # the system information to be embedded in the garden document itself (as opposed
+    # Systems just having a reference to their garden). There is nothing that would
+    # keep a LOCAL garden's embedded list of systems up to date currently, so we instead
+    # build the list of systems (and namespaces) at call time. Once the System
+    # relationship has been refactored, the need for this function should go away.
     garden: Garden = db.query_unique(Garden, connection_type="LOCAL")
 
     filter_params = {}
--- a/src/app/beer_garden/garden.py
+++ b/src/app/beer_garden/garden.py
@@ -52,10 +52,11 @@ def get_gardens(include_local: bool = True) -> List[Garden]:
         All known gardens
 
     """
-    gardens = db.query(Garden)
+    filter_params = {}
+    if not include_local:
+        filter_params["connection_type__ne"] = "LOCAL"
 
-    if include_local:
-        gardens += [local_garden()]
+    gardens = db.query(Garden, filter_params=filter_params)
 
     return gardens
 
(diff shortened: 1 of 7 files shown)


scott-taubman:
I rebased onto develop since the latest merge there was going to cause conflicts here.

src/app/beer_garden/garden.py
```diff
@@ -145,15 +144,18 @@ def remove_garden(garden_name: str) -> None:
...
-        None
-
+        The deleted garden
     """
-    garden = db.query_unique(Garden, name=garden_name, raise_missing=True)
+    garden = get_garden(garden_name)
 
-    for system in garden.systems:
+    # TODO: Switch to lookup by garden_name rather than namespace
+    systems = get_systems(filter_params={"namespace": garden_name})
```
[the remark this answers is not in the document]
scott-taubman: Per comments on #1079, namespace will likely just be treated as the garden name for the foreseeable future.

src/app/beer_garden/garden.py
```diff
@@ -52,7 +52,9 @@ def get_gardens(include_local: bool = True) -> List[Garden]:
         All known gardens
 
     """
-    gardens = db.query(Garden)
+    # This is necessary for as long as local_garden is still needed. See the notes
+    # there for more detail.
+    gardens = db.query(Garden, filter_params={"connection_type__ne": "LOCAL"})
```
Jake-Ross, line 57: I'm unsure why you're filtering for `"connection_type__ne"` here. Is this mongo magic?

scott-taubman: It's [mongoengine filter syntax](https://docs.mongoengine.org/guide/querying.html#query-operators).  The `db.query` call ultimate ends up doing:

```python
query_set = query_set.filter(**(kwargs.get("filter_params", {}))) 
```

Which in this case is essentially:
```python
queryset = Garden.objects.filter(connection_type__ne="LOCAL")
```

Jake-Ross: Makes sense to me. My Google-Fu was not strong enough to find that syntax.

Jake-Ross (approved):
Unit tests pass and look sane. I tested the local garden within the UI by running some requests and taking a look at the admin page to ensure nothing changed.

PR #1117, opened 2021-08-30
#1073 - Add a script for running integration tests locally

### LINKED ISSUE #1073: Local runs of integration tests
We need the ability to run BEERGARDEN integration tests on our workstations on a local branch.

#### Decisions:
* Is it better to run on only a single or a couple of versions of Python and dependencies? Then the tests would run quickly and wouldn't bog down the workflow process much. In that case, the tests would serve as a gut check that we haven't done anything crazy and broken everything.
* Is it better to run the full matrix of versions? In that case, we'd have strong evidence of the outcome of the tests when they're fired by GitHub actions at midnight. But it can take a long time to run.

#### Potential solutions:
1. Create a python tool that exactly mirrors the steps taken by the actions. Using the `docker` library for python should make many of the steps somewhat easy, though there are some steps in the actions that would take some work.
2. Create a python or bash script that mirrors the intent, but not the exact steps, taken by actions. Both of these options would require maintenance as the actions mature and the versions get bumped.
3. Wait for `act` to become usable without jumping through hoops. This has the benefit that the tests would live in only one place. The downside is that there's no knowing how long this will take (could be years, or days...).

jlrpnbbngtn:
Closes #1073 

There are minor adjustments to the integration tests [and they pass on GitHub](https://github.com/beer-garden/beer-garden/actions/runs/1172127398).

## Test Instructions
* Pull the branch, *cd* into *top level*`/test/local_integration` and view the `README.md` file there.


scott-taubman:
Some issues encountered during testing:
* `--test` seems to have to go at the end, while the usage instructions put it in the middle:
  ```
  ./integration_tests.py --test gardens_http /home/sataubm/dev/beer-garden/ /home/sataubm/dev/brewtils/
  ...
  integration_tests: error: argument --test: invalid choice: '/home/sataubm/dev/beer-garden/' 
  ```
* I had to teardown all of my existing containers to get the tests to run because of port conflicts.  I would suggest 1) only exposing ports that actually need to be externally accessible for the tests and 2) exposing those that must be exposed to non-standard ports so that they are less likely to conflict.
* When I did get the tests to run, the volume teardown process removed my existing volumes as well, not just those created by the test process.
* Doing `pip install -r requirements.txt -r requirements-dev.txt` wasn't enough to get the tests running.  I had to manually install brewtils as well.

Observations for future refinement:
* Much of the total test time is due to how long it takes for the beer garden instances to be actually up and running.  We could presumably save a lot of time if we could make the tests work without having to tear down and rebuild the containers between each section of the tests.

jlrpnbbngtn:
> #### Some issues encountered during testing:
> * `--test` seems to have to go at the end, while the usage instructions put it in the middle:
>   ```
>   ./integration_tests.py --test gardens_http /home/sataubm/dev/beer-garden/ /home/sataubm/dev/brewtils/
>   ...
>   integration_tests: error: argument --test: invalid choice: '/home/sataubm/dev/beer-garden/' 
>   ```

Yes, I see that now. I'll look into it. Surely `argparse` has a simple fix for this. Evidently, creating the args in the order you want is not it, though.

> * I had to teardown all of my existing containers to get the tests to run because of port conflicts.  I would suggest:
>   
>   * only exposing ports that actually need to be externally accessible for the tests and
>   * exposing those that must be exposed to non-standard ports so that they are less likely to conflict.

The second point is what I'll look into first. I believe that's an easy change. I do not know why rabbitmq, for example, needs access to more than one port. Perhaps it doesn't for our purposes, but I can't say at this time.

> * When I did get the tests to run, the volume teardown process removed my existing volumes as well, not just those created by the test process.

Yes it prunes any volume that's not in use. I don't see an easy workaround for this, but I'll think on it.

> * Doing `pip install -r requirements.txt -r requirements-dev.txt` wasn't enough to get the tests running.  I had to manually install brewtils as well.
> 

This is definitely something I hadn't thought about. The fix isn't just to install brewtils, it's to install *the brewtils that's under test*. In other words, if instead you'd already had brewtils installed but it was a different version/branch than the one your're putting into the container (via the command line argument), that could be problematic.

  I don't know if it's better to just document the gotcha in the README or if it's better to come up with a mechanical solution. Suggestions welcome.

Incidentally, the `requirements.txt` file in the top-level directory is the requirements for brewtils *inside the container*. I originally had it in the config directory and can't remember why I moved it. I think I'll put it back and rename it to avoid confusion.

> #### Observations for future refinement:
> * Much of the total test time is due to how long it takes for the beer garden instances to be actually up and running.  We could presumably save a lot of time if we could make the tests work without having to tear down and rebuild the containers between each section of the tests.

Absolutely and the choice to tear down and then bring back up between every test suite wasn't made lightly. As I've mentioned in face-to-face, there's sequencing involved in the tests. That is, a test will change the state and another test will depend on that change. This is bad. I see the solution as making the tests good which will in turn make simplifying and streamlining this tool possible.

scott-taubman:
Some follow-up thoughts related to how we might resolve some of those issues I ran into:

* Is there any reason to use docker volumes at all?  It doesn't seem like there's anything that needs to be persisted between each section of the tests (otherwise running a section in isolation (e.g. --test gardens_http) presumably wouldn't work.  You might be able to get around the volume pruning problem by just not using volumes.

* Similarly, you could probably get away without exposing any ports and having everything run inside the container network.  I know you looked at having the tests run inside a container originally and gave up on that idea, but based on where you ended up it seems like it should be doable.  Where you call pytest.main I would think you could instead call the equivalent of a docker run that spins up a container to run the tests, with that container being on the same container network as everything in your compose file.  Were you running into specific problems that made you think you were going to need to go down the docker-in-docker rabbit hole?

jlrpnbbngtn:
> Some follow-up thoughts related to how we might resolve some of those issues I ran into:
> 
> * Is there any reason to use docker volumes at all? ...

That's not a question I had considered before. I'd have to look into some things in order to answer it. But it's a good question and if the answer in *no*, that would make some things much simpler.

> * Similarly, you could probably get away without exposing any ports and having everything run inside the container network.  I know you looked at having the tests run inside a container originally and gave up on that idea, but based on where you ended up it seems like it should be doable.  Where you call pytest.main I would think you could instead call the equivalent of a docker run that spins up a container to run the tests, with that container being on the same container network as everything in your compose file.  Were you running into specific problems that made you think you were going to need to go down the docker-in-docker rabbit hole?

This seems like the correct next step in the evolution of this tool. Several design choices were  made mostly in an effort to get something up and running quickly. The idea is that the tool would address the state of the integration tests as they stand now, with the understanding that the tests would be refactored and expanded in the near future. So no sense in making something that's highly polished for the current environment if the environment is going away. In other words, shortcuts were deliberately taken.

But you're right that the current status suggests that just moving everything into the docker containers should be a low-overhead improvement, so I'm totally willing to spend a little time on that *before* tackling the integration tests as a whole.

scott-taubman:
I think what you have here is a fine starting point to build on as we try to improve the overall structure of the tests and the process of running them.  I would have @Jake-Ross go through the process of trying to run the tests as well to see if he runs into anything that I didn't that might need immediate attention.  If he is able to successfully get the tests running I would say just merge this when you're ready and continue to iterate on it.

jlrpnbbngtn:
> Some follow-up thoughts related to how we might resolve some of those issues I ran into:
> 
> * Is there any reason to use docker volumes at all?  ...

I hadn't actually considered this before. I'd have to look into some things before answering. But it's a good question and if the answer is *no* then some things would be much easier/simpler.

> * Similarly, you could probably get away without exposing any ports and having everything run inside the container network.  I know you looked at having the tests run inside a container originally and gave up on that idea, but based on where you ended up it seems like it should be doable.  Where you call pytest.main I would think you could instead call the equivalent of a docker run that spins up a container to run the tests, with that container being on the same container network as everything in your compose file.  Were you running into specific problems that made you think you were going to need to go down the docker-in-docker rabbit hole?

Github lost my last response to this and I don't want to try to recreate it. But, in short, yes this is a good next step and I think it would make sense to spend a little time on this *before* addressing the myriad issues with the integration tests as a whole.

Jake-Ross (commented):
I'm unsure of what is causing this, but attempting to run these tests times out for every test:

`urllib3.exceptions.TimeoutError: Timed out waiting to connect to beer-garden.`

I checked out docker for a post-mortem and found that the Beergarden containers that are stood up close with an error code a few seconds later. I'm still working on how to get more information as to why these close. 

```log
CONTAINER ID                                                       IMAGE                                         COMMAND                                                                                                                                            CREATED          STATUS                      PORTS                        NAMES
bfabcf8741f6699800d96765e798a0cb2df17d69c264e507be6d095e49884b0d   beer_garden_integration_tests                 "/bin/sh -c 'pip install -e $_BREWTILS_HOME && python $_BEER_GARDEN_APP_HOME/bin/app.py -c $CONF_DIR/config.yaml -l $CONF_DIR/app-logging.yaml'"   27 seconds ago   Exited (1) 22 seconds ago                                local_integration_beer-garden-child_1
44da81c7f6ed917ac5d8fb8a87d1126d8e814eefc8b94d1bf3ee455030dbf4b5   beer_garden_integration_tests                 "/bin/sh -c 'pip install -e $_BREWTILS_HOME && python $_BEER_GARDEN_APP_HOME/bin/app.py -c $CONF_DIR/config.yaml -l $CONF_DIR/app-logging.yaml'"   28 seconds ago   Exited (1) 23 seconds ago                                local_integration_beer-garden_1
```

Jake-Ross:
> I'm still working on how to get more information as to why these close.

Turns out that was pretty easy - the Beergarden container fails to build due to a change in some of the auth stuff that was merged to develop. After stepping back a few commits, the Beergarden container stayed alive and the tests ran fine.

To be specific, the container was killed by this:
```
  File "/beer_garden/src/app/beer_garden/db/mongo/models.py", line 554, in Role
    brewtils_model = brewtils.models.Role
AttributeError: module 'brewtils.models' has no attribute 'Role'
```
Changing this to LegacyRole to conform with the new auth commits does allow the container to stay alive, but there are still a wall of connection errors that I can't find the source of.

jlrpnbbngtn:
> I'm unsure of what is causing this, but attempting to run these tests times out for every test:

@Jake-Ross When you have a moment, would you mind putting a brief overview here of how you eventually overcame this problem?

scott-taubman:
> It's also why I was fine with just leaving this on its own branch without merging into the main branch, because it works just fine (or less horribly) with the setup just mentioned. FYI.

The problem I see with doing this, even in the short term, is that I don't think it lets you run any new or changed integration tests from other branches.  My understanding is that your script runs the integration tests from the parent directory of where it lives, meaning the tests that are in the branch the script lives in.  So while having it checked out on the side will let us run the existing integration tests against any new code we're working on in other branches, it doesn't really facilitate changes to the integration tests themselves, which to me is a large part of the value that we're after.

I really don't see the harm in merging this thing and continuing to iterate on it.

jlrpnbbngtn:
> I really don't see the harm in merging this thing and continuing to iterate on it.

We're in agreement on this. Comment was to explain my previous thinking only.

PR #1121, opened 2021-09-10
#1091 - Added sync_roles function for importing roles

### LINKED ISSUE #1091: Create a yaml config defining roles and provide a mechanism for loading it
We need a mechanism to load roles and permissions from a yaml file.  As part of this, we should create an exemplar config file with a reasonable default set of roles and permissions.

scott-taubman:
Closes #1091 

This PR adds the following:

* An example config, `roles.yaml`, which will eventually act as a reasonable set of default roles.
* A configuration parameter under `auth` called `role_definition_file` that you set to the location of your `roles.yaml` equivalent file.
* A `sync_roles` function that takes a dictionary containing role info and syncs the Roles in the database to match those definitions.
* A modified version of `ensure_roles` that runs on application startup.  This reads the yaml file referred to in the `auth.role_definition_file` configuration parameter and passes it over to the `sync_roles` function.

# Test Instructions
* Set the `auth.role_definition_file` configuration parameter in your application config file to the location of the `roles.yaml` example config file (or a copy of it if you house your config files elsewhere).
* Start beer-garden.
* Check that the Roles got created. You can do this in the DB directly, or by doing something like this in a python shell:
  ```python
  In [1]: from mongoengine import connect

  In [2]: from beer_garden.db.mongo.models import Role

  In [3]: connect(db="bg-parent1", host="mongodb://mongodb")  # change parameter values as appropriate for your setup
  Out[3]: MongoClient(host=['mongodb:27017'], document_class=dict, tz_aware=False, connect=True, read_preference=Primary())

  In [4]: [role.name for role in Role.objects.all()]
  Out[4]: ['job_manager', 'operator', 'read_only', 'superuser']
  ```
* Feel free to play around with the `roles.yaml` and verify that roles get added, removed, and modified when you expect them to.  Or just look at the unit tests and convince yourself that that would be the case.


src/app/example_configs/roles.yaml
```diff
@@ -0,0 +1,44 @@
+- name: "job_manager"
```
scott-taubman, line 1: The details of this file don't entirely matter right now.  It's just an example config that over time we'll build out to be whatever we think are reasonable default roles.

src/app/test/db/mongo/util_test.py
```diff
@@ -331,3 +299,48 @@ def test_ensure_local_garden_updates_garden_from_config(self, monkeypatch):
...
+
+        role_definition_file = config.get("auth.role_definition_file")
+
+        with patch(
+            "builtins.open", mock_open(read_data=role_definition_yaml)
+        ) as mock_file_read:
+            ensure_roles()
+            mock_file_read.assert_called_with(role_definition_file, "r")
+
+        assert len(Role.objects.filter(name="testrole1")) == 1
```
scott-taubman, line 338: Ideally the `sync_roles` call would be mocked here and I would just check that it got called (and not called in the next test), but I couldn't for the life of me get that mock to work.  I'm guessing there's something about the interplay between `pytest` and `patch` I'm not understanding. Either way I gave up and just resorted to letting it actually execute `sync_roles`, even though it kind of amounts to a duplicate of tests dedicated to that function.

src/app/test/role_test.py
```diff
@@ -0,0 +1,114 @@
...
+        with pytest.raises(ValidationError):
+            sync_roles(role_sync_data_missing_fields)
+
+    def test_sync_roles_removes_user_role_assignments(
+        self, role_sync_data, user_with_role_assignments
+    ):
+        """sync_roles should remove role assignments references Roles that are deleted
+        as part of the sync process
+        """
+        assert len(user_with_role_assignments.role_assignments) > 0
```
scott-taubman, line 109: I'm curious on opinions here.  It occurs to me that a lot of tests make assumptions about the state of the data before the test runs.  So in this case, my assertion at the end is:

```python
assert len(user_with_role_assignments.role_assignments) == 0
```

With **just** that assertion, someone could make a change to the test data such that `len(user_with_role_assignments.role_assignments)` starts at 0 and this test would continue to pass, even though it was no longer testing anything meaningful.  So I opted to put this upfront assertion to effectively validate that the data is in a state that makes my test meaningful, i.e. this test would now fail if there were no role_assignments to begin with, which would let us know that we've invalidated this test.

Jake-Ross: It sounds like a better idea to be explicit with exactly what behavior you are testing. Tests that have different assumptions about the data being tested should probably have different fixtures, or should modify the fixture in their own namespace. This assertion doesn't seem out of the ordinary given its function.

jlrpnbbngtn: I think this is a fine solution. `pytest` also supports xunit style, method-level setup and teardown, but that would be overkill in this situation IMO.

jlrpnbbngtn (approved):
A few comments, but no blockers. LGTM but won't merge until you've had the opportunity to look at the comments. Then either one of us can merge.

src/app/beer_garden/role.py
```diff
@@ -0,0 +1,97 @@
...
+
+        role_obj.save()
+
+
+def _delete_roles_not_in_list(roles: list) -> int:
+    """Delete all Roles in the database not represented in the provided list by name.
+
+    Args:
+       roles: A list of dictionaries containing role data, with any role not present
+         in the list being a candidate for deletion.
```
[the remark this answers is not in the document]
scott-taubman: No, the code is in line with the intention, which is that any roles not present will be deleted. So I can see how calling it a "candidate" for deletion is misleading. I'll update the wording.

src/app/beer_garden/db/mongo/util.py
```diff
@@ -40,70 +38,20 @@ def ensure_local_garden():
...
-                _create_role(role)
-            except NotUniqueError:
-                logger.warning("Role %s already exists" % role.name)
-
-    for role in mandatory_roles:
-        _create_role(role)
+    if role_definition_file:
+        logger.info(f"Syncing role definitions from {role_definition_file}")
+
+        with open(role_definition_file, "r") as filestream:
```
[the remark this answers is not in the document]
scott-taubman: Yeah, I thought about this a bit and I guess I never circled back and addressed it.  Basically the behavior right now if there is a problem either reading or parsing the file is that an unhandled exception gets thrown (`FileNotFoundError` or `marshmallow.exceptions.ValidationError`) and the startup of the application is effectively halted.

I considered allowing this to fail quietly (log an error and just move on), but that feels very likely to just result in someone never realizing that some change they  made to the role file is never actually being applied.

Maybe the right thing to do is handle those exceptions and raise a custom one with a message making it clear that there was a problem loading or parsing the role file.  That wouldn't change the behavior as it stands, but it should make it more clear why the application is not starting (at least in the parsing case.  The `FileNotFoundError` is pretty clear already).

src/app/beer_garden/role.py
```diff
@@ -0,0 +1,97 @@
...
+    """Remove a Role. This will also remove any references to the Role, such as those
+    in User role assignments.
+
+    Args:
+        role: The Role document object.
+
+    Returns:
+        None
+    """
+    remove_role_assignments_for_role(role)
```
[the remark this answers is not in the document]
scott-taubman: In this case it feels clear enough what's happening, but that pattern seems like a good one in cases where the intent could be unclear.

scott-taubman (commented):
Responded to comments. I'll make the indicated changes and push it up for re-review.

src/app/beer_garden/db/mongo/util.py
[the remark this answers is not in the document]
scott-taubman: I added friendlier error handling for 3 specific cases:

* The file doesn't exist at all
* The file exists, but doesn't conform to the input schema
* The file exists and the format is good, but something in the data is invalid (which I think can only be that you've specified an invalid permission for a role).

src/app/beer_garden/db/mongo/util.py
```diff
@@ -133,7 +102,7 @@ def ensure_users(guest_login_enabled):
         Principal(
             username="admin",
             hash=custom_app_context.hash(default_password),
-            roles=[LegacyRole.objects.get(name="bg-admin")],
```
scott-taubman, line 136: This reference to one of the old LegacyRole's, which I removed as part of this branch, was an issue that I didn't catch until I tested starting from a clean DB.  This entire `ensure_users` function will go or be rewritten at some point, but for now I've opted to make the minimum changes necessary to have the code not blow up.

jlrpnbbngtn (approved):
All LGTM.

PR #1123, opened 2021-09-16
Deep copying events before sending to handlers

### LINKED ISSUE #978: Events mangling event objects

Jake-Ross:
Closes #978 

Sending references to events before sending them to handlers might mangle the event between handler calls. This PR sends a copy instead to prevent that behavior.

## To Test
Affected handlers are in the following files. Doing things in the UI that create events involving each of these modules would be sufficient testing.
```python
beer_garden.application.handle_event,
beer_garden.garden.handle_event,
beer_garden.plugin.handle_event,
beer_garden.requests.handle_event,
beer_garden.router.handle_event,
beer_garden.systems.handle_event,
beer_garden.scheduler.handle_event,
beer_garden.log.handle_event,
beer_garden.files.handle_event,
beer_garden.local_plugins.manager.handle_event
```

### DIFF
--- a/src/app/beer_garden/api/http/__init__.py
+++ b/src/app/beer_garden/api/http/__init__.py
@@ -4,6 +4,7 @@ import os
 import ssl
 import types
 from typing import Optional, Tuple
+from copy import deepcopy
 
 from apispec import APISpec
 from brewtils.models import Event, Events, Principal
@@ -370,6 +371,6 @@ def _event_callback(event):
         beer_garden.requests.handle_event,
     ]:
         try:
-            handler(event)
+            handler(deepcopy(event))
         except Exception as ex:
             logger.exception(f"Error executing callback for {event!r}: {ex}")
--- a/src/app/beer_garden/events/handlers.py
+++ b/src/app/beer_garden/events/handlers.py
@@ -1,5 +1,6 @@
 # -*- coding: utf-8 -*-
 import logging
+from copy import deepcopy
 from brewtils.models import Event
 
 import beer_garden.config
@@ -45,6 +46,6 @@ def garden_callbacks(event: Event) -> None:
         beer_garden.local_plugins.manager.handle_event,
     ]:
         try:
-            handler(event)
+            handler(deepcopy(event))
         except Exception as ex:
             logger.exception(f"Error executing callback for {event!r}: {ex}")
--- a/src/app/beer_garden/events/handlers.py
+++ b/src/app/beer_garden/events/handlers.py
@@ -34,7 +34,7 @@ def garden_callbacks(event: Event) -> None:
 
     # These are all the MAIN PROCESS subsystems that care about events
     for handler in [
-        #beer_garden.application.handle_event,
+        beer_garden.application.handle_event,
         beer_garden.garden.handle_event,
         beer_garden.plugin.handle_event,
         beer_garden.requests.handle_event,
--- a/src/app/beer_garden/events/handlers.py
+++ b/src/app/beer_garden/events/handlers.py
@@ -34,7 +34,7 @@ def garden_callbacks(event: Event) -> None:
 
     # These are all the MAIN PROCESS subsystems that care about events
     for handler in [
-        beer_garden.application.handle_event,
+        #beer_garden.application.handle_event,
         beer_garden.garden.handle_event,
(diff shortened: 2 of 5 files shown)


Jake-Ross:
I tried to create a unit test to ensure that the `event` can't be mangled by when it's `deepcopy`'d, but importing the handlers module seems to have some sort of circular dependency that disallows it.

```python
>>> import beer_garden.events.handlers
  File "<stdin>", line 1, in <module>
  File "/home/jrmartu/workspace/beer-garden/src/app/beer_garden/events/handlers.py", line 7, in <module>
    import beer_garden.files
  File "/home/jrmartu/workspace/beer-garden/src/app/beer_garden/files.py", line 23, in <module>
    import beer_garden.router as router
  File "/home/jrmartu/workspace/beer-garden/src/app/beer_garden/router.py", line 141, in <module>
    "FILE_CREATE": beer_garden.files.create_file,
AttributeError: module 'beer_garden' has no attribute 'files'
```

`handlers` imports `files`, which imports `router`, which imports `files` again. Is this a circular dependency, or am I not importing correctly?

jlrpnbbngtn:
> `handlers` imports `files`, which imports `router`, which imports `files` again. Is this a circular dependency, or am I not importing correctly?

This is certainly a circular import. Note the comment at the top of the only place beer_garden.router is used in files.py:

*HEADS UP - THIS IS PROBABLY BROKEN*

This sounds familiar.... ;-)

At any rate, if you move the line

`import beer_garden.router as router`

from the top of files.py and put it as the first line in `forward_file`, you should be able to proceed.

I believe this to be safe, though it's kludgy and this kind of problem probably reveals a design flaw. You might consider opening a ticket to redesign the modules to completely eliminate the circular dependency.

src/app/beer_garden/api/http/__init__.py
```diff
@@ -370,6 +371,6 @@ def _event_callback(event):
         beer_garden.requests.handle_event,
     ]:
         try:
-            handler(event)
+            handler(deepcopy(event))
```
jlrpnbbngtn, line 374: There's a built-in assumption here that I feel should be examined.

Without getting into the specifics of how any one handler does its business, the code here assumes that if a handler throws an exception, the repr of the *original* version of the event would make for a meaningful error message to the user. In other words, it assumes that the handler doesn't mutate the event in a way that I'll care about (and will be able to see) before it blows up.

Perhaps in our specific case this won't matter at all, but as a general pattern I can see this being problematic.

src/app/beer_garden/events/handlers.py
```diff
@@ -33,7 +34,7 @@ def garden_callbacks(event: Event) -> None:
 
     # These are all the MAIN PROCESS subsystems that care about events
     for handler in [
-        beer_garden.application.handle_event,
+        #beer_garden.application.handle_event,
```
jlrpnbbngtn, line 37: I'm being pedantic, so sorry, but I think that commented out code should be deleted in nearly every case. We can find the history through git if we need to see how things used to be done.

jlrpnbbngtn (commented):
A couple of comments you should review to see if you agree. Also, linter is failing. Otherwise looks good.

src/app/beer_garden/events/handlers.py
Jake-Ross: I actually wasn't removing it, as it's likely still required. However, there is a weird issue that prevents patching this module:

`beer_garden.application` isn't actually a module, it's instantiated in `beer_garden.__main__`. When `handlers.py` runs, one of the imports puts it in the namespace. However, this is *not* instantiated when running a test. I'm still trying to figure out an elegant way to get around that (it's stumped me for a bit).

src/app/beer_garden/api/http/__init__.py
Jake-Ross: It's intuition that handlers should be writing information on errors into the *exception* they throw back here, not the event, right? 

Seeing the original event seems like enough information for a normal user. Showing how an event handler modified an event in an error message seems like information only a dev would need to know while debugging.

Jake-Ross:
Reviewed on Friday with @scott-taubman, ready for merge.

PR #1128, opened 2021-09-21
One line revert was not actually one line

jlrpnbbngtn:
Integration tests are repaired. Though there remains a question of whether we're actually testing the functionality that we want to see in the garden stomp tests, but that discussion can wait for another day. In any case, local and remote integration tests pass after these edits.

Fundamental problem was in what seemed to be an innocent correction to the router code, adding in a default for a dictionary's `get` call. There's a comment in `src/app/beer_garden/router.py` that warns against doing the same again. But as the comment implies, we're doing checks before any routing occurs to handle different cases, but it seems unlikely that code ever runs. So we should look into that as well (but beware the attendant *whole other cans of worms*).

### DIFF
--- a/src/app/beer_garden/router.py
+++ b/src/app/beer_garden/router.py
@@ -471,7 +471,14 @@ def _pre_route(operation: Operation) -> Operation:
             operation.model.namespace = config.get("garden.name")
 
     elif operation.operation_type == "SYSTEM_READ_ALL":
-        if operation.kwargs.get("filter_params", {}).get("namespace", "") == "":
+        # what looks like a sensible edit is to change the next line to:
+        #   operation.kwargs.get("filter_params", {}).get("namespace", "") == "":
+        # but that will break everything, as it turns out. So, unless
+        # operation.kwargs has a key called filter_params whose value is a dictionary
+        # that has a key called namespace and its value is an empty string, the follow-
+        # ing never runs. Intuition says that's not what was intended here, so
+        # TODO: look into this
+        if operation.kwargs.get("filter_params", {}).get("namespace") == "":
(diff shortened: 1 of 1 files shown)


src/app/beer_garden/router.py
```diff
@@ -471,7 +471,14 @@ def _pre_route(operation: Operation) -> Operation:
...
     elif operation.operation_type == "SYSTEM_READ_ALL":
-        if operation.kwargs.get("filter_params", {}).get("namespace", "") == "":
+        # what looks like a sensible edit is to change the next line to:
+        #   operation.kwargs.get("filter_params", {}).get("namespace", "") == "":
+        # but that will break everything, as it turns out. So, unless
+        # operation.kwargs has a key called filter_params whose value is a dictionary
+        # that has a key called namespace and its value is an empty string, the follow-
+        # ing never runs. Intuition says that's not what was intended here, so
+        # TODO: look into this
+        if operation.kwargs.get("filter_params", {}).get("namespace") == "":
```
scott-taubman, line 481: The reason this change breaks everything is that it was now the case that if you defined no filter params at all, we would effectively apply a default filter limiting it to the local garden only.  This is clearly different behavior than before, where it would only default if we got here and a namespace filter was present, but set to empty string.

I agree that it is not obvious from the code **why** someone would supply a filter of `namespace = ""`, nor why the expected behavior in that case would be that we treat that as `namespace = <local garden>`.  It wouldn't surprise me at all if that behavior isn't in fact desired, but it didn't matter because it's never induced (i.e. nothing actually gets us here with the `namespace = ""` filter param).

jlrpnbbngtn (commented):
Looks good.

PR #1129, opened 2021-09-22
#1038 - Better check for when requests need to be stored in GridFS

### LINKED ISSUE #1038: Mongo Request model gridfs storage

Jake-Ross:
Closes #1038 

Refactors `save()` in the `Request` Mongo model to transfer the request parameters or request output to GridFS. According to #1038, setting a cap of 5MB on both of these should keep the total size of the Request below 16MB (including other request data that needs to be saved), which is Mongo's limit..

## To test
- Run the unit tests

The unit tests may get more coverage with better Mongo model fixtures, but right now there are none.

### DIFF
--- a/src/app/beer_garden/db/mongo/models.py
+++ b/src/app/beer_garden/db/mongo/models.py
@@ -74,6 +74,8 @@ __all__ = [
     "User",
 ]
 
+REQUEST_MAX_PARAM_SIZE = 5 * 1_000_000
+
 
 class MongoModel:
     brewtils_model = None
@@ -416,29 +418,20 @@ class Request(MongoModel, Document):
     def save(self, *args, **kwargs):
         """Save a request, moving request attributes to GridFS if too big"""
         self.updated_at = datetime.datetime.utcnow()
-        # 5MB
-        max_param_size = 5 * 1_000_000
-        max_output_size = 5 * 1_000_000
         encoding = "utf-8"
 
-        def get_size(data):
-            # This func may not work for multi-byte encodings (e.g. utf-16)
-            klen = sum(len(k.encode(encoding)) for k in data.keys())
-            vlen = sum(len(v.encode(encoding)) for v in data.values())
-            return klen + vlen
-
         if self.parameters:
-            if get_size(self.parameters) > max_param_size:
+            params_json = json.dumps(self.parameters)
+            if len(params_json) > REQUEST_MAX_PARAM_SIZE:
                 self.logger.debug("Parameters too bg, storing in GridFS")
-                self.parameters_gridfs.put(
-                    json.dumps(self.parameters), encoding=encoding
-                )
+                self.parameters_gridfs.put(params_json, encoding=encoding)
                 self.parameters = None
 
         if self.output:
-            if len(str(self.output)) > max_output_size:
+            output_json = json.dumps(self.output)
+            if len(output_json) > REQUEST_MAX_PARAM_SIZE:
                 self.logger.info("Output size too big, storing in gridfs")
-                self.output_gridfs.put(json.dumps(self.output), encoding=encoding)
+                self.output_gridfs.put(output_json, encoding=encoding)
                 self.output = None
 
         super(Request, self).save(*args, **kwargs)
--- a/src/app/beer_garden/db/mongo/models.py
+++ b/src/app/beer_garden/db/mongo/models.py
@@ -401,39 +401,45 @@ class Request(MongoModel, Document):
     logger = logging.getLogger(__name__)
 
     def pre_serialize(self):
+        """Pull any fields out of GridFS"""
         encoding = "utf-8"
-        """If string output was over 16MB it was spilled over to the GridFS storage solution"""
+
         if self.output_gridfs:
-            self.logger.debug("~~~Retrieving output from gridfs")
+            self.logger.debug("Retrieving output from GridFS")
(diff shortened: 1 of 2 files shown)


src/app/beer_garden/db/mongo/models.py
```diff
@@ -401,39 +400,45 @@ class Request(MongoModel, Document):
...
-        parameter_size = sys.getsizeof(self.parameters)
 
-        if self.output:
-            parameter_size = sys.getsizeof(self.parameters)
-            if parameter_size > max_size:
-                self.logger.info("~~~Parameter size too big, storing in gridfs")
+        def get_size(data):
+            # This func may not work for multi-byte encodings (e.g. utf-16)
+            klen = sum(len(k.encode(encoding)) for k in data.keys())
+            vlen = sum(len(v.encode(encoding)) for v in data.values())
```
scott-taubman, line 427: I haven't fully tested or reviewed this yet, but quick comment here because I immediately ran into a failure on this because this line seems to assume that the values are some encodable type?  Specifically I tasked something that takes a boolean parameter (I think it was "echo") and this blew up.

I think I originally suggested just stringifying the parameters and checking the length.  Ultimately that's how the parameters_gridfs get stored (see the `json.dumps(self.parameters)` line a little below this one. I'm open to the idea of that being problematic for its own reasons.

I'll look at this some more tomorrow, but just wanted to get this comment on here in the meantime.

scott-taubman: Yeah, following up on this, the stringified json (the results of the json.dumps) seems like the thing you want to check the length of.  So I think maybe what you want to do for both the parameters and the output is to do the `json.dumps` first, then just check the length of that where you're doing the `get_size` call.  So something like:

```python
parameters_json = json.dumps(self.parameters)
output_json = json.dumps(self.output)

if self.parameters:
    if len(parameters_json) > max_param_size:
        ...
        ...
```

src/app/test/db/mongo/models_test.py
```diff
@@ -242,6 +242,50 @@ def test_template_check(self):
...
+            namespace="barfoo",
+        )
+        req.parameters = {"message": "hi"}
+        req.output = "bye"
+        req.parameters_gridfs.put = Mock()
+        req.output_gridfs.put = Mock()
+        return req
+
+    def test_save_stores_in_gridfs_after_maxsize(self, request_model):
+        request_model.parameters = {"message": "a" * 6 * 1_000_000}
```
scott-taubman, line 262: The tests you have here are generally fine, but I suggest we tweak them so that they can work independent of the actual `max_param_size` and `max_output_size` values.  There are two main reasons to do this:

* The tests as written are a bit too dependent on what the actual max size settings happen to be, which isn't really what you care about of the purposes of the test. This means that if someone came along and changed those max sizes to 7MB instead of 5MB, the tests now fail.  The 5MB limit isn't what we're trying to test, it's that whatever the limit is, it gets enforced.

* Generating a 6MB string is all fast and fine, but what if the limit was 500MB? Or 5GB? Generating a sufficiently large object in those cases wouldn't be great.

So I think what you want to do instead is move those max size variables somewhere mockable (constants at the top of the file maybe?) and then mock the values in your tests to be some arbitrary, small value that your tests are in control of.

scott-taubman (changes requested):

src/app/beer_garden/db/mongo/models.py
Jake-Ross: Unsure of where I got this from, but was under the impression that the Request.parameters could only be type `Dict` with `String`s inside. Most recent commit fixes bools breaking it, as it no longer tries to encode (even though it needs the encoding for the `.put()` method).

src/app/test/db/mongo/models_test.py
Jake-Ross: I moved these to a constant at the beginning of `models.py`. I kept the encoding there, though, as I'm pretty sure the encoding will never need to be changed from UTF-8 for our purposes.

scott-taubman (approved):
Looks good and works as expected from my tests.

PR #1136, opened 2021-09-29
Issue/1101 - Delete stale systems when syncing child garden

### LINKED ISSUE #1101: Garden Sync with incompatible systems
It looks like the following scenario results in a complete failure of the sync:

1. Parent and child are synced
2. Stop the parent garden so it doesn't receive updates from the child garden
2. A system is removed on the child
3. A "new" system with the same name+version, ~but different commands~, is created on the child. EDIT: The commands don't need to be different - just removing and re-adding the system (so it has a different ID)
4. Start the parent
4. Parent initiates a sync

The parent will attempt to update the "replaced" system definition but will fail because of the commands mismatch.

jlrpnbbngtn:
Closes #1101

To verify the fix for the corner case (running BG in a parent-child configuration,, taking the parent offline, changing Systems on the child, bringing the parent back up and then syncing), a straightforward way is to follow the steps in #1101 while tailing the log of the parent and then seeing that an error is no longer thrown. One could also track the entries in the database across the operations to make sure they are consistent with the actions taken and expected results. This was done with single parent and single child with each of http and stomp entrypoints and tested with multiple children and a grandchild with http. Corner case error appears to be resolved in all of these scenarios.

For the general case, two unit tests are provided to verify that any unknown Systems associated with a remote garden are removed prior to saving the garden to the database.

### DIFF
--- a/src/app/beer_garden/db/mongo/models.py
+++ b/src/app/beer_garden/db/mongo/models.py
@@ -757,62 +757,66 @@ class Garden(MongoModel, Document):
     }
 
     def deep_save(self):
-        logger = logging.getLogger(self.__class__.__name__)
-
-        # if the call to this method is on a child garden object, we ensure that
-        # when saving the systems, unknowns are deleted
         if self.connection_type != "LOCAL":
-            # import moved here to avoid a circular import loop
-            from beer_garden.systems import get_systems, remove_system
-
-            def _get_system_triple(system: System) -> Tuple[str, str, str]:
-                return (
-                    system.namespace,
-                    system.name,
-                    system.version,
-                )
-
-            our_namespaces = set(self.namespaces).union(
-                set(map(attrgetter("namespace"), self.systems))
-            )
-            # we leverage the fact that systems must be unique up to the triple of their
-            # namespaces, names and versions
-            child_systems_already_known = {
-                _get_system_triple(system): str(system.id)
-                for system in get_systems(
-                    filter_params={"local": False, "namespace__in": our_namespaces}
-                )
-            }
-
-            for system in self.systems:
-                triple = _get_system_triple(system)
-
-                if triple in child_systems_already_known:
-                    system_id_to_remove = child_systems_already_known.pop(triple)
-
-                    if system_id_to_remove != str(system.id):
-                        logger.debug(
-                            f"Removing System <{triple[0]}"
-                            f", {triple[1]}"
-                            f", {triple[2]}> with ID={system_id_to_remove}"
-                            f"; doesn't match ID={str(system.id)}"
-                            " for known system with same attributes"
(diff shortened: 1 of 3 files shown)


src/app/test/db/mongo/models_test.py
```diff
@@ -612,3 +645,35 @@ def test_only_one_local_garden_may_exist(self, testgarden):
...
+        new_vers = set(
+            map(
+                lambda x: str(getattr(x, "version")), child_garden.systems  # noqa: B009
+            )
+        )
+
+        assert self.v1_str not in new_vers and self.v2_str in new_vers
+        assert new_ids.intersection(orig_ids) == set()
+
+        child_system_v2.delete()
```
scott-taubman, line 679: This cleanup seems like it should happen in the fixture.

src/app/test/db/mongo/models_test.py
```diff
@@ -612,3 +645,35 @@ def test_only_one_local_garden_may_exist(self, testgarden):
         should raise an exception"""
         with pytest.raises(NotUniqueError):
             Garden(name=f"not{testgarden.name}", connection_type="LOCAL").save()
+
+    def test_child_garden_system_update(self, child_garden, child_system_v2):
```
scott-taubman, line 649: This test is a good one to have, but should we also have one that tests that the update works when a new system with a matching "triple" shows up?  That's the case that inspired these changes in the first place.

src/app/beer_garden/db/mongo/models.py
```diff
@@ -755,8 +757,60 @@ class Garden(MongoModel, Document):
...
 
     def deep_save(self):
-        for system in self.systems:
-            system.save()
+        logger = logging.getLogger(self.__class__.__name__)
+
+        # if the call to this method is on a child garden object, we ensure that
+        # when saving the systems, unknowns are deleted
+        if self.connection_type != "LOCAL":
+            # import moved here to avoid a circular import loop
```
scott-taubman, line 765: It feels appropriate to me to have this logic contained in its own function.  Then the `deep_save` would just look something like:

```python
def deep_save(self)
    if self.connection_type != "LOCAL":
        self._update_associated_systems()
        
    self.save()
```

scott-taubman (changes requested):

src/app/beer_garden/db/mongo/models.py
jlrpnbbngtn: This is a pattern I see in your code a lot. I'm not against the habit, so sure. Done.

scott-taubman (approved):
I think we can merge this, but let's wait until we fix the build issue on develop first (hopefully #1138 does the trick) before introducing additional changes.

PR #1137, opened 2021-09-29
#817 - Separate local and remove plugin logging config files

### LINKED ISSUE #817: Separate local and remote plugin logging config files
We should split these out by default for the rpm install.

Jake-Ross:
Closes #817

The default for plugin logging config is now split between `local-plugin-logging.yaml` and `remote-plugin-logging.yaml`. RPM has been tested to write these configs on installation.

## To Test
- To test, you can either build this branch as an RPM and test that it correctly builds the configs, or make `./after_install` executable and run the script from there. I did method the former, so YMMV on the latter.


rpm/centos7/scripts/after_install.sh
```diff
@@ -65,9 +66,10 @@ if [ ! -f "$APP_LOG_CONFIG" ]; then
 fi
 
 # Generate plugin logging config if it doesn't exist
-if [ ! -f "$PLUGIN_LOG_CONFIG" ]; then
+if [ ! -f "$REMOTE_PLUGIN_LOG_CONFIG" ] || [ ! -f "$LOCAL_PLUGIN_LOG_CONFIG" ]; then
```
scott-taubman, line 69: I wouldn't expect the remote and local plugin log config files that get generated to be any different and looking at the changes to `config.py` it appears that they are in fact the same.  That makes me think that the change to have separate `local_conf` and `remote_conf` parameters isn't really what you want.  Instead, I think it makes sense to just run this command twice, once with `--config-file "$LOCAL_PLUGIN_LOG_CONFIG"` and then again with `--config-file "$REMOTE_PLUGIN_LOG_CONFIG"`.

This becomes moot if you do the above, but I think as written this would overwrite a present config file if the other one happened to be missing, which seems like not the desired result.

src/app/beer_garden/config.py
```diff
@@ -179,8 +179,10 @@ def generate_plugin_logging(args: Sequence[str]) -> dict:
 
     Args:
         args: Command line arguments
-            --config-file: Configuration will be written to this file (will print to
-                stdout if missing)
+            --local_conf: Local plugin configuration will be written to this file
```
scott-taubman, line 182: This is related to the comment about just running the `generate_plugin_logging_config` command twice, but basically I think you don't necessarily need to change this file at all.

scott-taubman (changes requested):

rpm/centos7/scripts/after_install.sh
Jake-Ross: This script is only run on installation, so I can't see a situation this would be run and the script overwrites a file. Regardless, I do agree that it's weird to do it this way. Fixed in latest.

scott-taubman (approved):
Looks good.  I built and installed the RPM and everything ended up where I expected.

PR #1138, opened 2021-09-30
#817 - Update plugin logging references

### LINKED ISSUE #817: Separate local and remote plugin logging config files

scott-taubman:
Closes #817 

This updates references to the `plugin-logging.yaml` that needed to be updated to the newly split out `remote-plugin-logging.yaml` and `local-plugin-logging.yaml`, but were overlooked in the original PR for #817.

There are some references to a `plugin-logging.yaml` in the `test/local_integration` folder as well, but as far as I can tell those are all referencing a copy that is local to that folder, rather than the one that used to exist in the example_configs folder.  For that reason, I left those references untouched.


jlrpnbbngtn (approved):
LGTM (famous last words ;-)

PR #1145, opened 2021-10-05
#1144 - New get_current_user to retrieve User from access token

### LINKED ISSUE #1144: Update get_current_user to get the appropriate User object for a request
The `get_current_user` method in the `BaseHandler` class needs to be updated to get the User object based on the JWT auth token that gets supplied from the `/api/v1/login` endpoint.

The permissions that are stamped into the JWT should be used to populate the permissions on the User object so that they are not unnecessarily re-calculated.

If no JWT is present or the JWT is expired, a 401 status should be returned.

scott-taubman:
Closes #1144 

## Description
This PR overhauls the `get_current_user` method on the `BaseHandler` to retrieve the user based on a supplied JWT auth token that the user had previously received via the `/api/v1/login` endpoint. For context, some very basic info on Tornado's `current_user` call can be found [here](https://www.tornadoweb.org/en/stable/web.html#tornado.web.RequestHandler.current_user).

This PR also removes all remaining remnants of the old (unused) authorization code. I had to bite the bullet and do this because adding the new stuff was just too confusing with all of the old references in there.

**NOTE**: I completely removed three v1 api endpoints: `permissions`, `role` and `token`. That is obviously a "breaking" change to the api, but seeing as there was no path to meaningfully using those endpoints, I don't think there's any value in suffering the pain of maintaining them as we overhaul the auth code.

## Code Review Instructions
This PR is purposefully split into two commits and will be much easier to review if the two are looked at separately:

* [`7204333`](https://github.com/beer-garden/beer-garden/pull/1145/commits/7204333ce535fff7ff45f1c2faeca127edbbdefb): This initial commit is a ton of code cleanup related to the old authorization code, as detailed in the description above.
* [`876f53e`](https://github.com/beer-garden/beer-garden/pull/1145/commits/876f53e2c9ac1ee709831faf54fa2a053c14f801): This is the commit with the actual new stuff. The changes here are mostly pretty straightforward.

## Test Instructions
I'm going to list some URLs that reference `localhost:8080`. Substitute in appropriate values for however you are running your beer-garden instance. Also, I'm using `curl`, but you're free to use the swagger docs or whatever else you're comfortable with.

The only endpoint that currently enforces the authorization requirement is the new `whoami` endpoint.  That's what we'll be using in our tests here.

### Auth Disabled
Make sure your auth --> enabled setting in your yaml config is set to `false`.

* If you hit the new `whoami` endpoint, you should see the anonymous user returned:
  ```shell
  curl GET http://localhost:8080/api/v1/whoami
  # {"role_assignments": [], "id": null, "username": "anonymous"}
  ```

### Auth Enabled
Make sure your auth --> enabled setting in your yaml config is set to **`true`** this time. You'll have to restart beer-garden after changing the setting.

* Try to access the `whoami` endpoint without supplying an auth token:
  ```shell
  curl http://localhost:8080/api/v1/whoami
  # {"message": "Authorization required"}
  ```
* Create a user to test with:
  ```shell
  curl -X POST --data '{"username": "obiwan", "password": "jedimaster"}' http://localhost:8080/api/v1/users
  ```
* Login with your new user:
  ```shell
  curl -s -X POST --data '{"username": "obiwan", "password": "jedimaster"}' http://localhost:8080/api/v1/login
  ```
  You are going to need the token that comes back in that response. Copy it and put it in a file somewhere.  If you have `jq` installed, you can just run the following to get the token to a file:
  ```shell
  curl -s -X POST --data '{"username": "obiwan", "password": "jedimaster"}' http://localhost:8080/api/v1/login | jq -r '.token' > /tmp/obiwan_token
  ```
* Now let's try to hit the `whoami` endpoint again, this time using our auth token:
  ```shell
  curl -H "Authorization: Bearer `cat /tmp/obiwan_token`" http://localhost:8080/api/v1/whoami
  # {"id": "615c967fe87a256e32aa94fe", "username": "obiwan", "role_assignments": []}
  ```
* Just to be safe, let's provide an invalid token and make sure it's rejected:
  ```shell
  curl -H "Authorization: Bearer some.invalid.token" http://localhost:8080/api/v1/whoami
  # {"message": "Access token invalid"}
  ```


src/app/beer_garden/api/http/authentication/__init__.py
```diff
@@ -49,9 +49,14 @@ def generate_access_token(user: User) -> str:
     jwt_headers = {"alg": "HS256", "typ": "JWT"}
     jwt_payload = {
         "sub": str(user.id),
-        "exp": datetime.utcnow() + timedelta(hours=12),
+        "exp": _get_token_expiration(),
```
scott-taubman, line 52: I changed this to a function call to make it easy to mock in my unit test.  It's otherwise identical to what was there.

src/app/beer_garden/api/http/handlers/v1/user.py
```diff
@@ -147,3 +147,26 @@ def post(self):
         create_user(**user_data)
 
         self.set_status(201)
+
+
+class WhoAmIAPI(BaseHandler):
```
scott-taubman, line 152: I added this endpoint specifically to provide a simple thing to test against. It may or may not be worth keeping around in the end, but certainly having it doesn't hurt anything and until we're done with all of the auth changes it'll be useful to have.

src/app/beer_garden/db/mongo/models.py
```diff
@@ -886,6 +887,50 @@ class User(Document):
         "indexes": [{"name": "unique_index", "fields": ["username"], "unique": True}],
     }
 
+    _permissions_cache: Optional[dict] = None
+
+    @property
+    def permissions(self) -> dict:
```
scott-taubman, line 893: Hopefully between the docstrings and the actual usage it's clear why I added this permissions caching related stuff to the User model.  Do ask questions or provides suggestions though if it's not so clear.

src/app/beer_garden/api/http/exceptions.py
```diff
@@ -0,0 +1,51 @@
+from typing import Optional
+
+from tornado.web import HTTPError
+
+
+class BaseHTTPError(HTTPError):
```
scott-taubman, line 6: See #1146 for some info on this. I originally just kept following the same bad pattern here and then opened that issue to fix it everywhere, but decided to come back and just get things started.

src/app/beer_garden/api/http/handlers/v1/user.py
```diff
@@ -147,3 +148,26 @@ def post(self):
         create_user(**user_data)
 
         self.set_status(201)
+
+
+class WhoAmIAPI(AuthorizationHandler):
```
scott-taubman, line 153: I added this endpoint specifically to provide a simple thing to test against. It may or may not be worth keeping around in the end, but certainly having it doesn't hurt anything and until we're done with all of the auth changes it'll be useful to have.

src/app/beer_garden/db/mongo/models.py
jlrpnbbngtn: This looks fine. Different from the other caching property, this doesn't rely on a property access as the sole method to manipulate the internal state of an object, unlike what the framework forced you to do above.

src/app/beer_garden/api/http/handlers/authorization_handler.py
```diff
@@ -0,0 +1,119 @@
...
+            raise AuthorizationRequired
+
+        try:
+            token = auth_header.split(" ")[1]
+            algorithm = jwt.get_unverified_header(token)["alg"]
+
+            return jwt.decode(token, key=secret_key, algorithms=[algorithm])
+        except (jwt.InvalidSignatureError, jwt.DecodeError):
+            raise InvalidToken
+        except (jwt.ExpiredSignatureError):
```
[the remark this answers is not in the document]
scott-taubman: Fair enough.  Clean this up.

PR #1155, opened 2021-10-13
#1097 - Ignore hidden directories

### LINKED ISSUE #1097: Exclude plugins directories that begin with a period
Every time Beergarden starts up, if you're using the [example_plugins](https://github.com/beer-garden/example-plugins) repo, you'll run into errors like this:

```
beer_garden.local_plugins.manager - ERROR - Error loading plugin at plugins/.git: Config file plugins/.git/beer.conf does not exist
beer_garden.local_plugins.manager - ERROR - Error loading plugin at plugins/.github: Config file plugins/.github/beer.conf does not exist
```

Beergarden should skip scanning hidden dirs (dirs that begin with `.`) for plugins.

jlrpnbbngtn:
Fixes #1097 

## Testing
1. For ignoring directories that start with a dot: create such a directory in the *plugins* directory your installation points to. Make sure the BG logging is set to *DEBUG* and note that a logging message with *"Ignoring hidden plugin path"* is displayed. There are unit tests that further verify this.
2. For `None` making its way out of `scan_path`: ignoring dot files essentially solves this. However, one could manually trace execution through `scan_path`, noting that the called function, `_create_runners`, has no path to create a runner whose arguments are `None`. There was a bit of refactoring of the `PluginManager` class as part of this ticket. There are unit tests that verify that behavior is now predictable in the cases of bad paths/config files passed and when methods that return `Runner`s/`Instance`s are called without arguments.

### DIFF
--- a/src/app/beer_garden/local_plugins/manager.py
+++ b/src/app/beer_garden/local_plugins/manager.py
@@ -167,9 +167,7 @@ class PluginManager(StoppableThread):
         username=None,
         password=None,
     ):
-        super().__init__(
-            **{"logger": logging.getLogger(__name__), "name": "PluginManager"}
-        )
+        super().__init__(logger=logging.getLogger(__name__), name="PluginManager")
 
         self._display_name = "Plugin Manager"
         self._runners: List[ProcessRunner] = []
@@ -401,7 +399,7 @@ class PluginManager(StoppableThread):
             List of Runners to run the plugins.
         """
         if paths is None and self._plugin_path is None:
-            self.logger.exception("PluginManager.scan_path has no path to scan")
+            self.logger.error("PluginManager.scan_path has no path to scan")
             return []
 
         plugin_paths = paths or self._plugin_path.iterdir()


src/app/beer_garden/local_plugins/manager.py
```diff
@@ -277,159 +343,229 @@ def stop_one(
...
+            return []
 
+        plugin_paths = paths or self._plugin_path.iterdir()
         new_runners = []
 
         try:
             for path in plugin_paths:
                 try:
-                    if path.is_dir() and path not in self.paths():
+                    if self._is_valid_plugin_path(path, self.paths(), self.logger):
```
jlrpnbbngtn, line 417: All checks for determining whether a path is a valid plugin directory are factored out into a single function, rather than having some here and some elsewhere.

src/app/test/local_plugins/manager_test.py
```diff
@@ -93,23 +127,26 @@ def test_exception(self, monkeypatch, tmp_path, loader):
         loader.load_plugins(path=tmp_path)
 
 
-@pytest.mark.skip
 class TestLoadNew(object):
-    def test_empty_path(self, tmp_path, manager):
-        manager.load_new(path=tmp_path)
-        assert manager._runners == {}
+    def test_path_with_no_plugins(self, tmp_path, manager):
```
jlrpnbbngtn, line 131: Since these are adjacent in functionality to what needed to be tested for this ticket, it was low-effort to fix them and reuse the new logic.

src/app/beer_garden/local_plugins/manager.py
```diff
@@ -277,159 +343,225 @@ def stop_one(
...
 
-        for runner in runners:
-            self._runners.remove(runner)
+        for the_runner in the_runners:
+            self._runners.remove(the_runner)
+
+    def _get_runner_id(self) -> str:
+        """Get a 10-letter string to serve as a runner ID."""
+        if self._runner_id_generator is None:
+            self._runner_id_generator = self._runner_id_generator_factory()
```
scott-taubman, line 517: I'm not the expert on generators, but isn't this going to effectively reset your stored `seen` set every time because you're calling the generator fresh each time?  Meaning I think you're going to end up with the same behavior, where duplicates are still very much possible.

It seems like we should just generate an actual uuid, but also I'm not sure you need to fix the issue you were trying to address here in this particular PR.

scott-taubman: Per our offline discussion, I completely glossed over the fact that you're' storing the `__runner_id_generator` and then only assigning if it isn't already set.  In which case disregard my concern and carry on.

PR #1156, opened 2021-10-13
#1066 Add an API endpoint for ad-hoc runs of a job

### LINKED ISSUE #1066: Add an api endpoint for ad-hoc runs of a job
We want to be able to make an API call that will do an ad-hoc run of an existing scheduled job.  The endpoint should accept the following input:

* job id (string, path parameter)
* interval reset (boolean, query parameter)

So a request would something like:

/api/v1/jobs/<id>/execute?reset_interval=true

Jake-Ross:
Closes #1066 

Requires brewtils [PR#359](https://github.com/beer-garden/brewtils/pull/359).

There are a bunch of ways to 'execute a job ad-hoc' that all would accomplish the same thing, so how this problem is approached could be changed drastically if need be. 

The way this PR accomplishes 'executing' is by creating an 'anonymous' job with the scheduler with the trigger `datetime.now()`. There may be an issue with this, as creating a job in the scheduler with a conflicting ID will overwrite the previous job. However, setting the ID to ```None``` allows both the original job and the 'anonymous' job to live in harmony. Since the anonymous job only exists for a moment before spawning a new process to run the request, it's unlikely that an id of `None` will conflict with other ad-hoc executions (but may be possible?).

## To Test
- Create a job.
- Use the above brewtils branch to use EasyClient.execute_job(). 
```python
from brewtils.rest.easy_client import EasyClient
client = EasyClient(bg_host='localhost', bg_port=2337, ssl_enabled=False)
client.can_connect()
# True
client.execute_job('<job id>')
```
- You should try this with the job both paused and unpaused.

### DIFF
--- a/src/app/beer_garden/api/http/handlers/v1/job.py
+++ b/src/app/beer_garden/api/http/handlers/v1/job.py
@@ -1,9 +1,9 @@
 # -*- coding: utf-8 -*-
-from mongoengine.errors import ValidationError
 from brewtils.errors import ModelValidationError, NotFoundError
 from brewtils.models import Operation
 from brewtils.schema_parser import SchemaParser
 from brewtils.schemas import JobExportInputSchema, JobSchema
+from mongoengine.errors import ValidationError
 
 from beer_garden.api.http.base_handler import BaseHandler
 


src/app/beer_garden/api/http/handlers/v1/job.py
```diff
@@ -57,6 +57,12 @@ async def patch(self, job_id):
           ```JSON
           { "operation": "update", "path": "/status", "value": "RUNNING" }
           ```
+
+          You can also execute an existing job with:
+          ```JSON
+          { "operation": "update", "path": "/execute", "value": "True" }
+          { "operation": "update", "path": "/reset_interval", "value": "true" } <-- how do this?
```
scott-taubman, line 64: The existing API is kind of weird and thus not something I would suggest emulating here. And even if we liked the existing pattern, I'm not sure it applies here.  The status update stuff at least is in line with what a PATCH operation implies (making a partial change to the resource), but an adhoc execution of a scheduled job isn't really changing said job, so PATCH doesn't feel like the appropriate method.  See the approach suggested in the original issue write-up.  If there were specific concerns or disagreements about that we can certainly discuss it.

src/app/beer_garden/scheduler.py
```diff
@@ -358,6 +358,28 @@ def remove_job(self, job_id, **kwargs):
...
 
+    def execute_job(self, job_id, reset_interval=False, **kwargs):
+        """Executes the job ad-hoc
+
+        Args:
+            job_id: The job id
+            reset_interval: Whether to set the job's interval begin time to now
+        """
+        job = db.query_unique(Job, id=job_id)
+        self.add_job(
```
scott-taubman, line 369: 2 thoughts here:

* I wonder / worry about any side effects associated with adding a job to the scheduler that isn't persisted in the jobstore. We wouldn't want this persisted in the jobstore, but I'm curious if the schedulers internal state being out of sync with the database record of things is going to cause us headaches at some point.

* Did you explore if it's viable to just call `run_job` instead?  The docstring suggests that `run_job` is intended to be run in a separate process, but it might be worth understanding why that is (I'm assuming some async / blocking related concerns) and see if we can work around those concerns.

src/app/beer_garden/scheduler.py
```diff
@@ -670,6 +692,22 @@ def remove_job(job_id: str) -> None:
...
+    Args:
+        job_id: The Job ID
+        reset_interval: Whether executing should reset the job interval trigger
+
+    Returns:
+        The spawned Request
+    """
+    logger.info('Job executed from routing')
+
+    return db.query_unique(Job, id=job_id)
```
scott-taubman, line 710: Given the way this works, this may be the spot where you need to use `reset_interval` to set the value of `reschedule_on_finish` on the Job.  Maybe.  If it's more complicated than that then it might not be worth supporting the `reset_interval` thing.

src/app/beer_garden/scheduler.py
Jake-Ross: To your other point, since the ad-hoc jobs have a one-time `DateTime` interval, they shouldn't be able to run afterwards. However, there is no debug message (form APscheduler, not BG) saying the job is *removed* like when deleting a normal job through the UI. Since the ad-hoc jobs would all have the same id of `None`, I believe they are being overwritten as more jobs are executed ad-hoc, meaning not much clutter. There may be a way to delete jobs after they're run via APScheduler that I'm not aware of.

Issue #1158, opened 2021-10-14
Support Python version >= 3.6 only


jlrpnbbngtn:
Remove v3,4/v3.5 from the integration tests.

PR #1159, opened 2021-10-14
#1158 modern python only

jlrpnbbngtn:
See commit messages.


scott-taubman (approved):
LGTM.

We should also re-examine the brewtils versions that we're testing against, but that can be done as a separate exercise.

Issue #1149, opened 2021-10-07
Run isort on all python files


scott-taubman:
The python files in the code base do not generally have their imports nicely sorted.  We should fix this by just running isort on all of the python files using the isort config already present in `beer-garden/src/app`.

- [x] #1160
- [x] Update the Makefile to include an isort check during the linting check

jlrpnbbngtn:
Also, ensure isort is run as part of the github checks.

ExpressHermes:
I would like to work on this issue.

scott-taubman:
> I would like to work on this issue.

@ExpressHermes We welcome you to submit a PR for this if you would like to do so.  Our [contributing guide](https://github.com/beer-garden/beer-garden/blob/develop/CONTRIBUTING.md) is pretty generic, but it has the basics for getting started.  Thanks for your interest!

PR #1161, opened 2021-10-14
#1149 - Updated files for PR autochecks

jlrpnbbngtn:
Updated Makefile and friends.


src/app/requirements.txt
```diff
@@ -52,14 +52,10 @@ idna==3.2
     # via requests
 imagesize==1.2.0
     # via sphinx
-importlib-metadata==4.6.4
```
scott-taubman, line 55: It seems unexpected that some things got removed form the requirements.txt when you added the isort requirement.  This seems to stem from the fact that you used python 3.8 to generate this file.  On python 3.7,  these required packages stay.  Since 3.7 is the minimum supported version, we should probably always generate this file using that version.

PR #1166, opened 2021-10-19
#1154 - Add permissions checks to various API endpoints

### LINKED ISSUE #1147: Add authentication / authorization support to the UI
The backend API now supports basic authorization support.  We need to update the UI accordingly so that it allows users to login and then send their authorization token along in cases where auth in enabled in the backend.  The following will be needed:

- [ ] A login page that takes a username and password as input and calls the /api/v1/login endpoint
- [ ] Storage of the token returned from the user login
- [ ] Sending the stored token in the "Authorization: Bearer" header for all calls to the api
- [ ] Redirect users to the login page if an api call returns a 401 response code
- [ ] A logout button that removes the authorization token from storage and then sends the user to the login page

### LINKED ISSUE #1154: Add permission-based authorization checks to API endpoints (round 1)
Now that role and permissions are in place, and we can authenticate users and provide an authorization token, we need to update the API endpoints to actually authorize users and only let them see / operate on data that they have permissions for.  A few notes:

* Each endpoint should generally require a single permission from those available in `beer_garden/api/authorization.py`.  The appropriate permission to use should be fairly obvious in most cases.
* For list type endpoints, data should be filtered down to the objects that the user has permission to see.
* For create, retrieve or update, the result of the permission check is pretty much boolean.

Since there is a lot of changes involved here, we can break things up into multiple issues / PRs.  The scope of this one will cover the following endpoints:

* gardens
* systems
* requests
* users

scott-taubman:
Closes #1154
Closes #1147 

This PR enables permissions based authorization checks on the following endpoints:

* /api/v1/gardens
* /api/v1/gardens/<garden_name>
* /api/v1/systems
* /api/v1/systems/<id>
* /api/v1/requests
* /api/v1/requests/<id>
* /api/v1/users
* /api/v1/users/<username>

With the auth setting disabled, these endpoints continue to operate the way they always have.  When auth is enabled, a valid access token must be supplied in the request "Authorization" header, and the user must have the appropriate permission to view or act on objects handled by those endpoints.

This PR also updates the UI code to work with the new authentication process, allowing users to login and logout appropriately and sending the authorization token along with any API requests when the user is logged in.

# Code Review Notes
The UI changes resulted in a bunch of auto-formatting via prettier.  I separated out the formatting changes from the functional changes, so it's probably useful to look at [this commit](https://github.com/beer-garden/beer-garden/pull/1166/commits/969ec22a94034eba4b9b9785616169d5aaf0d852) to understand what's actually changed with the UI.  Some formatting changes snuck in there too, but it's still much cleaner than the full diff.

# Test Instructions
Some assumptions up front:
* You have a parent and a child garden in your setup.  In the examples they will be named "parent1" and "child1".  Substitute any instances of those in my examples with the names of your gardens.
* You have the roles from the `roles.yaml` example config loaded.
## Auth disabled
* First, leave auth disabled and use beer garden as you normally would. Specifically, access the System, Request, and Garden pages and make sure you're seeing what you expect.
* With auth still disabled, task a system with the "echo" command and something other than the "echo" command.  We're going to use that as one of our permissions-based filtering tests later, so we need an "echo" request and non-"echo" request to exist.
## Auth enabled
**NOTE** With auth enabled, all plugins will fail to start and will not be taskable, since the support for authenticating hasn't been added to brewtils yet.  This is okay and should not impact the things we want to test.
* Enable auth in your garden configuration.
* Now we're going to create some users. In a python shell you can run the following:
  ```python
  from beer_garden.db.mongo.models import Garden, Role, RoleAssignment, User
  from mongoengine import connect

  # substitute bg-parent1 with your parent garden db name and mongodb with your mongodb host
  connect(db="bg-parent1", host="mongodb://mongodb")

  read_only = Role.objects.get(name="read_only")
  operator = Role.objects.get(name="operator")

  p1_read_only_ra1 = RoleAssignment(
      role=read_only, domain={"scope": "Garden", "identifiers": {"name": "parent1"}}
  )
  echo_operator_ra1 = RoleAssignment(
      role=operator, domain={"scope": "System", "identifiers": {"name": "echo"}}
  )

  p1_read_only = User(username="p1-read_only")
  p1_read_only.set_password("password")
  p1_read_only.role_assignments = [p1_read_only_ra1]
  p1_read_only.save()

  echo_operator = User(username="echo-operator")
  echo_operator.set_password("password")
  echo_operator.role_assignments = [echo_operator_ra1]
  echo_operator.save()
  ```
* With all of that setup, try accessing the beer garden UI.  You should be greeted with a login prompt.
* Login as the "p1-read_only".  You should find that on the Systems page, you now only see systems for the parent1 garden.  You should not see any for the child1 garden.
* Try to task a system.  Since we're logged in as a user with read only access, you should get a an error saying access denied when you try to create the request.
* On the Request page, you should see the 2 requests you created earlier, one for "echo" and for some other system.
* Now log out and this time log in as the "echo-operator" user.
* You should now see only the "echo" system on the System page, but for all gardens rather than just parent1.
* On the Request page, you should see only the request for the "echo" system from earlier.  The non-"echo" request should not be visible.
* Attempt to task "echo".  It should let you submit the tasking, though it will stall out or fail due to the earlier note regarding plugins.

Feel free to play around with it some more or do other tests using the API directly.  But the above should be a good slice of what this PR is attempting to achieve.

### DIFF
--- a/src/app/beer_garden/authorization.py
+++ b/src/app/beer_garden/authorization.py
@@ -124,7 +124,7 @@ def user_permitted_objects(
 
 def user_permitted_objects_filter(
     user: "User", model: Type[Document], permission: str
-) -> QCombination:
+) -> Optional[QCombination]:
     """Generates a QCombination that can be used to filter a QuerySet down to the
     objects for which the user has the given permission
 


scott-taubman:
FYI I discovered a bug that I introduced at the last second and apparently hadn't retested to discover.  So if someone had tried to test this and was struggling, that bug (which was preventing plugins from starting) should be fixed now.

Issue #358, opened 2019-11-06
Command metadata


hazmat345:
We should add a metadata field to the command model.

PR #1175, opened 2021-10-29
1132 - Add status_updated_at field to request model

Jake-Ross:
Requires Brewtils [#358](https://github.com/beer-garden/brewtils/pull/367)

Adds the `status_updated_at` field to the mongo model. Included is a unit test that ensures it is not changed on a `save()`.

### DIFF
--- a/src/app/beer_garden/db/mongo/models.py
+++ b/src/app/beer_garden/db/mongo/models.py
@@ -109,6 +109,10 @@ class MongoModel:
     def pre_serialize(self):
         pass
 
+    @property
+    def changed_fields(self):
+        return getattr(self, "_changed_fields", [])
+
 
 # MongoEngine needs all EmbeddedDocuments to be defined before any Documents that
 # reference them. So Parameter must be defined before Command, and choices should be
@@ -312,6 +316,7 @@ class Request(MongoModel, Document):
     command_type = StringField(choices=BrewtilsCommand.COMMAND_TYPES)
     created_at = DateTimeField(default=datetime.datetime.utcnow, required=True)
     updated_at = DateTimeField(default=None, required=True)
+    status_updated_at = DateTimeField(default=datetime.datetime.utcnow, required=True)
     error_class = StringField(required=False)
     has_parent = BooleanField(required=False)
     hidden = BooleanField(required=False)
@@ -331,6 +336,7 @@ class Request(MongoModel, Document):
             {"name": "status_index", "fields": ["status"]},
             {"name": "created_at_index", "fields": ["created_at"]},
             {"name": "updated_at_index", "fields": ["updated_at"]},
+            {"name": "status_updated_at_index", "fields": ["status_updated_at"]},
             {"name": "comment_index", "fields": ["comment"]},
             {"name": "parent_ref_index", "fields": ["parent"]},
             {"name": "parent_index", "fields": ["has_parent"]},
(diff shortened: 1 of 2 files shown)


scott-taubman (changes requested):
See my comments on the brewtils side of things.  Ultimately, you want to actually set the status_updated_at field on the model here when appropriate.  The place you'd generally do that is  in a pre-save hook, most of which we've currently go living in the `clean` method.  You can just add another check there to see if `status` is among the changed fields (`self._changed_fields`) and if it is, set that `status_updated_at` time to now.

Jake-Ross:
> You can just add another check there to see if `status` is among the changed fields (`self._changed_fields`) and if it is, set that `status_updated_at` time to now.

This (`self._changed_fields`) looks like a mongo thing, but I can't find a way to access it within the `clean()` method. Latest push just checks to see if the status changed manually.

src/app/beer_garden/db/mongo/models.py
```diff
@@ -474,12 +476,24 @@ def clean(self):
                 f"consistent with has_parent value of {self.has_parent}"
             )
 
+        try:
```
scott-taubman, line 479: I'm going to guess that the problem you ran into with trying to use `_changed_fields` is that it doesn't get set (or even exist) if we're dealing with a new object.  You already have the default value getting set for `status_updated_at` on new Requests, so we shouldn't have to worry about that case.  Which means I think you can just something like this:

```python
# In the MongoModel base class that Request builds off of:
@property
def changed_fields(self):
    return getattr(self, "_changed_fields", [])

# And then in place of this try / except block:
if "status" in self.changed_fields:                                                                                                                                                                       
   self.status_updated_at = datetime.datetime.utcnow()
```

That allows us to avoid the extra database lookup that you've got in there now.

scott-taubman (changes requested):

src/app/beer_garden/db/mongo/models.py
Jake-Ross: Changes are made. Had to force-push a new commit, since I learned the hard way that `git pull --rebase` does not do what I thought it did. 

There is some strange behavior, however - see the unit tests. If the first `save()` is removed from the second unit test, where the `status` field is actually changed, the unit test fails. `_changed_fields` seems to be accurate after the first save, however.

scott-taubman: I think this is the expected behavior based on the way the fixture is setup.  The fixture doesn't actually call `save()`, so in the test you're dealing with model instance that hasn't actually been saved to the DB yet.  That means the first time you call `save()` in the test, you're going down the path where `_created` is `True` and `_changed_fields` is not set.  Basically, without that first save, you're not really changing the `status` field, you're setting its initial value.

All that is to say, give the fixture you're working with, your test is doing the right thing.

src/app/beer_garden/db/mongo/models.py
```diff
@@ -474,12 +480,16 @@ def clean(self):
...
+        if "status" in self.changed_fields:
+            self.status_updated_at = datetime.datetime.utcnow()
+
     def clean_update(self):
         """Ensure that the update would not result in an illegal status transition"""
         # Get the original status
         old_status = Request.objects.get(id=self.id).status
 
         if self.status != old_status:
+            self.status_updated_at = datetime.datetime.now()
```
scott-taubman, line 492: Sorry I missed this in the last couple rounds of review, but this should be `utcnow()` instead of `now()`.

I think last time I checked I convinced myself that this `clean_update` method will never get called, but it's still probably better to make the fix.

scott-taubman, line 492:
```suggestion
            self.status_updated_at = datetime.datetime.utcnow()
```

scott-taubman: Sorry I missed this in the last couple rounds of review.

I think last time I checked I convinced myself that this `clean_update` method will never get called, but it's still probably better to make the fix.

scott-taubman (approved):

PR #1181, opened 2021-11-09
#1062 - Run now button

### LINKED ISSUE #1062: Scheduled jobs "run now" button
It would be useful to have a button the job view page that would cause the job to be executed immediately. Bonus points if the user also gets to pick if that execution resets the interval.

This issue covers the frontend portion of the work.  This requires that a new API endpoint be created first.  See issue #1066.

jlrpnbbngtn:
Closes #1062 

#### Manual Testing
1. Schedule a job from the *Scheduler* page. It doesn't matter at this time what the trigger type of the job is.
2. From the *Request Scheduler* page, click the link for the job (i.e., the job's name ) in the *Job Name* column of the table on that page.
3. The resulting page has a *Run Now* button. Click it. Note that an information message is immediately shown in BG's logging indicating the job's been run. Note that a request with the time the button was clicked shows in the *Request* page and that the table entry reflects the correct job.
4. It's assumed that at some point in the future the "run now" API endpoint will accept a value indicating whether the interval should be reset when scheduling an ad hoc run of a job that has an interval trigger type. The code for this PR anticipates that scenario, but it is turned off by default. To exercise the functionality:
    - Temporarily change line 124 of `beer-garden/src/ui/src/js/controllers/job_view.js` from `return false;` to `return true;`. (Alternately, change the line to what the comment in that function says it should eventually be. In that case, the job must have an *interval* trigger type in order to show the popup).
    - Repeat step 3. above. Note that a console message indicating whether the interval should be reset is shown and reflects whether the *Yes* or *No* button were pushed in the resulting popup. Note that an ad hoc job is not scheduled if the *Cancel* button is chosen instead.
    - Change line 124 back.

### DIFF
--- a/src/app/beer_garden/scheduler.py
+++ b/src/app/beer_garden/scheduler.py
@@ -367,7 +367,7 @@ class MixedScheduler(object):
         job = db.query_unique(Job, id=job_id)
         self.add_job(
             run_job,
-            trigger=DateTrigger(datetime.now(), timezone="UTC"),
+            trigger=DateTrigger(datetime.utcnow(), timezone="UTC"),
             trigger_type="date",
             coalesce=job.coalesce,
             kwargs={"job_id": job.id, "request_template": job.request_template},
--- a/src/ui/package.json
+++ b/src/ui/package.json
@@ -40,7 +40,7 @@
     "objectpath": "^2.0.0",
     "startbootstrap-sb-admin-2": "^3.3.7",
     "swagger-ui-dist": "^3.25.0",
-    "tv4": "~1.0.15",
+    "tv4": "~1.3.0",
     "ui-select": "^0.19.8"
   },
   "devDependencies": {
--- a/src/ui/src/js/controllers/job_view.js
+++ b/src/ui/src/js/controllers/job_view.js
@@ -102,7 +102,9 @@ export function jobViewController(
 
     JobService.runAdHocJob(jobId, $scope.resetTheInterval).then(
       function (response) {
-        console.log(`Ad hoc run of ID ${jobId}; reset interval: ${resettingInterval}`);
+        console.log(
+          `Ad hoc run of ID ${jobId}; reset interval: ${resettingInterval}`
+        );
         $state.go("base.jobs");
       },
       function (response) {
@@ -136,7 +138,7 @@ export function jobViewController(
       let popupInstance = $uibModal.open({
         animation: true,
         template: modalTemplate,
-        controller: "JobRunNowModalController"
+        controller: "JobRunNowModalController",
       });
 
       popupInstance.result.then(
@@ -148,8 +150,8 @@ export function jobViewController(
       );
     } else {
       runAdHoc(jobId);
-    }    
-  }
+    }
+  };
 
   loadJob();
 }
@@ -167,6 +169,6 @@ export function jobRunNowModalController($scope, $uibModalInstance) {
   };
 
   $scope.cancelRunNow = function () {
-    $uibModalInstance.dismiss("cancel")
+    $uibModalInstance.dismiss("cancel");
   };
 }
--- a/src/ui/src/js/controllers/job_view.js
+++ b/src/ui/src/js/controllers/job_view.js
@@ -121,7 +121,7 @@ export function jobViewController(
    * when the BG API can make sense of a reset interval value
    */
   function isIntervalTrigger(triggerType) {
-    return true;
+    return false;
   }
 
   /*
--- a/src/ui/src/js/controllers/job_view.js
+++ b/src/ui/src/js/controllers/job_view.js
@@ -121,7 +121,7 @@ export function jobViewController(
    * when the BG API can make sense of a reset interval value
    */
   function isIntervalTrigger(triggerType) {
-    return false;
+    return true;
   }
 
   /*
--- a/src/ui/src/js/controllers/job_view.js
+++ b/src/ui/src/js/controllers/job_view.js
@@ -1,10 +1,12 @@
 import { formatDate, formatJsonDisplay } from "../services/utility_service.js";
+import modalTemplate from "../../templates/reset_interval_modal.html";
 
 jobViewController.$inject = [
   "$scope",
   "$rootScope",
   "$state",
   "$stateParams",
+  "$uibModal",
   "JobService",
 ];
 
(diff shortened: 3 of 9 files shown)


src/app/beer_garden/scheduler.py
```diff
@@ -367,7 +367,7 @@ def execute_job(self, job_id, reset_interval=False, **kwargs):
         job = db.query_unique(Job, id=job_id)
         self.add_job(
             run_job,
-            trigger=DateTrigger(datetime.now(), timezone="UTC"),
+            trigger=DateTrigger(datetime.utcnow(), timezone="UTC"),
```
scott-taubman, line 370: I assume this works as intended, but unless there's some compelling reason not to which should just remove the trigger (and presumably trigger_type) as the [official documentation](https://apscheduler.readthedocs.io/en/3.x/userguide.html#adding-jobs) suggests.  It specifically says:

>**Tip:**  To run a job immediately, omit trigger argument when adding the job.

src/ui/src/js/controllers/job_view.js
```diff
@@ -90,5 +94,81 @@ export default function jobViewController(
     loadJob();
   });
 
+  /*
+   * Execute the job service's "run ad hoc job" functionality.
+   */
+  function runAdHoc(jobId) {
+    const resettingInterval = $scope.resetTheInterval;
+
+    JobService.runAdHocJob(jobId, $scope.resetTheInterval).then(
```
scott-taubman, line 103: It probably doesn't matter, but you create the `resttingInteverval` constant above and then don't actually use it here.

src/ui/src/templates/reset_interval_modal.html
```diff
@@ -0,0 +1,18 @@
+<div class="modal-header">
+    <h4>Reset the Job Interval?</h4>
```
scott-taubman, line 2: It would be good to provide some additional wording here, since I'm not sure most people will understand what this is asking.  Something like:

**Reset the Job Interval?**
Select "Yes" to update the next run time of this job based on the time of this current run.
Select "No" to keep the job's existing next run time.

That's not fantastic so feel free to workshop it a bit.  The point is that I think we need something to convey what this option is about.

src/ui/package.json
```diff
@@ -40,7 +40,7 @@
     "objectpath": "^2.0.0",
     "startbootstrap-sb-admin-2": "^3.3.7",
     "swagger-ui-dist": "^3.25.0",
-    "tv4": "~1.0.15",
+    "tv4": "~1.3.0",
```
scott-taubman, line 43: What drove this update?

scott-taubman (changes requested):
Tests pass and it looks reasonable to me.  Made mostly minor comments that should be easy to address.

src/ui/src/js/controllers/job_view.js
jlrpnbbngtn: The function below closes over the environment established here.

src/ui/package.json
jlrpnbbngtn: The LTS version of *npm* is 6.14.15. Using that version of *npm*, the library at  version 1.0.15 throws up a number of warnings that don't exist at a newer version. The author doesn't appear to publish a change log on his github, but I'm assuming a minor version bump will not break its API.

src/ui/src/templates/reset_interval_modal.html
jlrpnbbngtn: Yes, I was waiting for your feedback on this after you'd mentioned the issue originally (but I didn't take notes at that time).

jlrpnbbngtn: Made the update with some minor wordsmithing.

src/app/beer_garden/scheduler.py
jlrpnbbngtn: Alas, cannot change at this time. The `add_job` function specifically checks that the `trigger` argument is not `None`. Though I could change that function too, I'm not confident that such a change wouldn't break something else unexpectedly.

src/ui/src/templates/reset_interval_modal.html
scott-taubman: You somehow came up with something more wordy than I did :rofl:.

My only objection to what you have now is the use of the phrase "original job", which even gets special emphasis in one instance.  This seems very confusing, since from the user's perspective there is just the job that they're trying to interact with.  The fact that they are technically spawning a new job on the backend and that it's the next run time on that "original job" that gets impacted is all just implementation details, so I think referencing the "original job" is confusing.

I'm happy to save any further wordsmithing until we come back and actually enable that pop up.  Since it won't show up yet anyway we can probably just proceed.

scott-taubman (approved):
Made further comments regarding the pop-up wording, but I think we can continue to refine that later.  I'm okay merging this as is.

src/ui/src/js/run.js
```diff
@@ -178,20 +178,18 @@ export default function appRun(
...
-      );
-      loginModal.closed.then(
-          () => {
-            loginModal = undefined;
-          }
+        () => {
+          $rootScope.changeUser(TokenService.getToken());
+          location.reload();
+        },
+        _.noop // Prevents annoying console log messages
```
kilo-alpha-yankee, line 185: I like the code clean up here.

src/ui/src/templates/reset_interval_modal.html
```diff
@@ -0,0 +1,30 @@
...
+              class="btn btn-secondary"
+              value="No"
+              ng-click="setIntervalReset(false)" />
+       <input type="button"
+              class="btn btn-danger"
+              value="Cancel"
+              ng-click="cancelRunNow()" />
+       </div>
+</div>
+
```
kilo-alpha-yankee, line 30: The diff is claiming there's no newline at the end of this file. Is that an eslint rule you have?

kilo-alpha-yankee: Ah, never mind it looks like that's what you added with the change

src/ui/src/js/controllers/job_view.js
kilo-alpha-yankee: I agree it seems weird that you would set the constant and not pass it to the function. I wonder if that pattern makes the const act like a variable.

src/ui/src/templates/reset_interval_modal.html
jlrpnbbngtn: No, I'd be interested to know what you think it should say. And then I'll make it say that :-)

kilo-alpha-yankee (approved):

src/ui/src/templates/reset_interval_modal.html
jlrpnbbngtn: Further minor wording changes made.

PR #1184, opened 2021-11-10
#1182 - Generate requirements.txt from setup.py

### LINKED ISSUE #1182: Remedy release process deficiencies
Two specific issues were encountered while preparing the last release:

1. Beer-garden's `setup.py` has dependency versions that are out of sync with the `requirements.{txt,in}` files, with the potential to introduce difficult to assess bugs in the RPM creation.
2. The version of `wrapt` that's current on PyPI will not work with beer-garden at this time. The library version should be pinned to an older version in all the appropriate places.

The best fix may be to address these two issues in the most direct way. But it might be better to figure out a mechanism by which only one of `setup.py` or `requirements.txt` will be the authoritative source of versions and have the other automatically sync with it.

scott-taubman:
Closes #1182 

This PR slightly changes the way we manage the project dependencies.  It does the following:

* Updated the version pinning in the `setup.py` to match what's currently in the `requirements.in`.
* Moved the test and dev dependencies from the `requirements.in` to a `requirements-dev.in`
* Removed `requirements.in`
* Added a `deps-compile` command to the Makefile that generates the `requirements.txt` and a new `requirements-dev.txt` from the `setup.py` and `requirements-dev.in` respectively.
* Updated the `deps` command in the Makefile so that it now installs both the `requirements.txt` and `requirements-dev.txt` dependencies.

The net effect of all of this is that we now only need to manage the dependencies for running beer garden in the setup.py.  Any dependencies specific to development and testing are managed separately int he `requirements-dev.in`.

## Test Instructions
* Remove the `requirements.txt` and `requirements-dev.txt`
* From the `/src/app` directory, run `make deps-compile`
* This should recreate the `requirements.txt` and `requirements-dev.txt` files for you.
* Running `make deps` should do a pip install using those two files for you.


scott-taubman:
Quick note that the `wrapt` issue also mentioned in #1182 was on the brewtils side and it looks like we committed the fix to that when we were fixing the last release.  So I don't believe there's anything that needs to happen regarding that here.

src/app/setup.py
```diff
@@ -28,19 +28,21 @@ def find_version(version_file):
...
-        "pyrabbit2==1.0.7",
+        "prometheus-client<1",
+        "pyyaml<6",
+        "pyrabbit2<2",
         "pytz<2021",
         "ruamel.yaml<0.17",
         "stomp.py<6.2.0",
-        "tornado==6.0.4",
+        "tornado<7",
         "yapconf>=0.3.7",
```
jlrpnbbngtn, line 46: I think you based this off an old `requirements.txt`. https://github.com/beer-garden/beer-garden/pull/1167 discusses that the `wheel` package needs to be explicitly installed in Python 3.7+.

jlrpnbbngtn (approved):

PR #1189, opened 2021-11-15
#1187 - Improve UI handling of refresh tokens

### LINKED ISSUE #1187: Improve UI refresh token handling
There are still several issues with the UIs handling of refresh tokens.  Known issues include:

* Multiple requests to the API that would result in a 401 end up stepping on each other, with each initiating a refresh request when a refresh token is present.
* Refresh requests with an invalid or expired token are not properly handled and result in the stored token values being set, despite the fact that the response did not include updated tokens.

scott-taubman:
Closes #1187

This PR attempts to fix the UI's refresh token handling.  Specifically it fixes a couple of behaviors:

* It's now possible to open new tabs or close and re-open a tab without having your session expire.
* Supplying an invalid token to the `/refresh` endpoint is handled more gracefully (shouldn't result in an unhandled rejection error in the console).

# Test Instructions
* Enable auth and login via the UI.
* Close the tab and then navigate back to beer-garden.  You should still be logged in.
* Ctrl click on something to open it in a new tab.  It should open and not prompt you to login again.  Try navigating around on the original tab, refreshing the page, etc.  You should still not be prompted to login.
* Click "Log Out" on one and then navigating around or refresh in another.  You should be prompted to login since your session has been invalidated across all tabs.

### DIFF
--- a/src/ui/src/js/configs/http_interceptor.js
+++ b/src/ui/src/js/configs/http_interceptor.js
@@ -30,6 +30,8 @@ authInterceptorService.$inject = ["$q", "$injector"];
  * @return {Object}                                  Interceptor object
  */
 export function authInterceptorService($q, $injector) {
+  let inFlightAuthRequest = null;
+
   return {
     responseError: (rejection) => {
       // 401 means 'needs authentication'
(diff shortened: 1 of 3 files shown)


src/ui/src/js/run.js
```diff
@@ -115,17 +115,22 @@ export default function appRun(
     // SystemService.loadSystems();
 
     $rootScope.loadUser(token).then(() => {
-      $state.reload();
```
scott-taubman, line 118: This wasn't strictly related to this PR, but as far as I could tell this wasn't a valid statement and only resulted in an error getting spit out on the console.  I got annoyed at seeing the error so I just took this line out.

src/ui/src/js/services/token_service.js
```diff
@@ -11,14 +11,14 @@ tokenService.$inject = ["$http", "localStorageService"];
 export default function tokenService($http, localStorageService) {
   const service = {
     getToken: () => {
-      return localStorageService.get("token", "sessionStorage");
+      return localStorageService.get("token");
```
scott-taubman, line 14: There didn't seem to be a way to gracefully handle multiple tabs / windows while storing the token in session storage.  Moving it to the local storage has the effect of all tabs / windows essentially sharing the same login session.

src/ui/src/js/configs/http_interceptor.js
```diff
@@ -39,33 +41,36 @@ export function authInterceptorService($q, $injector) {
...
         // expired but there's a refresh token in storage. We use the refresh
         // token to get a new access token then re-attempt the original request.
         let refreshToken = TokenService.getRefresh();
         if (refreshToken) {
-          return TokenService.doRefresh(refreshToken).then(
-            (response) => {
-              // Set the Authorization header to the updated default
-              rejection.config.headers.Authorization =
-                $http.defaults.headers.common.Authorization;
+          if (!inFlightAuthRequest) {
```
scott-taubman, line 51: This whole `inFlightAuthRequest` thing is to avoid attempting to refresh multiple times when there are multiple calls to the API on a single page and all of them result in a 401.  Without this, each call would result in its own refresh attempt, with them all basically stepping on each other because each refresh invalidates the preceding one.

west270 (approved):

PR #1190, opened 2021-11-15
#1178 - Improve job import API validation

### LINKED ISSUE #1178: Job Import API lacks validation
It appears that the job import endpoint will create jobs for possibly any valid json, even if it's not a valid job definition.  A bit more research needed into the exact cases where this happens and more details TBD.

scott-taubman:
Closes #1178

This PR fixes the lack of validation in the job import API such that it no longer creates Job entries for definitions that fail to pass validation.

It does this largely by replacing the direct interactions with pymongo in the `create_jobs` function with a simple wrapper around the `create_job` function, which saves each entry through the mongoengine model.

# Test Instructions
* Use the import API to import the following, which includes both a valid job definition and an invalid one:
  ```json
  [
    {
      "name": "well_formed_job",
      "misfire_grace_time": 3,
      "trigger": {
        "timezone": "utc",
        "run_date": 1700000000000
      },
      "coalesce": true,
      "trigger_type": "date",
      "timeout": 30,
      "max_instances": 3,
      "request_template": {
        "metadata": {
          "request": "stuff"
        },
        "command": "speak",
        "parameters": {
          "message": "hey!"
        },
        "instance_name": "default",
        "output_type": "STRING",
        "command_type": "ACTION",
        "namespace": "ns",
        "system": "system",
        "system_version": "1.0.0",
        "comment": "hi!"
      }
    },
    {
      "name": "job_with_a_bunch_of_missing_fields"
    }
  ]
  ```
* The response should contain only a single job ID.  Note that the response format now matches what the swagger documentation suggests it's going to be.
* If you check the Jobs (Through the API, UI, or database. Whatever is easiest for you), there should be only one new Job after your import, not 2.
* If you want to sanity check, do the same test on the develop branch.  You'll end up with 2 jobs created, one of which has only the name field populated, which is what this PR is fixing.

### DIFF
--- a/src/app/beer_garden/scheduler.py
+++ b/src/app/beer_garden/scheduler.py
@@ -5,27 +5,18 @@ The schedule service is responsible for:
 * CRUD operations of `Job` operations
 * Triggering `Job` based `Requests`
 """
-import json
 import logging
 import threading
 from datetime import datetime, timedelta
-from operator import attrgetter, methodcaller
 from os.path import isdir
 from typing import Dict, List, Optional
 
-import mongoengine
-import pymongo
-import pymongo.database
 from apscheduler.schedulers.background import BackgroundScheduler
 from apscheduler.triggers.interval import IntervalTrigger as APInterval
+from brewtils.errors import ModelValidationError
 from brewtils.models import DateTrigger, Event, Events, FileTrigger, Job, Operation
-from brewtils.schema_parser import SchemaParser
-from bson import ObjectId, json_util
+from mongoengine import ValidationError
 from pathtools.patterns import match_any_paths
-from pymongo.client_session import ClientSession
-from pymongo.collection import Collection
-from pymongo.cursor import Cursor
-from pymongo.results import InsertManyResult
 from watchdog.events import (
     EVENT_TYPE_CREATED,
     EVENT_TYPE_DELETED,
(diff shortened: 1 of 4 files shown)


jlrpnbbngtn (dismissed):
LGTM

src/app/beer_garden/scheduler.py
```diff
@@ -556,71 +546,29 @@ def create_job(job: Job) -> Job:
...
+        try:
+            created.append(create_job(job))
+        except (ModelValidationError, ValidationError) as exc:
+            rejected.append((job, str(exc)))
 
-    # but we're only interested in the `id` field
-    return list(
-        map(attrgetter("id"), map(publish_the_job_created, jobs_created_cursor))
-    )
+    return {"created": created, "rejected": rejected}
```
jlrpnbbngtn, line 571: This causes the wrapper that easy_client uses to enter into an infinite loop with an eventual *RecursionError* thrown. The function *serialize* has a warning in its docstring that

> This is potentially recursive

and maybe that wasn't such a good idea.

At any rate, and without jumping deeply into the code, I'm guessing the  brewtils schema needs to be updated and maybe this will go away...

jlrpnbbngtn (changes requested):

src/app/beer_garden/scheduler.py
scott-taubman: Very interesting.  I wouldn't expect there to be any interaction between this function and the easy_client, but I will work on reproducing what you found and get to the bottom of it.

jlrpnbbngtn (approved):

Issue #369, opened 2019-12-02
Command attributes as parameters


hazmat345:
It's a fairly common pattern to be able to represent the same data in different formats based on what the user wants. It would be useful to be able to mark a `@command` as having an output type that's basically a parameter. This would need to be a deliberate choice by the plugin developer as the plugin would need to be written in a way that supports it (inspecting the `output_type` and constructing the output appropriately).

It would also be nice to be able to set the command type (INFO vs ACTION) as a parameter.

TheBurchLog:
A potential way to implement this could look like the following

```python
@command(output_types=["JSON", "HTML"], output_type="JSON")
```

hazmat345:
OBE.

PR #1199, opened 2021-11-22
#1173 - Add 'reset_interval' option to ad-hoc executions of jobs

### LINKED ISSUE #1173: Add 'reset_interval' option to ad-hoc executions of jobs
Related to #1066. A `reset_interval` argument was not included in the API, but should be included as either a URL Parameter (e.g. `/api/v1/jobs/<id>/execute?reset_interval=true`) or as JSON information in the body of the POST request. 

Jobs with triggers that are not `interval`s should not be affected.

Jake-Ross:
Closes #1173

Does not require Brewtils [#369](https://github.com/beer-garden/brewtils/pull/375), but this PR is useful for testing.

# To Test
- Create a job with an `interval` trigger of 30 seconds.
- Wait 15 seconds, then execute the job using the `execute_job` function from the Brewtils [#369](https://github.com/beer-garden/brewtils/pull/375) PR.
- Ensure the job is executed at the expected times - one time when the job is manually executed, then again 30 seconds later (45 seconds after the job was initially created). You can see execution times in Beergarden's logs in the console.
- Create a job without an `interval` trigger, ensure execution returns a `400` when `reset_interval` is True.

### DIFF
--- a/src/app/beer_garden/api/http/handlers/v1/job.py
+++ b/src/app/beer_garden/api/http/handlers/v1/job.py
@@ -367,7 +367,9 @@ class JobExecutionAPI(AuthorizationHandler):
         _ = self.get_or_raise(Job, JOB_CREATE, id=job_id)
 
         reset_interval = (
-            True if self.get_argument("reset_interval", "False").lower() == "true" else False
+            True 
+            if self.get_argument("reset_interval", "False").lower() == "true"
+            else False
         )
 
         try:
--- a/src/app/beer_garden/api/http/handlers/v1/job.py
+++ b/src/app/beer_garden/api/http/handlers/v1/job.py
@@ -6,6 +6,7 @@ from brewtils.schemas import JobExportInputSchema, JobSchema
 from mongoengine.errors import ValidationError
 
 from beer_garden.api.authorization import Permissions
+from beer_garden.api.http.exceptions import BadRequest
 from beer_garden.api.http.handlers import AuthorizationHandler
 from beer_garden.db.mongo.models import Job
 from beer_garden.scheduler import create_jobs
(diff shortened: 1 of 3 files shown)


src/app/beer_garden/api/http/handlers/v1/job.py
```diff
@@ -365,10 +366,18 @@ async def post(self, job_id):
...
+        )
+
         try:
-            await self.client(Operation(operation_type="JOB_EXECUTE", args=[job_id]))
+            await self.client(
+                Operation(operation_type="JOB_EXECUTE", args=[job_id, reset_interval])
+            )
         except ValidationError:
             raise NotFoundError
+        except ModelValidationError:
```
scott-taubman, line 379: You should capture the message here and use it as the `reason` on the `BadRequest`.  Like:

```python
except ModelValidationError as exc:
    raise BadRequest(reason=f"{exc}")
```

The way the `write_error` method works in the `BaseHandler` makes this not useful yet, but once we get around to fixing that handling in the `BaseHandler`, having the reason listed here will make it so that a useful message is returned with the response, as opposed to just "Bad request".

src/app/beer_garden/scheduler.py
```diff
@@ -647,7 +653,13 @@ def execute_job(job_id: str) -> Job:
     Returns:
         The spawned Request
     """
-    return db.query_unique(Job, id=job_id, raise_missing=True)
+    job = db.query_unique(Job, id=job_id, raise_missing=True)
+    job.reset_interval = reset_interval
+
+    if reset_interval and job.trigger_type != "interval":
+        raise ModelValidationError()
```
scott-taubman, line 660: As alluded to in another comment, you should have a message here with this exception.

```python
raise ModelValidationError("reset_interval can only be used with trigger type of interval")
```

scott-taubman (changes requested):
Some super minor changes to be made, but otherwise looks good.

src/app/beer_garden/scheduler.py
scott-taubman: As a side note, `ModelValidationError` doesn't quite feel like the right exception here, but I don't feel super strongly about it.  Just a regular `ValueError` would probably have been my pick.  For reference, `ValueError` is described as:

> Raised when an operation or function receives an argument that has the right type but an inappropriate value.

That feels like it describes this situation.

src/app/beer_garden/api/http/handlers/v1/job.py
```diff
@@ -365,10 +366,18 @@ async def post(self, job_id):
         """
         _ = self.get_or_raise(Job, JOB_CREATE, id=job_id)
 
+        reset_interval = (
+            True if self.get_argument("reset_interval", "False") == "True" else False
```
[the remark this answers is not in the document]
Jake-Ross: Good call. It was surprising to see that the `get_argument` method doesn't convert to boolean automatically.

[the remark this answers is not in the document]
jlrpnbbngtn: The `if` and `else` are superfluous. `self.get_argument("reset_interval", "False").lower() == "true"` is equivalent.

PR #1204, opened 2021-12-01
#1197 - Add a debian based docker image

### LINKED ISSUE #1197: Create a non-alpine beer-garden docker image
Alpine as the base image for python projects is problematic for various reasons. We should make a non-alpine version available. Using the official python slim image, which is debian based, is probably a good choice.

scott-taubman:
Closes #1197 

This PR adds a Dockerfile for a debian (python:3.10-slim) based image that will get built and published as part of the standard release process.

The existing alpine image places and looks for files in a bunch of non-standard places, which I wasn't really keen on replicating with this debian image.  Because of this, the alpine and debian images are not compatible in terms of how you would instruct someone to use them. That said, there currently are no official instructions on how to use the alpine-based image.  You can see on [dockerhub](https://hub.docker.com/r/bgio/beer-garden) that there is no description for the image.

I went ahead and created a set of instructions for the debian image (see the README.md in this PR). My thought is that we'd do the following:

* Publish the debian-based image in addition to the alpine one, with the same tags, just with a `-debian` suffix.
* The tags that currently get created (e.g. `3.9.0`, `3`, `latest`) will continue to be based off of the alpine image for now.
* Add alpine specific tags that point to the same image as those current tags (e.g. `3.9.0-alpine`, `3-alpine`).
* Give warning in the release notes that in a coming release, the `debian` image will become the officially supported one and the vanilla tags (e.g. `3.10.0`, `3`, `latest`) will switch to being based off it rather than the `alpine` image.  Anyone who needs to continue using the legacy image can switch to using the `-alpine` tags immediately and will see no disruption once the change is made.
* At the point that we make that switch, publish the `README.md` in this PR as the description on the official image on dockerhub.

## Testing Instructions
* from within the `src/app` directory, run the following:
  ```shell
  make docker-build VERSION=3.8.0
  ```
* This should build both the alpine image and the new debian based one.  You can check the images built like so:
  ```shell
  $ docker image ls | grep 'bgio/beer-garden'
  bgio/beer-garden                             3-debian                fae574b2b896   About an hour ago   185MB
  bgio/beer-garden                             3.8.0-debian            fae574b2b896   About an hour ago   185MB
  bgio/beer-garden                             3                       ce8af8f3efa4   About an hour ago   78.3MB
  bgio/beer-garden                             3-alpine                ce8af8f3efa4   About an hour ago   78.3MB
  bgio/beer-garden                             3.8.0                   ce8af8f3efa4   About an hour ago   78.3MB
  bgio/beer-garden                             3.8.0-alpine            ce8af8f3efa4   About an hour ago   78.3MB
  bgio/beer-garden                             latest                  ce8af8f3efa4   About an hour ago   78.3MB
  ```
* Now try following the [README.md](https://github.com/scott-taubman/beer-garden/blob/issue/1197-debian-docker-image/src/app/docker/README.md) instructions regarding how to run the container.  There a whole lot of environmental dependencies, so be sure to adapt the commands listed to your environment (you may, for instance, need to specify a network for your container).


src/app/example_configs/app-logging.yaml
```diff
@@ -16,7 +16,7 @@ handlers:
     encoding: utf8
     formatter: simple
     maxBytes: 10485760
-    filename: "beer-garden.log"
+    filename: "/var/log/beer-garden/beer-garden.log"
```
jlrpnbbngtn, line 19: Hard coding this to `/var/log` removes the ability to run BG from the command line using this file without root access. I'd prefer if the file could be used for both scenarios without editing. Perhaps, `tmp` instead?

jlrpnbbngtn (changes requested):
See comments

src/app/docker/README.md
```diff
@@ -0,0 +1,128 @@
...
+that can be used as a reference for starting out.
+
+## Local plugins
+
+By default, beer-garden will look in `/var/lib/beer-garden/plugins`. If you wish
+to run local plugins, you should mount the directory containing the plugins on
+the host to that directory on the container.
+
+> **NOTE**: Currently local plugins can only be configured using the
+> configuration file method of configuration described above. Plugins will not
```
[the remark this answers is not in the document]
scott-taubman: Updated.

src/app/example_configs/app-logging.yaml
scott-taubman: Yeah, this felt bad when I did it, and then I guess I forgot to go back and figure out what I really wanted to do here.  I reverted this back to what it was and instead generate a new app-logging.yaml during the docker build which is configured to output just to stdout, which seems like the right default config in the context of a docker container.

PR #1208, opened 2021-12-06
#1207 - Add global permission check to forward endpoint

### LINKED ISSUE #1207: Add authorization checks to the forward endpoint
The forward endpoint needs to support authorization.  Since all manner of operations can be executed on remote gardens via the forward endpoint and the endpoint is really only intended to be used for garden to garden communications, it is easiest to treat access to the endpoint as an all or nothing proposition.  Either you can an entity can forward requests to the garden or it cannot.  More granular access control can be considered at a later time.

scott-taubman:
Closes #1207

This PR adds an authorization check to the `/api/v1/forward` endpoint.

### Assumptions
* Your `superuser` role has the "event:forward" permission. Use the `example_configs/roles.yaml` or update your own config file if you are using some custom role config.

### Test with auth disabled
  ```shell
  curl -si -X POST -H "Content-Type: application/json" --data '{"operation_type": "GARDEN_READ", "args": ["somegarden"]}' "http://localhost:8080/api/v1/forward" | head -1
  # HTTP/1.1 204 No Content
  ```
### Test with auth enabled:
  * Create users with and without access:
    ```python
    from mongoengine import connect
    from beer_garden.db.mongo.models import Role, RoleAssignment, User

    # modify db and host parameters are appropriate for your environment
    connect(db="bg-parent1", host="mongodb://mongodb")

    read_only = Role.objects.get(name="read_only")
    role_assignment = RoleAssignment(role=read_only, domain={"scope": "Global"})
    user = User(username="ro_user", role_assignments=[role_assignment])
    user.set_password("password")
    user.save()

    superuser = Role.objects.get(name="superuser")
    role_assignment = RoleAssignment(role=superuser, domain={"scope": "Global"})
    user = User(username="su_user", role_assignments=[role_assignment])
    user.set_password("password")
    user.save()
    ```
  * Login and hit the forward endpoint:
    ```shell
    # Test for user without access: ro_user
    curl -H "Content-Type: application/json" --data '{"username": "ro_user", "password": "password"}' "http://localhost:8080/api/v1/token" | jq -r '.access' > /tmp/ro_user.token
    curl -si -H "Content-Type: application/json" -H "Authorization: Bearer `cat /tmp/ro_user.token`" --data '{"operation_type": "GARDEN_READ", "args": ["somegarden"]}' "http://localhost:8080/api/v1/forward" | head -1
    # HTTP/1.1 403 Unauthorized

    # Test for user with access: su_user
    curl -H "Content-Type: application/json" --data '{"username": "su_user", "password": "password"}' "http://localhost:8080/api/v1/token" | jq -r '.access' > /tmp/su_user.token
    curl -si -H "Content-Type: application/json" -H "Authorization: Bearer `cat /tmp/su_user.token`" --data '{"operation_type": "GARDEN_READ", "args": ["somegarden"]}' "http://localhost:8080/api/v1/forward" | head -1
    # HTTP/1.1 204 No Content
    ```

### DIFF
--- a/src/app/beer_garden/api/http/handlers/v1/forward.py
+++ b/src/app/beer_garden/api/http/handlers/v1/forward.py
@@ -4,12 +4,13 @@ import asyncio
 from brewtils.errors import TimeoutExceededError
 from brewtils.schema_parser import SchemaParser
 
-from beer_garden.api.http.base_handler import BaseHandler
+from beer_garden.api.authorization import Permissions
+from beer_garden.api.http.handlers import AuthorizationHandler
+
+EVENT_FORWARD = Permissions.EVENT_FORWARD.value
 
 
-class ForwardAPI(BaseHandler):
-
-    # @Todo Create new Persmission
+class ForwardAPI(AuthorizationHandler):
     async def post(self):
         """
         ---
@@ -37,7 +38,7 @@ class ForwardAPI(BaseHandler):
             schema:
                 $ref: '#/definitions/Forward'
         responses:
-          200:
+          204:
             description: Forward Request Accepted
           400:
             $ref: '#/definitions/400Error'
@@ -46,6 +47,8 @@ class ForwardAPI(BaseHandler):
         tags:
           - Forward
         """
+        self.verify_user_global_permission(EVENT_FORWARD)
+
         operation = SchemaParser.parse_operation(
             self.request.decoded_body, from_string=True
         )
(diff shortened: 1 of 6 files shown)


src/app/beer_garden/api/http/handlers/v1/forward.py
```diff
@@ -37,7 +38,7 @@ async def post(self):
             schema:
                 $ref: '#/definitions/Forward'
         responses:
-          200:
+          204:
```
scott-taubman, line 41: Not related to this PR, but I noticed that this was wrong as the endpoint actually returns a 204.  At least in the one code path that appears to actually work.

jlrpnbbngtn (approved):
LGTM

PR #1212, opened 2021-12-07
#1191 - Garden config validation

### LINKED ISSUE #1191: Garden configuration validation
A garden's configuration info (connection type and parameters for the 2 kinds of connections) are written to the database without any kind of validation.

jlrpnbbngtn:
Closes #1191

### DIFF
--- a/src/app/beer_garden/db/mongo/api.py
+++ b/src/app/beer_garden/db/mongo/api.py
@@ -79,7 +79,7 @@ def from_brewtils(obj: ModelItem) -> MongoModel:
         # first step in decoupling from Brewtils
         from beer_garden.db.schemas.garden_schema import GardenSchema
 
-        model_dict = GardenSchema().dump(obj).data
+        model_dict = GardenSchema(strict=True).dump(obj).data
     else:
         model_dict = SchemaParser.serialize(obj, to_string=False)
     mongo_obj = MongoParser.parse(model_dict, type(obj), from_string=False)
@@ -118,7 +118,7 @@ def to_brewtils(
         # first step in decoupling from Brewtils
         from beer_garden.db.schemas.garden_schema import GardenSchema
 
-        schema = GardenSchema()
+        schema = GardenSchema(strict=True)
         serialized = schema.dumps(obj).data
         parsed = schema.loads(serialized, many=many).data
     else:
--- a/src/app/beer_garden/db/mongo/api.py
+++ b/src/app/beer_garden/db/mongo/api.py
@@ -75,7 +75,13 @@ def from_brewtils(obj: ModelItem) -> MongoModel:
         The Mongo model item
 
     """
-    model_dict = SchemaParser.serialize(obj, to_string=False)
+    if isinstance(obj, brewtils.models.Garden):
+        # first step in decoupling from Brewtils
+        from beer_garden.db.schemas.garden_schema import GardenSchema
+
+        model_dict = GardenSchema().dump(obj).data
+    else:
+        model_dict = SchemaParser.serialize(obj, to_string=False)
     mongo_obj = MongoParser.parse(model_dict, type(obj), from_string=False)
     return mongo_obj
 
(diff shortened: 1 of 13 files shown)


src/app/test/db/mongo/models/models_test.py
```diff
@@ -654,138 +652,3 @@ def test_user_delete_cascades_to_user_token(self, user_token):
         assert len(UserToken.objects.filter(id=user_token.id)) == 1
         user_token.user.fetch().delete()
         assert len(UserToken.objects.filter(id=user_token.id)) == 0
-
```
jlrpnbbngtn, line 657: These tests were moved into `test/db/mongo/models/garden_model_test.py`. There is no other change in this file.

src/app/beer_garden/db/schemas/garden_schema.py
```diff
@@ -0,0 +1,133 @@
...
+logger = logging.getLogger(__name__)
+
+
+class GardenBaseSchema(Schema):
+    """Class to give Marshmallow Schemas the desired behavior of throwing
+    exceptions on errors when marshalling/unmarshalling. Otherwise, each line of code
+    utilizing these would need to pull apart MarshalResult objects in order to return
+    a meaningful error."""
+
+    @pre_load
```
scott-taubman, line 19: Is there a reason to call explicitly call this during `pre_load` and `pre_dump`?  It should get naturally called during the load process and I'm not sure you actually want it to run when doing a `dump`...

src/app/beer_garden/db/schemas/garden_schema.py
```diff
@@ -0,0 +1,133 @@
...
+            many = True
+
+        _ = self.validate(obj, many=many)
+        dumped = super().dump(obj, many=many, update_fields=update_fields, **kwargs)
+
+        return dumped
+
+
+def _port_validator(value):
+    return 1024 <= value < 65535
```
scott-taubman, line 68: I don't think anything would preclude a garden from having it's public connection on a port lower than 1024.  That part of the check seems unnecessarily limiting.

src/app/beer_garden/db/schemas/garden_schema.py
```diff
@@ -0,0 +1,133 @@
...
+    status = fields.Str(allow_none=True)
+    status_info = fields.Nested(StatusInfoSchema, allow_none=True)
+    connection_type = fields.Str(allow_none=False)
+    connection_params = fields.Nested(GardenConnectionsParamsSchema, allow_none=True)
+    namespaces = fields.List(fields.Str(), allow_none=True)
+    systems = fields.Nested(SystemSchema, many=True, allow_none=True)
+
+    @post_load
+    def make_object(self, data):
+        return BrewtilsGarden(**data)
```
scott-taubman, line 133: This worries me a bit, for reasons that I elaborated on in the comment on the `from_brewtils` changes.

src/app/beer_garden/db/mongo/api.py
```diff
@@ -75,7 +75,13 @@ def from_brewtils(obj: ModelItem) -> MongoModel:
         The Mongo model item
 
     """
-    model_dict = SchemaParser.serialize(obj, to_string=False)
+    if isinstance(obj, brewtils.models.Garden):
```
scott-taubman, line 78: This makes me nervous.  I feel like one of the benefits of having the schema live in beer-garden should be to also get away from the dependence on the brewtils models.  That's a thing that we should do by moving away from this whole `to_brewtils` and `from_brewtils` dance over time where we can.

This looks like it's here so that existing code can attempt to use these new schemas as a sort of drop-in replacement where they had been using the brewtils ones.  I haven't gone dumpster diving into all of the places where that might be the case, but if that's what is indeed happening here my gut tells me that this is going to lead to subtly different and difficult to debug behavior.

src/app/beer_garden/db/schemas/garden_schema.py
```diff
@@ -0,0 +1,111 @@
...
+    http = fields.Nested("HttpConnectionParamsSchema", allow_none=True)
+    stomp = fields.Nested("StompConnectionParamsSchema", allow_none=True)
+
+
+class GardenSchema(GardenBaseSchema):
+    id = fields.Str(allow_none=True)
+    name = fields.Str(allow_none=False)
+    status = fields.Str(allow_none=True)
+    status_info = fields.Nested(StatusInfoSchema, allow_none=True)
+    connection_type = fields.Str(allow_none=False)
```
scott-taubman, line 104: This field being required breaks the child garden registration, as discussed.  I know there's more work after this PR, so I'm just putting a note here for you to keep in mind as you continue.

PR #1219, opened 2021-12-08
#1218 - Populate children in the request endpoint response

### LINKED ISSUE #1218: Request endpoint not providing children
The `/api/v1/request` endpoint is not populating the child requests as expected.  Instead, they are always coming back as null.  This is because the children are not stored in the database and are not a property on the model, but instead get populated exclusively by the `get_request` call, which is no longer being used by the endpoint method.

scott-taubman:
Closes #1218 
This Pull request requires brewtils: 3.8.0

This PR fixes the request endpoint so that `children` gets properly populated.

## Test Instructions
The unit test should pretty well cover this one, but for a functional test, just task the `nested-calls` command and then refresh the request page after it initially loads.  With this fix you should still see the child request on that page after the reload.  Before the fix they would disappear.

**NOTE**: Before the reload you can end up seeing a duplicate child event listed.  This is not a new bug and won't be addressed here.  It's an artifact of the requests finishing sufficiently fast that the initial API response contains the children, and then UI receives a published event on the websocket connection regarding the child request and just creates a second entry for it.

**NOTE 2**: There was a bug with the `nested-calls` example plugin that prevents it from starting.  Be sure to pull the latest master branch and use that version of `nested-calls` if you care to manually test this.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -122,6 +122,19 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
+      # Custom tag extractor to find Brewtils Branch
+      - name: Find Brewtils Branch
+        uses: TheBurchLog/body-env-tag-action@1.0
+        with:
+          tag: "brewtils:"
+          env-variable: "brewtils"
+          default-value: "develop"
+          tag-position: -1
+
+      # Used the hashtag here instead of slash allow for branches to have slashes in them
+      - name: Update Brewtils Branch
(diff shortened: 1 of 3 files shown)


.github/workflows/pull_request.yml
```diff
@@ -122,6 +122,19 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
+      # Custom tag extractor to find Brewtils Branch
```
scott-taubman, line 125: The code coverage action needs this in cases where a specific brewtils version is required (this same thing happens on the regular unit test action).  Since this PR is one such case, I just added it here.

jlrpnbbngtn (approved):
LGTM

PR #1222, opened 2021-12-08
Fix gridfs output handling

### LINKED ISSUE #1221: Request output stored in GridFS not being populated
As the title suggests, when the Request gets sufficiently large that the output has to be stored in GridFS, that output is not being returned as part of the response on the request endpoint (`/api/v1/request`).

scott-taubman:
Closes #1221

This PR reverts the `RequestAPI.get` method back to using the `Operation` flow for retrieving the request.  This results in an extra database lookup explicitly for the authorization check, but ensures that all of the data that is not native to the Request model gets properly added to the request for the response.  In most other cases the `Operation` flow was left intact when adding the authorization checks and it should have been kept here as well, so this is effectively correcting that mistake.

Additionally, this PR reverts a likely unintentional change where gridfs output was being stored in its serialized form rather than raw.  A separate comment on this PR goes into more detail on that.

## Test Instructions
The unit test covers this one pretty well.  If you want to functionally test this you could do the following:

* Task the "echo_json" command with something that's going to spill over and store in gridfs (Hand-hacking the `REQUEST_MAX_PARAM_SIZE` used by the Request model is an easy way to achieve that).
* View the request through the UI.  You should notice:
  * The output is actually displayed (it wasn't prior to these fixes)
  * It is displayed as proper JSON (it would be stringified JSON without the second fix)

### DIFF
--- a/src/app/beer_garden/db/mongo/models.py
+++ b/src/app/beer_garden/db/mongo/models.py
@@ -445,7 +445,7 @@ class Request(MongoModel, Document):
             output_json = json.dumps(self.output)
             if len(output_json) > REQUEST_MAX_PARAM_SIZE:
                 self.logger.info("Output size too big, storing in gridfs")
-                self.output_gridfs.put(output_json, encoding=encoding)
+                self.output_gridfs.put(self.output, encoding=encoding)
                 self.output = None
 
         super(Request, self).save(*args, **kwargs)
--- a/src/app/beer_garden/api/http/handlers/v1/request.py
+++ b/src/app/beer_garden/api/http/handlers/v1/request.py
@@ -16,7 +16,6 @@ from beer_garden.api.authorization import Permissions
 from beer_garden.api.http.base_handler import event_wait
 from beer_garden.api.http.exceptions import BadRequest
 from beer_garden.api.http.handlers import AuthorizationHandler
-from beer_garden.db.mongo.api import MongoParser
 from beer_garden.db.mongo.models import Request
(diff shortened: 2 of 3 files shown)


src/app/beer_garden/db/mongo/models.py
```diff
@@ -445,7 +445,7 @@ def save(self, *args, **kwargs):
             output_json = json.dumps(self.output)
             if len(output_json) > REQUEST_MAX_PARAM_SIZE:
                 self.logger.info("Output size too big, storing in gridfs")
-                self.output_gridfs.put(output_json, encoding=encoding)
+                self.output_gridfs.put(self.output, encoding=encoding)
```
scott-taubman, line 448: This is kind of a separate gridfs output issue that happened to be discovered while troubleshooting the original issue.  We made a change in 3.7.1 to how we decide when to store in gridfs.  As part of that change, we switched from storing the raw output to storing the serialized output.  I don't **think** that was intentional, but if it was then that was likely poor direction on my part.  I believe the intent was to serialize the output as a means of roughly estimating the storage size, but not to alter how the output is actually stored.  Storing the serialized version of the output breaks both the UI rendering and presumably any code users have written against request output that was expecting it in the old format.  This reverts the actual `put` line back to what it was in [3.6.0](https://github.com/beer-garden/beer-garden/blob/3.6.0/src/app/beer_garden/db/mongo/models.py#L436).

jlrpnbbngtn (approved):
LGTM

PR #1226, opened 2021-12-14
#877 - Added backend support for event blacklist by command

### LINKED ISSUE #877: Selectively prevent requests from being published to parent
We need to be able to prevent specific requests from being published to the parent. This should probably be a field on the Command that's able to be set at the `@command` level as well as the `@system` level.

This should be done after #777

west270:
closes #877
Added a new model for black listing commands to the database for preventing requests from corresponding commands to be sent to parent.
The requests blocked are base on the combination of namespace, system name, and command name.

Testing

1. start up a parent and child garden
2. in a python console import `CommandBlackList` and connect to child database
3. add command to black list 
```
CommandBlackList(
            namespace="child_namespace", system="echo", command="say"
        ).save()
```
4. Run corresponding command from either the parent or child

Results
If command is ran from the parent the request should be stuck in the created status.
If command is ran from child the parent should not know about the request.

### DIFF
--- a/src/app/beer_garden/api/stomp/manager.py
+++ b/src/app/beer_garden/api/stomp/manager.py
@@ -9,6 +9,7 @@ import beer_garden.log
 import beer_garden.requests
 import beer_garden.router
 from beer_garden.api.stomp.transport import Connection, parse_header_list
+from beer_garden.db.mongo.util import is_in_command_black_list
 from beer_garden.events import publish
 from beer_garden.events.processors import BaseProcessor
 
@@ -159,15 +160,16 @@ class StompManager(BaseProcessor):
                     garden_name=event.payload.name, skip_key=skip_key
                 )
 
-        for value in self.conn_dict.values():
-            conn = value["conn"]
-            if conn:
-                if conn.is_connected():
-                    if value["headers_list"]:
-                        for headers in value["headers_list"]:
-                            conn.send(event, headers=headers)
-                    else:
-                        conn.send(event)
+        if not is_in_command_black_list(event):
+            for value in self.conn_dict.values():
+                conn = value["conn"]
+                if conn:
+                    if conn.is_connected():
+                        if value["headers_list"]:
+                            for headers in value["headers_list"]:
+                                conn.send(event, headers=headers)
+                        else:
+                            conn.send(event)
 
     def handle_event(self, event):
         """Main event entry point
--- a/src/app/beer_garden/api/stomp/manager.py
+++ b/src/app/beer_garden/api/stomp/manager.py
@@ -9,8 +9,7 @@ import beer_garden.log
 import beer_garden.requests
 import beer_garden.router
 from beer_garden.api.stomp.transport import Connection, parse_header_list
-from beer_garden.db.mongo.util import is_in_command_black_list
-from beer_garden.events import publish
+from beer_garden.events import event_blacklisted, publish
 from beer_garden.events.processors import BaseProcessor
 
 logger = logging.getLogger(__name__)
@@ -160,7 +159,7 @@ class StompManager(BaseProcessor):
                     garden_name=event.payload.name, skip_key=skip_key
                 )
 
-        if not is_in_command_black_list(event):
+        if not event_blacklisted(event):
             for value in self.conn_dict.values():
                 conn = value["conn"]
                 if conn:
--- a/src/app/beer_garden/db/mongo/models.py
+++ b/src/app/beer_garden/db/mongo/models.py
@@ -76,7 +76,7 @@ __all__ = [
     "Role",
     "RoleAssignment",
     "User",
-    "CommandBlackList",
+    "CommandPublishingBlackList",
 ]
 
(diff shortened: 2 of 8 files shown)


src/app/test/db/mongo/util_test.py
```diff
@@ -268,6 +281,41 @@ def test_missing(self, model_mocks):
         assert role.save.called is True
 
 
+class TestCommandBlackList(object):
```
scott-taubman, line 284: If you move the blacklist function into `beer_garden/events/__init__.py` as suggested earlier, then this would similarly move to `tests/events/init_test.py`

src/app/beer_garden/db/mongo/models.py
```diff
@@ -849,6 +850,17 @@ class RawFile(Document):
     file = FileField()
 
 
+class CommandBlackList(Document):
```
scott-taubman, line 853: Let's call this `CommandPublishingBlackList` for the sake of clarity.

src/app/test/db/mongo/util_test.py
```diff
@@ -19,17 +23,24 @@ def model_mocks(monkeypatch):
     role_mock = Mock()
     job_mock = Mock()
     principal_mock = Mock()
+    black_list_mock = Mock()
```
scott-taubman, line 26: I know you modeled this on the other tests here, but I don't know that there's a compelling reason to use mocks here.  Especially for the Event model.  Try removing the mocks and working with actually model instances.  I suspect you won't run into any issues, but if you do we can work through them.

scott-taubman (changes requested):

src/app/beer_garden/db/mongo/util.py
```diff
@@ -388,3 +389,17 @@ def _should_create_admin():
     # anonymous users, we assume they do not want to re-create the admin
     # user.
     return False
+
+
+def is_in_command_black_list(event):
```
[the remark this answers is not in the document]
jlrpnbbngtn: > 
> ```python
> def event_blacklisted(event: Event) -> bool:
>     """Docstring goes here"""
>     return _event_blacklisted_by_name(event) or _event_blacklisted_by_command(event)
> ```
> 

I prefer `event_is_blacklisted` over `event_blacklisted` because it makes it clear that you're calling a predicate function. Otherwise, no disagreement.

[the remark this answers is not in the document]
scott-taubman: `event_is_blacklisted` is fine with me.  Let's go with that.

src/app/beer_garden/api/stomp/manager.py
```diff
@@ -159,15 +160,16 @@ def _event_handler(self, event):
...
-                        conn.send(event)
+        if not is_in_command_black_list(event):
+            for value in self.conn_dict.values():
+                conn = value["conn"]
+                if conn:
+                    if conn.is_connected():
+                        if value["headers_list"]:
+                            for headers in value["headers_list"]:
+                                conn.send(event, headers=headers)
+                        else:
```
jlrpnbbngtn, line 171: The first two `if`s can be collapsed into
```python
if conn and conn.is_connected():
    if value["headers_list"]:
        for header ...
    else :
        ...
```
I feel that -- in the general case -- it makes it clearer when multiple tests need to be true in order for code to run that the tests should appear together. (though this is totally stylistic preference, so it's up for debate)

On the other hand, I think it would be better if we try to avoid things like,
```python
if conn:
```
in favor of
```python
if conn is not None:
```
Also as a general rule, I think that a lot of danger for slippery bugs lies in *truthiness* checks. I feel it's always better to be explicit in what you're checking. So the same would apply to `if value["headers_list"]`.

scott-taubman: I think that bit only changed in the diff because it got nested under an additional if check.  It's otherwise unchanged from before, so probably should stay that way.

src/app/test/db/mongo/util_test.py
```diff
@@ -19,11 +19,13 @@ def model_mocks(monkeypatch):
     role_mock = Mock()
     job_mock = Mock()
     principal_mock = Mock()
+    event_mock = Mock()
```
scott-taubman, line 22: I'm guessing the updates to this file are accidental leftovers after you moved the code and tests.  You'll want to clean these up.

src/app/test/events/init.py
```diff
@@ -0,0 +1,57 @@
...
+import pytest
+from mock import Mock
+
+from beer_garden import config
+from beer_garden.db.mongo.models import CommandPublishingBlackList
+from beer_garden.events import event_blacklisted
+
+
+@pytest.fixture
+def model_mocks(monkeypatch):
```
scott-taubman, line 11: I think I commented on this in my original review, but is there a reason to use a `Mock()` here as opposed to just using the `Event` model?

scott-taubman (changes requested):

scott-taubman (approved):

PR #1232, opened 2021-12-16
#1215 - Cascade delete RawFile when deleting Request

### LINKED ISSUE #1215: Clean up files with corresponding Request
It is currently the case that when a Request with a file parameter gets deleted, the corresponding RawFile does not get cleaned up.  The behavior should be that the RawFile is deleted anytime the request is removed, whether that be from manual deletion, the pruning process, or any other means.

### LINKED ISSUE #1216: Improve GridFS handling
When Requests get sufficiently large, they are stored in GridFS.  The handling of this right now is not especially intelligent.  A couple of observations:

* Creating a single "large" Request object results in 4 files in the GridFS store being created (see `db.fs.files.count()`).
* Deleting a Request results in only 2 of these corresponding GridFS files getting deleted.  That leaves 2 that are just sort of orphaned.

We should improve our usage of GridFS as to reduce the amount of database bloat.

scott-taubman:
Closes #1215 
Closes #1216

This PR makes two improvements to the handling of FileField fields:

* The gridfs file referenced by FileFields in the `Request` and `RawFile` models will now be properly cleaned up when the model instance itself is deleted.
* The `Request` model's save() method no longer produces or orphans a bunch of duplicate gridfs entries.

To achieve this, the following was done:

* A `request` LazyReferenceField with a CASCADE delete rule was added to the RawFile model
* The existing `Request.save()` logic was moved into a `_pre_save()` and refined to remove gridfs duplication.
* A new `_post_save()` method on the Request model was added, which updates the `request` reference field on the appropriate RawFile to point to the newly saved Request.
* A `FileFieldHandlingQuerySet` and corresponding `FileFieldHandlingQuerySetNoCache` custom QuerySet is now used by RawFile and Request to ensure that the GridFS files referenced get properly deleted.  Without this the references would be deleted, but not the actual GridFS files.

Additionally, the pruner has been updated to clean up RawFiles on the same interval as Files.

## Test Instructions
To see the changes in action:
* For ease of testing, make sure you don't have any gridfs files in your database (`db.fs.files.drop()` will do the trick).
* Task one of the commands of the `file_bytes` example plugin.
* Task some other request, ensuring that the parameters or output (or both) go over the 5MB (or hand-hacked smaller value) threshold for being stored in gridfs.
* In the database, note the value of `db.fs.files.count()`.  There should be one entry each for the `file_bytes` request as well as the parameters and/or output of the other request (so 2 or 3 depending on what your test data).
* In a python console, do a `Request.objects.all().delete()` or `Request.objects.no_cache().all().delete()`.
* In the database, check that `db.fs.files.count()` is now 0.

### DIFF
--- a/src/app/test/db/mongo/models_test.py
+++ b/src/app/test/db/mongo/models_test.py
@@ -7,7 +7,8 @@ import pytest
 from brewtils.errors import ModelValidationError, RequestStatusTransitionError
 from brewtils.schemas import RequestTemplateSchema
 from mock import Mock
-from mongoengine import NotUniqueError, ValidationError, connect
+from mongoengine import GridFSProxy, NotUniqueError, ValidationError, connect
+from mongomock.gridfs import enable_gridfs_integration
 
 import beer_garden.db.api as db
 import beer_garden.db.mongo.models
@@ -20,6 +21,7 @@ from beer_garden.db.mongo.models import (
     Instance,
     Job,
     Parameter,
+    RawFile,
     Request,
     Role,
     RoleAssignment,
@@ -27,6 +29,8 @@ from beer_garden.db.mongo.models import (
     User,
     UserToken,
 )
+
+enable_gridfs_integration()
 
 
 class TestCommand(object):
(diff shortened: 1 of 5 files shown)


src/app/test/db/mongo/models_test.py
```diff
@@ -28,6 +30,8 @@
     UserToken,
 )
 
+enable_gridfs_integration()
```
scott-taubman, line 33: Something worth noting here: there's a fundamental incompatibility when using this `enable_gridfs_integration` that is going to require us to reorganize the unit tests to get around.  With this in here, it is currently the case that:

* Running the full suite of unit tests works fine
* Running just this test file in isolation does not work
* Running tests for any class in this file in isolation **other than** TestGarden works fine.

The conclusion I came to when trying to resolve this is that `enable_gridfs_integration` establishes some mocks that could break things depending on the order in which things get imported.

This test file is too big and should be broken up anyway, so I think we split out the tests for each model into their own test file and that would solve most of this.  I'm not going to do that here though because I believe there's another branch in progress that started to do this.

scott-taubman:
Put this back in draft for now because it's looking likely that this will be combined with the work for #1216 .

scott-taubman:
Closing this PR, which is now included in #1236

PR #1237, opened 2021-12-22
#1213 - Improve handling of error events

### LINKED ISSUE #1213: Error events are both noisy and confusing
In various situations, ERROR events get published that are of no discernable value, which clutters logs in the backend and creates confusion on the frontend.

For example, let's say we have a parent and child garden.  Someone attempts to task a command directly on the child garden, but an exception is thrown during the creation process.  The child will then publish an error event that gets propagated up to the parent, which will show something like this in its logs:

```shell
2021-12-07 18:09:45,396 - [MainProcess] - beer_garden.events.handlers - ERROR - Error event: <Event: namespace=None, garden=child1, name=REQUEST_CREATED, timestamp=2021-12-07 18:09:45.386000, error=True, error_message=Something happened, metadata={}, payload_type=None, payload=None>
```

Additionally, the same event gets published to the parent (and child) websocket connection, which results in the Requests page showing the "Updates Detected" message, but then when you try to load the updates nothing new shows because the update is just that there was an ERROR event, not an actual new Request.  This is very confusing for the users.

The published event has very little usefulness in its current form. First, there isn't enough information provided in the published event to really understand the context of the error, why it happened, or what one is supposed to do about it.  Second, the error is something that is completely contained to the child garden.  It was not caused by something that was initiated through the parent and it's not something that can be fixed in the context of the parent.  So even if there was additional context, it's not clear that providing that context to the parent is really the appropriate thing to do.

I would suggest we do the following:
* Try to determine if there is a **currently existing case** where beer-garden makes good use of these ERROR events.
* If not, we should just stop publishing them.
* If there are cases where they are used, we should be more discerning in our handling of them.  Specifically:
  * Either stop publishing them to the websocket or update the UI to ignore such events in the context of decision to display the "Updates Detected" message.
  * Either stop publishing such events upstream to parent gardens, or update the event handling so that such events are silently dealt with by the parent.

west270:
closes #1213 

Added check to see if event errors can be sent to parent. It currently blocks error events that name starts with `REQUEST` and have no payload.

## Testing Instructions

- Start a parent and child garden
- Add a raise somewhere in the critical path of request creation. The first line of the remove_bytes_parameter_base64 function in requests.py is a good choice.
- Restart the child garden and attempt to task something from it directly (i.e. don't go through the parent).

Result should be that the Requests page does not show `Updates Detected` and the error event should not be published in the parent logs.

### DIFF
--- a/src/app/beer_garden/api/stomp/manager.py
+++ b/src/app/beer_garden/api/stomp/manager.py
@@ -9,7 +9,7 @@ import beer_garden.log
 import beer_garden.requests
 import beer_garden.router
 from beer_garden.api.stomp.transport import Connection, parse_header_list
-from beer_garden.events import event_blacklisted, publish
+from beer_garden.events import can_send_event_to_parent, publish
 from beer_garden.events.processors import BaseProcessor
 
 logger = logging.getLogger(__name__)
@@ -158,8 +158,7 @@ class StompManager(BaseProcessor):
                 self.remove_garden_from_list(
                     garden_name=event.payload.name, skip_key=skip_key
                 )
-
-        if not event_blacklisted(event):
+        if can_send_event_to_parent(event):
             for value in self.conn_dict.values():
                 conn = value["conn"]
                 if conn:
--- a/src/app/beer_garden/api/stomp/manager.py
+++ b/src/app/beer_garden/api/stomp/manager.py
@@ -9,7 +9,7 @@ import beer_garden.log
 import beer_garden.requests
 import beer_garden.router
 from beer_garden.api.stomp.transport import Connection, parse_header_list
-from beer_garden.events import can_send_event_to_parent, publish
+from beer_garden.events import event_blocklisted, publish
 from beer_garden.events.processors import BaseProcessor
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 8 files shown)


scott-taubman (changes requested):
What's here seems good and is pretty straightforward, which I like.

Personally, I feel like we could have kept `event_blacklisted` as the main function that gets called to do this check, and then what you have as `_send_event_error` would be named something like `_event_is_blacklisted_error` and get called by `event_blacklisted`.  The error event check is essentially just another type of blacklist, albeit one that is baked into the code rather than user configurable.

Two other changes I'd also like to see:
* Even though the error events don't get sent to the parent, they still end up getting published on the websocket.  This means that the UI on the garden that was tasked will still get the phantom "Updates Detected" message.  I think the easiest thing to do wrt that would be to just have the UI check if the event is an error before displaying that message.  I'm guessing that check would go somewhere in [here](https://github.com/beer-garden/beer-garden/blob/develop/src/ui/src/js/controllers/request_index.js#L234).
* This is a minor thing I missed in the last review, but [`_event_blacklisted_by_command`](https://github.com/beer-garden/beer-garden/blob/develop/src/app/beer_garden/events/__init__.py#L147) can return `None`, when really it should just default to returning `False` in the path where the event doesn't match any of the check criteria.  I only noticed this because my editor yelled at me that `event_blacklisted` is supposed to return a `bool`, but could potentially return `None` because it calls `_event_blacklisted_by_command`

scott-taubman (approved):

PR #1244, opened 2022-01-10
#1225 - Increment job error count in all cases

### LINKED ISSUE #1225: Job error count is not incremented for certain errors
The `success_count` and `error_count` on Job currently only track whether the request that spawned from the job succeeded or failed.  This means that the `error_count` does not include, for instance, times when the job threw an exception before successfully spawning a request, such as during the routing to determine the target garden.

This is problematic because the net result is that a job could be failing repeatedly, but the error counter isn't incrementing at all (perhaps remaining at 0), making it difficult for someone to notice that there's a problem.

I think there's a relatively easy solution to this, which would be to increment the `error_count` in the [Exception block](https://github.com/beer-garden/beer-garden/blob/develop/src/app/beer_garden/scheduler.py#L516) where we're currently just logging the error.

Jake-Ross:
Closes #1225 

Change to increment job error count on exception (as opposed to just when the garden returns an error). This should fix issues where the job isn't being run for reasons unrelated to the request, yet the job error count is not incremented.

## To Test
- Create a job using any command from the the `error` system, ensure that the error count increments on each run of the job.
- Create a job that raises an exception at the job level, as opposed to at the request level. In fact, it may even be sufficient to just insert a manual exception into the area the except block covers.

### DIFF
--- a/src/app/beer_garden/scheduler.py
+++ b/src/app/beer_garden/scheduler.py
@@ -20,13 +20,8 @@ from brewtils.models import (
     Events,
     FileTrigger,
     Job,
-<<<<<<< HEAD
     Operation,
     Request,
-=======
-    Request,
-    Operation,
->>>>>>> Fix linting
 )
 from mongoengine import ValidationError
 from pathtools.patterns import match_any_paths
--- a/src/app/beer_garden/scheduler.py
+++ b/src/app/beer_garden/scheduler.py
@@ -523,7 +523,7 @@ def run_job(job_id, request_template, **kwargs):
         if updates != {}:
             db.modify(db_job, **updates)
     except Exception as ex:
-        logger.exception(f"Error executing {db_job}: {ex}")
+        logger.error(f"Error executing {db_job}: {ex}")
         db.modify(db_job, {"inc__error_count": 1})
 
     # Be a little careful here as the job could have been removed or paused
--- a/src/app/beer_garden/scheduler.py
+++ b/src/app/beer_garden/scheduler.py
@@ -20,8 +20,13 @@ from brewtils.models import (
     Events,
     FileTrigger,
     Job,
+<<<<<<< HEAD
     Operation,
     Request,
+=======
+    Request,
+    Operation,
+>>>>>>> Fix linting
 )
 from mongoengine import ValidationError
 from pathtools.patterns import match_any_paths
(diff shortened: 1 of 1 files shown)


src/app/beer_garden/scheduler.py
```diff
@@ -490,8 +498,8 @@ def run_job(job_id, request_template, **kwargs):
         request = beer_garden.router.route(
             Operation(
                 operation_type="REQUEST_CREATE",
-                model=request_template,
-                model_type="RequestTemplate",
+                model=Request.from_template(request_template),
```
scott-taubman, line 501: This change is already on develop, which made me look at the git log on this branch.  It looks like your starting point is pretty far off from the current develop.  Try to clean that up and make the current head of develop your starting point so the the PR only shows things you're actually changing here.

src/app/beer_garden/scheduler.py
```diff
@@ -512,9 +520,11 @@ def run_job(job_id, request_template, **kwargs):
             logger.debug(f"{db_job!r} request completed with SUCCESS status")
             updates["inc__success_count"] = 1
 
-        db.modify(db_job, **updates)
+        if updates != {}:
+            db.modify(db_job, **updates)
     except Exception as ex:
         logger.exception(f"Error executing {db_job}: {ex}")
```
scott-taubman, line 526: Not related to your changes, but `logger.exception` isn't intended to be used in this context.  Per the [logging docs](https://docs.python.org/3/library/logging.html#logging.Logger.exception):

>  This method should only be called from an exception handler.

Since you're here, let's change this to `logger.error` to clean this up.

scott-taubman (changes requested):

src/app/beer_garden/scheduler.py
```diff
@@ -523,7 +523,8 @@ def run_job(job_id, request_template, **kwargs):
         if updates != {}:
             db.modify(db_job, **updates)
     except Exception as ex:
-        logger.exception(f"Error executing {db_job}: {ex}")
+        logger.error(f"Error executing {db_job}: {ex}")
+        db.modify(db_job, {"inc__error_count": 1})
```
scott-taubman, line 527: This line needs to be:

```python
db.modify(db_job, inc__error_count=1)
```

`modify` is expecting to deal with kwargs.  As is this just ends up raising another exception rather than incrementing the error counter.

jlrpnbbngtn: Adding a double start is equivalent, but your way is more straightforward and therefore preferable.

scott-taubman (approved):

PR #1245, opened 2022-01-11
#1243 - Add authorization checks to websocket event publishing

### LINKED ISSUE #1243: Add authorization checks to websocket event publishing
The data that gets published over the websocket connection is currently not subject to the same authorization checks as the http API requests.  This should be changed so that messages only get published for events about an entity that the user has read access to.

scott-taubman:
Closes #1243 

This PR adds authorization checks to the events that get published on the websocket connections.  Prior to these changes, every connection received every message, regardless of the user's actual permissions.  With this PR, the flow goes like this:

* Client connects to the websocket
* An "AUTHORIZATION_REQUIRED" message is published to the client.
* The UI has a callback to handle this message and send an "UPDATE_TOKEN" message with the current access token.
* The server processes the "UPDATE_TOKEN" message and updates its record of the access token for this connection.  It will also publish an acknowledgement message named "TOKEN_UPDATED".
* Whenever messages would be published on the websocket connection, the user's permissions are checked (using the access token on record from above) against the object that the message is about. If the user has access, the message is published. If they do not, then nothing is published to that client.
* If while checking the user's permissions the token is deemed to be invalid or expired, an "AUTHORIZATION_REQUIRED" message is published **in lieu of the original message**.
  * At least for now, there is no attempt to queue up and deliver missed messages.  If the token on record is allowed to expire (it is not pre-emptively refreshed), messages will be missed while awaiting re-authorization.
  * The UI will pre-emptively publish an "UPDATE_TOKEN" message when a token refresh takes place.  However, this will only happen if another API call resulted in a refresh happening.  In this iteration the UI does not periodically check if the token is due for a refresh.  This means that if a user has not been navigating around the UI and having API calls made, the websocket connection will eventually have an expired token associated with it and no means of refreshing on its own.  Dealing with this specific scenario feels better left to a separate PR.  (Late update: I may have figured out how to accommodate this somewhat easily, so we could potentially tack that on to this PR). 

## Test Instructions
* Make sure auth is enabled on your garden
* Setup at least one user account with read access to systems and requests on your garden
* With the developer pane open in your browser, go to your beer garden instance and login.
* Look at the messages coming through on the `/api/v1/socket/events` websocket. You should see the behavior described above.
* Task something and verify that event messages come through that same websocket connection.

Another useful way to test things is to use the [websocat](https://github.com/vi/websocat) command line tool.  This will let you connect to the websocket and not immediately send the token.  Then if you task something (or do anything that would generate an event), you should see "AUTHORIZATION_REQUIRED" messages come through rather than the actual event messages.  It would look like this:

```shell
$ websocat ws://localhost:8080/api/v1/socket/events
{"name": "AUTHORIZATION_REQUIRED", "payload": "Access token required"}
{"name": "AUTHORIZATION_REQUIRED", "payload": "Valid access token required"}
{"name": "AUTHORIZATION_REQUIRED", "payload": "Valid access token required"}
{"name": "AUTHORIZATION_REQUIRED", "payload": "Valid access token required"}
```
Then you can simple type and press Enter to send a message.  The below alternates between messages entered into the websocat client and responses back from the server:
```shell
{"name": "UPDATE_TOKEN", "payload": "notavalidtoken"}
{"name": "AUTHORIZATION_REQUIRED", "payload": "Access token update message contained an invalid token"}
{"name": "INVALID_TYPE", "data": "dataisnotpayload"}
{"name": "BAD_MESSAGE", "payload": "Invalid message received. Error was: {'payload': ['Missing data for required field.'], 'name': ['Not a valid choice.']}"}
```
This is an easy way to play around with the websocket communications.  If you do send a message with a valid token, you should start getting actual event messages.

### DIFF
--- a/src/app/beer_garden/api/http/authentication/__init__.py
+++ b/src/app/beer_garden/api/http/authentication/__init__.py
@@ -180,7 +180,7 @@ def decode_token(encoded_token: str, expected_type: str = None) -> dict:
     token_type = decoded_token["type"]
     if expected_type and token_type != expected_type:
         raise InvalidTokenException(
-            f"Incorrect token type. Expected {expected_type}," "received {token_type}."
+            f"Incorrect token type. Expected {expected_type} received {token_type}."
         )
 
     return decoded_token
--- a/src/app/beer_garden/api/http/authentication/__init__.py
+++ b/src/app/beer_garden/api/http/authentication/__init__.py
@@ -73,13 +73,16 @@ def refresh_token_pair(refresh_token: str) -> dict:
             { "access": <str>, "refresh": <str> }
 
     Raises:
-        ExpiredTokenException: The supplied refresh token has expired, either due
-            to reaching it's natural expiration or having been revoked.
+        ExpiredTokenException: The supplied refresh token has expired, either due to
+            reaching it's natural expiration or having been revoked.
+        InvalidTokenException: The token could not be decoded or is the incorrect type
+            of token and is therefore invalid
     """
+    decoded_refresh_token = decode_token(refresh_token)
+
     try:
-        decoded_refresh_token = _decode_token(refresh_token)
         refresh_token_obj = UserToken.objects.get(uuid=decoded_refresh_token["jti"])
-    except (jwt.ExpiredSignatureError, UserToken.DoesNotExist):
+    except UserToken.DoesNotExist:
         raise ExpiredTokenException
 
     expiration = datetime.fromtimestamp(decoded_refresh_token["exp"], tz=timezone.utc)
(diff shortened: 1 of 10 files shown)


src/app/beer_garden/api/http/exceptions.py
```diff
@@ -46,16 +46,6 @@ class AuthenticationFailed(BaseHTTPError):
     reason: str = "Authentication Failed"
 
 
-class InvalidToken(BaseHTTPError):
```
scott-taubman, line 49: I decided to just get rid of these because the they were only being used in one place and with the refactor required for this branch I felt its usage was more confusing rather than less.

What the usage would have looked like with these here:
```python
except InvalidTokenException:
    raise InvalidToken
except ExpiredTokenException:
    raise ExpiredToken
```

And with these removed:

```python
except InvalidTokenException:
    raise RequestForbidden(reason="Authorization token invalid")
except ExpiredTokenException:
    raise AuthorizationRequired(reason="Authorization token expired")
```

I felt like that second one makes it more obvious what the HTTP error that's ultimately getting returned will be.

src/app/beer_garden/api/http/authentication/__init__.py
```diff
@@ -157,14 +234,3 @@ def _get_access_token_expiration() -> datetime:
 def _get_refresh_token_expiration() -> datetime:
     """Calculate and return the refresh token expiration time"""
     return datetime.utcnow() + timedelta(hours=12)
-
-
-def _decode_token(token) -> dict:
```
scott-taubman, line 162: This got refactored into `decode_token` above with necessary changes to make it usable with the `EventSocket` handler.

src/app/beer_garden/api/http/handlers/authorization_handler.py
```diff
@@ -172,63 +178,23 @@ def _anonymous_superuser(self) -> "User":
 
         return anonymous_superuser
 
-    def _get_user_from_request(self) -> "User":
```
scott-taubman, line 175: A modified version of this is now in `authentication/__init__.py` so that it could be reused in the `EventSocket` handler.

src/app/beer_garden/api/http/handlers/v1/event.py
```diff
@@ -1,18 +1,62 @@
...
+    """Check a that a user has access to view the supplied event"""
+    user_permitted = True
+    payload_type = event.payload_type.lower()
+
+    if event.payload and not user_has_permission_for_object(
+        user, f"{payload_type}:read", event.payload
+    ):
+        user_permitted = False
+
+    return user_permitted
```
jlrpnbbngtn, line 39: Could this whole thing just be
```python
return event.payload and not user_has_permission_for_object(
        user, f"{event.payload_type.lower()}:read", event.payload
    )
```
or is the intention to actually return `True` when the event doesn't have a payload?

src/app/beer_garden/api/http/processors.py
```diff
@@ -22,8 +20,6 @@ def put(self, event):
 def websocket_publish(item):
     """Will serialize an event and publish it to all event websocket endpoints"""
     try:
-        beer_garden.api.http.io_loop.add_callback(
-            EventSocket.publish, SchemaParser.serialize(item, to_string=True)
-        )
+        beer_garden.api.http.io_loop.add_callback(EventSocket.publish, item)
```
jlrpnbbngtn, line 23: Does this now receive the event after it's been serialized? In that case, the docstring should be updated.

jlrpnbbngtn (commented):
A couple of comments only. Otherwise GTG.

src/app/beer_garden/api/http/handlers/v1/event.py
scott-taubman: Returning true when there's no payload is intentional.  Basically if there's no payload then I treat the event as safe for everyone to receive.

src/app/beer_garden/api/http/processors.py
scott-taubman: Docstring updated.

PR #1249, opened 2022-01-21
#1214 - Fix job scheduler request handling

scott-taubman:
This PR fixes the job schedulers request handling so that it can properly keep track of which requests that it has kicked off have finished.  This was done by:

* Renaming `handle_event_for_entrypoints` to `handle_wait_events` to make it slightly more accurately named.
* Adding `handle_wait_events` to the list of processors that the Main process (which runs the scheduler) calls during its event processing.

It also fixes an issue where the auth check for events without a payload type would throw an error because the check function assumed there would always be a payload type.  A unit test has been added for this scenario.

Each of those fixes is isolated to their own commit, so look at the two commits if you're unsure which changes apply to which fix.

### DIFF
--- a/src/app/beer_garden/api/http/__init__.py
+++ b/src/app/beer_garden/api/http/__init__.py
@@ -381,7 +381,7 @@ def _event_callback(event):
     for handler in [
         beer_garden.router.handle_event,
         beer_garden.log.handle_event,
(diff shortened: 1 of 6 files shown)


jlrpnbbngtn (approved):
LGTM.

PR #1251, opened 2022-01-27
Issue/1179 UI permissions checks

### LINKED ISSUE #1179: Add updated permissions checks to hide UI elements.
It would be nice if we could hide certain UI elements if a user does not have access to interact with them.  For instance, hiding the "Admin" button in the top right if the user does not have the "system:update" or "garden:update"  permissions for anything.  There's more to it than that, because the user may have "garden:update" for one garden, but "garden:read" for some others, so the listing on that admin page would need to check individual garden level permissions and hide them.

It might be worth considering some server-side solutions to this.  Meaning, update the API so that there's a way for the UI to query for "only" the gardens that the user has access to administer.

scott-taubman:
requires beer-garden/brewtils#385
Closes #1179

This PR adds permissions checks to the UI for the following:

* The Admin menu option, as well as the System and Garden items within that menu.
* Whether the "make it happen" button on systems should be active or not (hiding it altogether created rendering wonkiness).
* The job create and import buttons on the job index page.
* The job management buttons on the job view page.


src/ui/src/js/run.js
```diff
@@ -275,13 +298,23 @@ export default function appRun(
...
   });
 
-  $interval(function() {
-    EventService.connect(TokenService.getToken());
-  }, 5000);
-
-  $interval(function() {
-    TokenService.preemptiveRefresh();
-  }, 30000);
+  $interval(
```
scott-taubman, line 301: These intervals were triggering the angular `digest()`, which re-checks and potentially re-renders everything, which meant the permission checks in the ng-show that I added all over the place would happen every interval for no real compelling reason.  Since neither interval should result in data changes that would make `digest()` necessary to call, I just updated these to not do that (the `false` that gets passed in now).

src/ui/src/js/run.js
```diff
@@ -151,13 +148,39 @@ export default function appRun(
     // );
   };
 
-  $rootScope.hasPermission = function(user, permissions) {
+  $rootScope.hasPermission = function(user, permission) {
```
scott-taubman, line 151: There's a way to do this that doesn't involve adding a bunch of stuff to the `rootScope`, but since that's largely the existing pattern I just stuck with it here, since it does make the code as it currently exists easier to work with.

jlrpnbbngtn (approved):
Code looks good and checked that the code does what it says it does.

PR #1254, opened 2022-02-02
#1192 - Garden connection UI import/export

### LINKED ISSUE #1192: UI update for garden parameter import/export
Add buttons to the Garden View page to import and export garden connection parameters.

jlrpnbbngtn:
Closes #1192 

#### Testing
Exercise the import and export functionality via the buttons. Note that:
1. There is minimal error checking and almost no validation. If an exported JSON file is mangled, the code makes a best-effort to avoid importing it. But if it's just augmented with garbage (that is otherwise legal JSON), or if it's entirely empty, it will import the garbage into the database. Caveat emptor. 
2. As a note on the STOMP tab says, if there are supposed to be headers but the fields are empty, the page must be refreshed. They should show up then. This is related to a shortcoming with angular-schema-forms and nested structures.

### DIFF
--- a/src/app/beer_garden/events/handlers.py
+++ b/src/app/beer_garden/events/handlers.py
@@ -50,7 +50,7 @@ def garden_callbacks(event: Event) -> None:
         try:
             handler(deepcopy(event))
         except Exception as ex:
-            logger.exception(
-                f"'{handler_tag}' event handler received an error executing callback"
-                f" for {event!r}: {ex}"
+            logger.error(
+                "'%s' handler received an error executing callback for event %s: %s"
+                % (handler_tag, repr(event), str(ex))
             )
--- a/src/app/beer_garden/events/handlers.py
+++ b/src/app/beer_garden/events/handlers.py
@@ -34,20 +34,23 @@ def garden_callbacks(event: Event) -> None:
         logger.debug(f"{event!r}")
 
     # These are all the MAIN PROCESS subsystems that care about events
-    for handler in [
-        beer_garden.application.handle_event,
-        beer_garden.garden.handle_event,
-        beer_garden.plugin.handle_event,
-        beer_garden.requests.handle_event,
-        beer_garden.requests.handle_wait_events,
-        beer_garden.router.handle_event,
-        beer_garden.systems.handle_event,
-        beer_garden.scheduler.handle_event,
-        beer_garden.log.handle_event,
-        beer_garden.files.handle_event,
-        beer_garden.local_plugins.manager.handle_event,
+    for handler, handler_tag in [
+        (beer_garden.application.handle_event, "Application"),
+        (beer_garden.garden.handle_event, "Garden"),
+        (beer_garden.plugin.handle_event, "Plugin"),
+        (beer_garden.requests.handle_event, "Requests"),
+        (beer_garden.requests.handle_wait_events, "Requests wait events"),
+        (beer_garden.router.handle_event, "Router"),
+        (beer_garden.systems.handle_event, "System"),
+        (beer_garden.scheduler.handle_event, "Scheduler"),
+        (beer_garden.log.handle_event, "Log"),
+        (beer_garden.files.handle_event, "File"),
+        (beer_garden.local_plugins.manager.handle_event, "Local plugins manager"),
     ]:
         try:
             handler(deepcopy(event))
         except Exception as ex:
-            logger.exception(f"Error executing callback for {event!r}: {ex}")
+            logger.exception(
+                f"'{handler_tag}' event handler received an error executing callback"
+                f" for {event!r}: {ex}"
+            )
--- a/src/ui/src/js/controllers/admin_garden_view.js
+++ b/src/ui/src/js/controllers/admin_garden_view.js
@@ -56,7 +56,6 @@ export default function adminGardenViewController(
   };
 
   $scope.successCallback = function(response) {
-    console.log('response', response);
     $scope.response = response;
     $scope.data = response.data;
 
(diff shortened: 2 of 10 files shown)


src/app/beer_garden/events/handlers.py
```diff
@@ -34,20 +34,23 @@ def garden_callbacks(event: Event) -> None:
...
-        beer_garden.plugin.handle_event,
-        beer_garden.requests.handle_event,
-        beer_garden.requests.handle_wait_events,
-        beer_garden.router.handle_event,
-        beer_garden.systems.handle_event,
-        beer_garden.scheduler.handle_event,
-        beer_garden.log.handle_event,
-        beer_garden.files.handle_event,
-        beer_garden.local_plugins.manager.handle_event,
+    for handler, handler_tag in [
```
jlrpnbbngtn, line 37: This is a leftover from a previous try. It just makes the error message dumped to the logs be more user friendly.

src/ui/src/js/services/garden_service.js
```diff
@@ -89,22 +116,184 @@ export default function gardenService($rootScope, $http) {
...
+    const simpleFieldMissing = getSimpleFieldPredicate(entryPointValues);
+    const httpSimpleFields = [
+      'ca_cert', 'client_cert', 'host', 'port', 'url_prefix',
+    ];
+    const allHttpFieldsEmpty = httpSimpleFields.every(simpleFieldMissing);
+
+    return allHttpFieldsEmpty;
+  };
+
+  GardenService.formToServerModel = function(data, model) {
```
scott-taubman, line 221: I haven't been able to find a scenario where all of this additional checking does what I think you're intended for it to do, at least in terms of it's potential usefulness to a user.  A few things I've tried:

1) Create a new garden, select "STOMP" as the connection type, change nothing else, click save.  It saved just fine, despite there being no values set for any of the STOMP parameters.  This is consistent with the existing behavior, but seemed like something these additions were aiming to prevent.
2) Clear all of the fields of the HTTP config, select HTTP as the connection type.  Click Export.  Nothing happens.  There's an error logged to the console, but nothing communicated to the user.
3) Same scenario, click "Save configuration".  All of the boxes get the green checkmark (despite being empty) making it appear as if things got validated and saved.  In reality, nothing happened and there's a warning message at the top saying as much, but I would have to scroll up to see that now.

It's not clear to me that the additional validation here is actually providing value, so I'd suggest we just drop it.  I'm open to hearing arguments though if you really thing there's something here.

src/ui/src/js/services/garden_service.js
```diff
@@ -89,22 +116,184 @@ export default function gardenService($rootScope, $http) {
...
-        model['connection_params'][field] = form[field];
       }
+      updatedConnectionParams[modelEntryPointName] = updatedEntryPoint;
     }
 
-    return model;
+    if (emptyConnections['http'] && emptyConnections['stomp']) {
+      throw Error('One of \'http\' or \'stomp\' connection must be defined');
+    } else if (emptyConnections['http'] && modelConnectionType === 'HTTP') {
+      throw Error('Connection type is \'HTTP\' but http connection parameters' +
```
scott-taubman, line 287: I lobbied for doing away with this validation in another comment, but in the event that this message sticks around, there's a space missing between "parameters" and "are".

scott-taubman (changes requested):

src/ui/src/js/services/garden_service.js
jlrpnbbngtn: Fixed. It was intentionally throwing an exception, but the exception handling code was removed with the validation code. Now the exception throwing code is removed as well and it should act consistently across the scenarios you describe.

src/ui/src/templates/import_garden_config.html
```diff
@@ -0,0 +1,58 @@
+<form ng-submit="doImport()">
+  <div class="modal-header">
+    <h3 class="modal-title" id="modal-title">Select File</h3>
+  </div>
+
+  <div class="modal-body" id="modal-body">
+    <p>
+      <span class="glyphicon glyphicon-info-sign fa-2x"></span> The config
```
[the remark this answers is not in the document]
jlrpnbbngtn: That's correct. Good catch.

src/app/beer_garden/events/handlers.py
```diff
@@ -34,20 +34,23 @@ def garden_callbacks(event: Event) -> None:
...
+        (beer_garden.scheduler.handle_event, "Scheduler"),
+        (beer_garden.log.handle_event, "Log"),
+        (beer_garden.files.handle_event, "File"),
+        (beer_garden.local_plugins.manager.handle_event, "Local plugins manager"),
     ]:
         try:
             handler(deepcopy(event))
         except Exception as ex:
-            logger.exception(f"Error executing callback for {event!r}: {ex}")
+            logger.exception(
```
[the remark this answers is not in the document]
jlrpnbbngtn: Fixed.

src/ui/src/js/services/garden_service.js
jlrpnbbngtn: This is removed.

src/ui/src/js/controllers/admin_garden_view.js
```diff
@@ -74,28 +90,146 @@ export default function adminGardenViewController(
...
+      if ('connection_params' in messages) {
+        const connParamErrors = messages['connection_params'];
+
+        for (const entryPoint in connParamErrors) {
+          if (connParamErrors.hasOwnProperty(entryPoint)) {
+            updateValidationMessages(entryPoint, connParamErrors[entryPoint]);
+
+            $scope.alerts.push({
+              type: 'warning',
+              msg: `Errors found in ${entryPoint} connection`,
```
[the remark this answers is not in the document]
jlrpnbbngtn: Per discussion...

scott-taubman (approved):
Behaves as expected in my test scenarios now and other comments have been addressed.

PR #1255, opened 2022-02-04
#1250 - Require garden:update in order to view connection parameters

### LINKED ISSUE #1250: Add an authorization check for viewing garden connection parameters
The gardens API response currently returns all of a garden's properties as long as the user has the "garden:read" permission.  This includes the garden connection information, which really shouldn't be viewable by users with just "garden:read" access, since it includes things like username and password.

We should instead have a separate authorization check to determine if the connection information should be included in the response or not.  Since the "garden:update" permission implies that the use can change the connection parameters, that's probably a reasonable permission to check when determining if they can read them as well.

scott-taubman:
Closes #1250 

This PR adds a check for the "garden:update" permission to decide whether the gardens API response should include the connection parameters.  It does this using an alternate schema if the user does not have the required permission.

Additionally, there was a TODO in the unit test file for the garden handler tests for adding some tests for POST, so I tackled that here as well.

## Test Instructions
The new unit tests cover this one okay, but to observe it in action you can do the following:

* As a user with the "garden:update" permission for at least one garden, hit the `/api/v1/gardens` and `/api/v1/gardens/<garden_name>` endpoints.  You can also just hit the UI logged in as the user and view the API calls in the dev panel.  In both cases you should see the "connection_params" included in the garden response.
* Do the same for a use without "garden:update" permissions.  The "connection_params" should not be included in the response.

### DIFF
--- a/src/app/beer_garden/api/http/handlers/v1/garden.py
+++ b/src/app/beer_garden/api/http/handlers/v1/garden.py
@@ -1,10 +1,14 @@
 # -*- coding: utf-8 -*-
+import json
+
 from brewtils.errors import ModelValidationError
 from brewtils.models import Operation
 from brewtils.schema_parser import SchemaParser
 
 from beer_garden.api.authorization import Permissions
 from beer_garden.api.http.handlers import AuthorizationHandler
+from beer_garden.api.http.schemas.v1.garden import GardenReadSchema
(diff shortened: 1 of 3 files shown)


jlrpnbbngtn (approved):
LGTM

PR #1261, opened 2022-02-19
#1258 - User management UI

### LINKED ISSUE #1258: Allow user management via the UI
There should be a means of managing users (also roles and role assignments) through the UI.

scott-taubman:
Closes #1258 

This PR adds the ability to update user role assignments via both the API and the UI.

Through the UI, if the logged in user has the global `user:update` permission, a "Users" menu item will be available in the "Admin" dropdown.  From that page a new user can be created or an existing user can be selected from the list.  Selecting an existing user brings up a simple management page where the user's role assignment can be managed or the user can be deleted.

# Test Instructions
* Make sure your role definitions are updated such that superuser (or some other role) has the `user:*` permissions.
* Login to the UI as a user with those `user:*` permissions in the "Global" scope.
* Select the "Users" menu item from the Admin drop down to get to the new pages.
* Test out creating, removing, and modifying users.  Make sure things work as you'd expect and the user experience seems somewhat reasonable.


src/ui/src/js/controllers/admin_user_view.js
```diff
@@ -0,0 +1,242 @@
...
+  };
+
+  const generateUserSF = function() {
+    $scope.userSchema = roleAssignmentSchema;
+    $scope.userForm = roleAssignmentForm;
+    $scope.userModel = serverModelToForm($scope.data);
+
+    // This gets set here because $scope.roleNames is not yet set at
+    // roleAssignmentSchema declaration time
+    $scope.userSchema.properties.role_assignments.items.properties.role_name.enum =
```
scott-taubman, line 130: The way the code is written, I believe it should be the case that the role names are guarenteed to be loaded by the time we get here (see the very last line of this file).  And then the form is redrawn immediately after this line.   Given that, what is supposed to happen is that the "Role" field is a dropdown populated by the actual list of roles that currently exist, per the `/api/v1/roles/` endpoint.  

For reasons that are totally unclear, this works like 80 - 90% of the time.  Every once in a while the role field will render as just a regular text input field rather than a dropdown.  I feel like this is a bug with the ASF library and not my code.  I tried a few variations on this, including defining the entire schema right here so that I didn't have to inject the enum property after the fact.  Nothing I tried fixed this inconsistent behavior.  If someone else has suggestions as to why they think that might be happening I'm open to trying to fix it.  Otherwise I'm chalking it up to an ASF bug and moving on.

src/ui/src/js/controllers/admin_user_index.js
```diff
@@ -0,0 +1,75 @@
...
+}
+
+newUserController.$inject = ['$scope', '$uibModalInstance'];
+
+/**
+ * newUserController - New User controller.
+ * @param  {$scope} $scope                        Angular's $scope object.
+ * @param  {$uibModalInstance} $uibModalInstance  Angular UI's $uibModalInstance object.
+ */
+export function newUserController($scope, $uibModalInstance) {
```
scott-taubman, line 65: This file was originally `admin_user.js` and got renamed here.  I made a bunch of changes to the other controller in this file, but didn't really touch this one.  As such, the new user dialog is pretty simple and does not provide a whole lot of feedback.  I might address that down the road, but for now it works so I left it as is.

jlrpnbbngtn (approved):
Functionality tested in the UI and code reviewed for glaring errors. Everything looks good. Merging.

PR #1266, opened 2022-02-22
#1263 - Show username on requests in the UI if auth is enabled

### LINKED ISSUE #1263: Record username on all requests
The username of the user that initiated a request should be recorded on the request.  This will help with auditing of requests and is necessary for allowing user's to have read access to their own requests (see #1264).

scott-taubman:
Closes #1263 

This PR adds the "requester" column to the request index and view pages if auth is enabled.  The requester is also visible on child requests that spawn (such as on the `nested-calls` example plugin).

The original issue (as written) was for recording the username on the request, but it turned out that was already happening.  So surfacing that info in the UI was all that was needed.

Test Instructions:
* With auth disabled, go to the requests page.  It should not show a requester column.
* With auth enabled, task something and go to the requests page.  You should see a requester column with the username of the user that spawned the request listed.

### DIFF
--- a/src/ui/src/js/controllers/request_index.js
+++ b/src/ui/src/js/controllers/request_index.js
@@ -1,6 +1,7 @@
 import {formatDate} from '../services/utility_service.js';
 
 requestIndexController.$inject = [
+  '$rootScope',
   '$scope',
   '$compile',
   'localStorageService',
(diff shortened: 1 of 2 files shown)


jlrpnbbngtn (approved):
Looks good.

PR #1268, opened 2022-02-23
#1264 - Allow implicit access to a user's own requests

### LINKED ISSUE #1264: Provide a way for users to always have access to their own requests
It would be useful if users could have access to see their own requests without having to be given access to see all requests for a given garden or system.

This requires #1263

scott-taubman:
Closes #1264 

This PR allows users to view requests where they are the requester, even if they do not have a role assignment explicitly granting access to the request.

This is done by defining a list of `OBJECT_OWNER_PERMISSIONS` that are conferred to users for any object where they are the owner.  This permission list is currently just `request:read`, and only requests are considered to have an owner, but this implementation allows for the potential expansion of both the eligible object types and the conferred permissions.

## Test Instructions
* Remove the "request:read" permission from the roles it is assigned to.
* Create requests as multiple users.
* Verify that each user can see their own requests, and only their own requests, since none of them will actually have "request:read" granted through any of their role assignments.

### DIFF
--- a/src/app/beer_garden/authorization.py
+++ b/src/app/beer_garden/authorization.py
@@ -10,6 +10,7 @@ from mongoengine import Document, DoesNotExist, Q, QuerySet
 from mongoengine.queryset.visitor import QCombination
 
 import beer_garden.db.mongo.models
+from beer_garden.api.authorization import Permissions
 from beer_garden.db.mongo.models import (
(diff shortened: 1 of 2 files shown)


jlrpnbbngtn (approved):
LGTM. Ops check good.

PR #1269, opened 2022-02-28
#1262 - Add password management

### LINKED ISSUE #1262: Add password management to the UI
The UI needs the following:

* Provide the means for all users to change their own password.
* Allow admins to reset the password of any user.

scott-taubman:
Closes #1262 

This PR adds password management to the UI.  There are two different places that this appears:

1) For all users, there is a new "Change Password" option in the top right ☰ dropdown menu.  This brings up a modal window where the user can change their own password.
2) On the admin only user management page where users' roles are managed, it is now also possible to change the user's password.

A new `/api/v1/password/change` API endpoint was also added to facilitate `#1`.

## Testing Instructions
* Try to change a password using the two methods described above.

### DIFF
--- a/src/app/beer_garden/api/http/schemas/v1/user.py
+++ b/src/app/beer_garden/api/http/schemas/v1/user.py
@@ -1,5 +1,5 @@
 from brewtils.schemas import RoleAssignmentDomainSchema
-from marshmallow import Schema, ValidationError, fields, post_load
+from marshmallow import Schema, ValidationError, fields, post_load, validate
 
 from beer_garden.db.mongo.models import Role, RoleAssignment, RoleAssignmentDomain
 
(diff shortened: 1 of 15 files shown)


src/app/beer_garden/api/http/schemas/v1/user.py
```diff
@@ -44,3 +44,12 @@ class UserPatchSchema(Schema):
...
     password = fields.Str()
     role_assignments = fields.List(fields.Nested(RoleAssignmentPatchSchema))
+
+
+class UserPasswordChangeSchema(Schema):
+    """Schema for changing a user's password"""
+
+    current_password = fields.Str(required=True)
+    new_password = fields.Str(
+        required=True, validate=validate.Length(min=1, error="Password required")
```
scott-taubman, line 54: The minimum length of 1 is just to ensure it's not possible to set the password to an empty string.  Any password rules more complicated than that requires a level of consideration that I wasn't interested in pursuing in the scope of this PR.

jlrpnbbngtn (approved):
Stood up to my attempts to break it. Code looks good. Merging.

Issue #1265, opened 2022-02-22
Sync users to remote gardens


scott-taubman:
User management in multi-garden setups would be much easier if a local garden could sync user info to remote gardens.  This could work similar to the "Garden Sync" capability, but in reverse.  The "Garden Sync" informs all remote gardens to send their system information up to a local garden.  In this case, the local garden would be sending user information down to all of the remote gardens.

Some considerations:
* Username should be used as the key for syncing.  The internal ids are unlikely to match across gardens.
* The local garden should filter the user's role assignments down to just those that are relevant to the remote garden.  Meaning if a user has role assignments for a garden that is not known to the remote garden (it is not in the remote garden's list of namespaces), that role assignment should not be included in the sync.
* Bonus points if an "auto-sync" option can be made available.  If this configuration option is enabled, any time a User changes, those changes would be synced to all remote gardens.

PR #1270, opened 2022-03-07
#1265 - Add user syncing across gardens (backend-only)

scott-taubman:
This is the backend portion of #1265.  A separate PR will be opened for the UI changes.

This PR adds backend support for syncing users from a local garden down to all remote gardens. As implemented it is an all or nothing proposition.  Doing a user sync will sync **all** users on the local garden with **all** remote gardens. Support for limiting the scope of the sync would not be difficult, but feels unnecessary at this time given the intended use case.

The workflow does something like this:
**NOTE:** other than the first bullet, all referenced functions are in `beer_garden/user.py`
* The `/api/v1/gardens` endpoint now supports a "sync_users" PATCH operation, which is used to initiate the user sync. This felt like the appropriate home both because of the all or nothing nature described earlier and because this feels conceptually like another for of the garden sync, which is kicked off the same way.
* The "sync_users" operation results in a call to `initiate_user_sync`, which goes through each remote garden, filters each user's role assignments down to the ones appropriate for that garden, and then routes a `USER_SYNC` operation to it.
* The remote garden handles the `USER_SYNC` operation by calling the `user_sync` function, which updates the user entries on that garden and then calls `initiate_user_sync` itself so that the process can continue to any gardens that are further downstream (what we often refer to as "grandchildren").
* The `update_user` function has been updated so that ever time a user is updated (such as via `user_sync`), a `USER_UPDATED` event gets published.  This is what will inform the local garden that the user sync was actually successful on the remote.
* The `USER_UPDATED` event is handled by `_handle_user_updated_event`, which creates a `RemoteUser` entry for the updated user.  `RemoteUser` tracks a simply serialized version of a a user's role assignments, per garden.
* The `user_synced_with_garden` function is not used yet, but it takes a `User` and `Garden` and checks the User's role assignments against those of the corresponding `RemoteUser`.  The idea is that rather than showing the full details for the RemoteUser entries in the UI, I think we'll just show the results of the `user_synced_with_garden` check for each remote garden, since all we really care about right now is if the remote garden's role assignments are up to date with the local user or not.

## Test Instructions
There's a lot of setup for this one.  The unit tests coverage should be pretty good, but here's the high level of how to manually test it:
* Setup some users with role assignments on a local garden.
* Setup multiple remote gardens.  You may want a "grandchild" garden as well.  Either don't setup users on those remote gardens, or at least make sure they have role assignments that differ from the local garden.
* Execute a `PATCH` on `/api/v1/gardens` with a body of `{"operation": "sync_users"}`.
* Check that the remote gardens' users have had their role assignment updated appropriately.
* The local garden should have `RemoteUser` entries created for the users on the remote gardens now.  You'll have to check this either via the model in a shell or in the database directly.

### DIFF
--- a/src/app/beer_garden/api/http/handlers/v1/garden.py
+++ b/src/app/beer_garden/api/http/handlers/v1/garden.py
@@ -12,6 +12,7 @@ from beer_garden.authorization import user_has_permission_for_object
 from beer_garden.db.mongo.api import MongoParser
 from beer_garden.db.mongo.models import Garden
 from beer_garden.garden import local_garden
+from beer_garden.user import initiate_user_sync
 
 GARDEN_CREATE = Permissions.GARDEN_CREATE.value
 GARDEN_READ = Permissions.GARDEN_READ.value
@@ -245,6 +246,7 @@ class GardenListAPI(AuthorizationHandler):
           updates to apply. Currently the only operations are:
 
           * sync
+          * sync_users
 
           ```JSON
           [
@@ -264,10 +266,8 @@ class GardenListAPI(AuthorizationHandler):
             schema:
               $ref: '#/definitions/Patch'
         responses:
-          200:
-            description: Execute Patch action against Gardens
-            schema:
-              $ref: '#/definitions/Garden'
+          204:
+            description: Patch operation has been successfully forwarded
           400:
             $ref: '#/definitions/400Error'
           404:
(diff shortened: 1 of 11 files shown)


src/app/beer_garden/api/http/handlers/v1/garden.py
```diff
@@ -264,10 +266,8 @@ async def patch(self):
             schema:
               $ref: '#/definitions/Patch'
         responses:
-          200:
-            description: Execute Patch action against Gardens
-            schema:
-              $ref: '#/definitions/Garden'
+          204:
```
scott-taubman, line 269: Despite what this said before, it seemed like the "sync" operation didn't return any content.  Neither does the "sync_users" operation, so I just changed this to a 204, which is more appropriate.

jlrpnbbngtn (approved):
Single minor comment. Otherwise LGTM.

PR #1272, opened 2022-03-08
#1271 - Slightly reduce request event spam and silence transition error events

### LINKED ISSUE #1271: Error events are still overly noisy
Some work has been done recently to reduce noisy ERROR event logging, specifically for request state transition errors.  These problems continue to persist however, especially for remote gardens that are more than a single hop away (e.g. "grandchild" gardens).

scott-taubman:
Closes #1271 

This PR aims to get rid of the error events for request state transitions entirely.  It also attempts to be at least nominally smarter about when to call `update_request` during the event processing, so that we're less likely to be publishing "REQUEST_UPDATED" events when in fact nothing has changed.  In the case of the websocket event spamming, this cuts the number of events down to about 1/3 of what they were in the case of a "grandchild" garden (From nearly 60 to around 20).

Event publishing and processing is still fundamentally flawed, but this PR should at least remove the error events and reduce the general event spamming.

## Test Instructions
Setup a remote garden that's two hops away (a "grandchild").  Task it from the local garden and make sure tasking works as expected.  In the local garden log, you should not see any ERROR events that look like:

```2022-03-08 15:51:31,606 - [MainProcess] - beer_garden.events.handlers - ERROR - Error event: <Event:...```

### DIFF
--- a/src/app/beer_garden/requests.py
+++ b/src/app/beer_garden/requests.py
@@ -893,17 +893,24 @@ def handle_event(event):
             # the subset of fields that change "corrects" the parent
             existing_request = db.query_unique(Request, id=event.payload.id)
 
-            if existing_request:
-                for field in ("status", "output", "error_class", "status_updated_at"):
-                    setattr(existing_request, field, getattr(event.payload, field))
-
-                try:
-                    update_request(existing_request)
-                except RequestStatusTransitionError:
-                    pass
-            else:
+            if existing_request is None:
                 # Attempt to create the request, if it already exists then continue on
                 try:
                     db.create(event.payload)
                 except NotUniqueException:
                     pass
+            elif event.name != Events.REQUEST_CREATED.name:
+                request_changed = False
+
+                for field in ("status", "output", "error_class", "status_updated_at"):
+                    new_value = getattr(event.payload, field)
+
+                    if getattr(existing_request, field) != new_value:
+                        request_changed = True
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/app/beer_garden/requests.py [hunk 1]
```diff
@@ -893,17 +893,24 @@ def handle_event(event):
...
-                except RequestStatusTransitionError:
-                    pass
-            else:
+            if existing_request is None:
                 # Attempt to create the request, if it already exists then continue on
                 try:
                     db.create(event.payload)
                 except NotUniqueException:
                     pass
+            elif event.name != Events.REQUEST_CREATED.name:
```
scott-taubman, line 902: This just ensures we don't do any work on a "REQUEST_CREATED" event if we've already recorded the event locally.  The only thing that could have changed at that point is "status_updated_at", which isn't really useful yet.

src/app/beer_garden/requests.py
```diff
@@ -893,17 +893,24 @@ def handle_event(event):
...
+                request_changed = False
+
+                for field in ("status", "output", "error_class", "status_updated_at"):
+                    new_value = getattr(event.payload, field)
+
+                    if getattr(existing_request, field) != new_value:
```
scott-taubman, line 908: This other check was added to make sure we only actually call `update_request` later on if something has in fact changed.  Again, just trying to reduce the event spamming.

jlrpnbbngtn (approved):

PR #1273, opened 2022-03-10
#1265 - UI updates for user syncing

### LINKED ISSUE #1265: Sync users to remote gardens

scott-taubman:
Closes #1265 

This PR adds the UI components (and the API changes to drive the UI) for the user syncing.  With this PR the following changes are present in the UI:

* On the User admin index page there is a "Sync Users" button next to the "Create Users" button.  Clicking this brings up a modal that explains what the syncing is all about and lets you click "Sync" to start the process.
  * The "Sync Status" for each garden is initially listed as "PENDING" if the garden is in a "RUNNING" state.  If the garden is not running, the sync status is just listed as "NOT RUNNING", indicating that we can't actually sync to that garden right now.
  * Once "Sync" is clicked, a spinner will be shown in place of "PENDING" to denote that the sync is in progress.  Once the sync is done, it will change to "COMPLETE".  **NOTE:** There is currently no timeout and no error reporting, so if the sync does not work for some reason, the spinner will currently just spin indefinitely.
  * When the sync finishes for each garden, the status will update in the modal.  Once they are all complete, you can either click outside of the modal to dismiss it or just click "Cancel".
* Also on the User index page, there is a new column labeled "Fully Synced" which will show either a ✓ (the user is synced with all remote gardens) or an X (they are not).
* On the individual User view page, there is a "Sync Status" table at the top which shows the breakdown of if the user is synced to each remote garden.  The idea being that if someone saw an X on the index page for a user, they could drill down and see the specifics of where they aren't synced.

All of the above items only show up if there are remote gardens for a user to be synced to.  If there are no remote gardens then  all of that sync information would just be clutter, so it is instead not displayed.

Some backend changes made to facilitate some of this new functionality:

* The API response for users now returns a "sync_status" attribute, which lists whether the user is synced to each remote garden.
* Some new websocket event publishing checks have been put in place.  Specifically:
  * USER_UPDATED events do not get published at all.  There's no reason for anyone to see those right now.
  * USERS_IMPORTED events require global "garden:update" access (the access required to do the sync in the first place).  This is to ensure that the existence of a remote garden is not inadvertently leaked through a websocket event that someone has no use for.

## Test Instructions
For testing, you'll basically want to setup some users and some remote gardens and play around with the new features described above, ensuring that they behave as you'd hope / expect.

### DIFF
--- a/src/app/beer_garden/api/http/schemas/v1/user.py
+++ b/src/app/beer_garden/api/http/schemas/v1/user.py
@@ -1,5 +1,14 @@
 from brewtils.schemas import RoleAssignmentDomainSchema, RoleAssignmentSchema
-from marshmallow import Schema, ValidationError, fields, post_dump, post_load, validate
+from brewtils.schemas import UserSchema as BrewtilsUserSchema
+from marshmallow import (
+    Schema,
+    ValidationError,
+    fields,
+    post_dump,
+    post_load,
+    pre_dump,
+    validate,
+)
 
 from beer_garden.db.mongo.models import Role, RoleAssignment, RoleAssignmentDomain
 
@@ -73,3 +82,17 @@ class UserSyncSchema(Schema):
         for item in data:
             for assignment in item.get("role_assignments", []):
                 assignment["role_name"] = assignment.pop("role")["name"]
+
+
+class UserSchema(BrewtilsUserSchema):
+    sync_status = fields.Dict(dump_only=True)
+
+    @pre_dump
+    def get_sync_status(self, user):
+        from beer_garden.user import user_sync_status
+
+        user.sync_status = user_sync_status(user)
+
+
+class UserListSchema(Schema):
+    users = fields.List(fields.Nested(UserSchema()))
--- a/src/app/beer_garden/user.py
+++ b/src/app/beer_garden/user.py
@@ -72,7 +72,9 @@ def initiate_user_sync() -> None:
     from beer_garden.router import route
 
     users = User.objects.all()
-    gardens = Garden.objects.filter(connection_type__nin=["LOCAL", None])
+    gardens = Garden.objects.filter(
+        connection_type__nin=["LOCAL", None], status="RUNNING"
+    )
 
     for garden in gardens:
         filtered_users = [
@@ -104,6 +106,7 @@ def user_sync(serialized_users: List[dict]) -> None:
         None
     """
     _import_users(serialized_users)
+    _publish_users_imported()
     initiate_user_sync()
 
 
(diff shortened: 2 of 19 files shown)


src/app/beer_garden/user.py
```diff
@@ -123,17 +126,35 @@ def user_synced_with_garden(user: User, garden: Garden) -> bool:
     # Avoiding circular imports
     from beer_garden.api.http.schemas.v1.user import UserSyncSchema
 
+    user = _filter_role_assigments_by_garden(user, garden)
+
     try:
         remote_user = RemoteUser.objects.get(username=user.username, garden=garden.name)
     except RemoteUser.DoesNotExist:
-        return False
+        return len(user.role_assignments) == 0
```
scott-taubman, line 134: As originally written this returned `False` if there was no `RemoteUser` entry, when actually there being no entry is the correct state if there are no relevant role assignments.

src/app/beer_garden/user.py
```diff
@@ -72,7 +72,9 @@ def initiate_user_sync() -> None:
     from beer_garden.router import route
 
     users = User.objects.all()
-    gardens = Garden.objects.filter(connection_type__nin=["LOCAL", None])
+    gardens = Garden.objects.filter(
+        connection_type__nin=["LOCAL", None], status="RUNNING"
```
scott-taubman, line 76: The `route` call seemed to fail if the garden wasn't running, resulting in a 500 error.  This ought not be the case, but I just decided to not attempt to route to gardens that aren't running.

src/ui/src/js/controllers/admin_user_index.js
```diff
@@ -41,35 +44,46 @@ export function adminUserIndexController($scope, $uibModal, UserService) {
...
+  };
 
   $scope.$on('userChange', () => {
     loadUsers();
   });
 
   loadUsers();
 }
-
-newUserController.$inject = ['$scope', '$uibModalInstance'];
```
scott-taubman, line 58: I moved this controller into its own file.  I needed to add another controller for the user index page and felt like it belonged in its own file, so I just moved this one too to keep it consistent.

src/app/beer_garden/api/http/schemas/v1/user.py
```diff
@@ -73,3 +82,17 @@ def dump_obj(self, data, many=False, **kwargs):
...
             for assignment in item.get("role_assignments", []):
                 assignment["role_name"] = assignment.pop("role")["name"]
+
+
+class UserSchema(BrewtilsUserSchema):
+    sync_status = fields.Dict(dump_only=True)
+
+    @pre_dump
+    def get_sync_status(self, user):
+        from beer_garden.user import user_sync_status
```
scott-taubman, line 92: The way the code is organized makes it increasingly difficult to avoid circular dependencies.  Specifically the relationship between the models.py and the user.py that houses all of the helpers for interacting with Users.

src/ui/src/js/controllers/admin_user_view.js
```diff
@@ -73,7 +73,7 @@ export default function adminUserViewController(
     // identifiers for us anyway.
     model.role_assignments.forEach((assignment) => {
       if (assignment.domain.scope === 'Global') {
-        assignment.domain.identifiers.name = '';
+        assignment.domain.identifiers = {name: ''};
```
scott-taubman, line 76: This isn't really related to the main issue, it's just fixing a bug I discovered while testing.

PR #1275, opened 2022-03-10
#1274 - Create admin user and superuser role on new garden startup

### LINKED ISSUE #1274: Create a default admin user for new installations
There is currently no default admin account created when a new environment is initialized.  This means that with auth enabled there is no real means of logging and doing anything unless you either use the trusted header authentication or do some out of band user setup.

We should provide a configuration option that allows users to define:
* The username for an admin account
* The starting password
* ~~The roles that should be assigned (they will be assigned for the "Global" scope).~~

To keep things simple, we can make it so that this default account is created only if no users exist already.  If any users exist, we'll assume that someone has set them up the way they want and that we shouldn't go adding a default admin account.

scott-taubman:
Closes #1274 

This PR adds steps to the garden initialization process that will create a default admin user, and create a default superuser role with all permissions that can be assigned to the admin user.

Some notes:
* The admin username and password can be customized via the app config.
* The admin user will not be created if any other users already exist
* The superuser role be created if it does not exist and there are no existing users.  Users existing indicates some setup has been done already and we preserve that setup.  An existing superuser role will never be modified.
* The code I touched had some references to the "guest_login_enabled" configuration option.  That option didn't do anything in the current iteration of beergarden, so I just opted to remove all references to it throughout the code.

## Test Instructions
Start a fresh instance of beergarden.  You should see log messages indicating that the admin user and superuser role are created.  Without any other setup, you should be able to login to the UI as admin / password (unless you changed the defaults in your config).

### DIFF
--- a/src/app/beer_garden/db/mongo/util.py
+++ b/src/app/beer_garden/db/mongo/util.py
@@ -1,6 +1,5 @@
 # -*- coding: utf-8 -*-
 import logging
-import os
 from typing import Union
 
 import yaml
@@ -13,9 +12,10 @@ from mongoengine.errors import (
     InvalidDocumentError,
     ValidationError,
 )
-from passlib.apps import custom_app_context
 
 from beer_garden import config
+from beer_garden.api.authorization import Permissions
+from beer_garden.db.mongo.models import Role, RoleAssignment, User
 from beer_garden.errors import ConfigurationError
 from beer_garden.role import sync_roles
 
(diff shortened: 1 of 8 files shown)


src/app/beer_garden/db/mongo/util.py
```diff
@@ -52,83 +52,60 @@ def ensure_roles():
     role_definition_file: Union[str, None] = config.get("auth.role_definition_file")
 
     if role_definition_file:
-        logger.info(f"Syncing role definitions from {role_definition_file}")
```
scott-taubman, line 55: I moved this into a separate function just to make it easier to mock in the unit tests.

src/app/test/db/mongo/util_test.py
```diff
@@ -315,19 +285,11 @@ def setup_class(cls):
     def teardown_method(self):
         beer_garden.db.mongo.models.Role.drop_collection()
 
-    def config_get_value(self, config_name):
```
scott-taubman, line 318: These got moved into mocks outside of the class so I could use them elsewhere.  That's why a bunch of unrelated tests got touched here.

jlrpnbbngtn (approved):
LGTM

PR #1281, opened 2022-03-22
#1280 - Improve login process

### LINKED ISSUE #1280: Login form lacks feedback
When logging in to the UI there is no feedback if you fail to authenticate.  Instead it just appears as if clicking "login" didn't do anything.  The form should provide feedback that the user failed to authenticate.

scott-taubman:
Closes #1280 

This PR does the following:

* Provides feedback on the login form when authentication fails by displaying a message that reads "Invalid username or password".
* Attempts to auto-login a user if trusted header authentication is enabled.  This makes for a smoother experience where a user is not required to click the login button if they have already authenticated via their a proxy server that is passing along trusted headers.

Work on these changes highlighted the fact that the HTTP response code for failed logins had been incorrect (it was 401).  This PR changes that to 400 so that it failed login responses do not get caught up in the http interceptor code intended for handling actual data returning endpoints.

## Testing instructions
* In the login dialog, enter an incorrect username or password.  You should see a message indicating the failure.
* Access beergarden through a trusted proxy.  It should automatically log you in.

### DIFF
--- a/src/app/beer_garden/api/http/exceptions.py
+++ b/src/app/beer_garden/api/http/exceptions.py
@@ -42,7 +42,7 @@ class AuthorizationRequired(BaseHTTPError):
 
 
 class AuthenticationFailed(BaseHTTPError):
-    status_code: int = 401
+    status_code: int = 400
     reason: str = "Authentication Failed"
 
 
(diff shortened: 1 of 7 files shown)


jlrpnbbngtn (approved):
LGTM

PR #1282, opened 2022-03-23
Issue/1228 Adds api for interacting with command block list

west270:
Adds api to add and remove  commands from the command publishing block list, and get the commands that are in the block list

## Testing Instructions
Make the api calls through terminal or swagger. If using authentication there needs to be an access token sent with api call.

1. Use the api to make a post to add command to block list. Example input: 
```
[
  {
    "system": "string",
    "command": "string",
    "namespace": "string"
  }
]
```
2. Use api to get list of commands in the command publishing block list, which should contain the command(s) added from the post.
3.  Use an id for one of the commands in the list that was returned to make an api call to delete a command, and repeat step 2 to verify command was removed.

If command is from child garden the command should be added or removed from parent and child if api call to post or delete are made from the parent. Also the entries on parent and child should have matching ids. To verify commands added to the child a get needs to be performed on the child.


src/app/beer_garden/api/http/handlers/v1/command_publishing_blocklist.py
```diff
@@ -0,0 +1,140 @@
...
+              $ref: '#/definitions/CommandPublishingBlocklist'
+          400:
+            $ref: '#/definitions/400Error'
+          50x:
+            $ref: '#/definitions/50xError'
+        tags:
+          - Command Block List
+        """
+        errors = []
+        for command in self.request_body:
```
scott-taubman, line 114: You'll want to use the marshmallow schemas to do input validation.  That will give you a clean way to check for required fields and such, and then you can catch the ValidationError from marshmallow if it's raised and reraise the appropriate exception from `beer_garden.api.http.exceptions`.  That will get you the appropriate HTTP response for validation errors.

There is also a `schema_validated_body` helper available that may or may not be useful for this.

src/app/beer_garden/api/http/handlers/v1/command_publishing_blocklist.py
```diff
@@ -0,0 +1,140 @@
...
+                    command["id"] = blocked_command._data["id"].__str__()
+
+                beer_garden.router.route(
+                    Operation(
+                        operation_type="COMMAND_BLOCK_LIST_ADD",
+                        kwargs={"command": command},
+                        target_garden_name=command["namespace"],
+                    )
+                )
+            except Exception as e:
```
scott-taubman, line 137: Generic note:  Don't catch all exceptions.

src/app/beer_garden/api/http/schemas/v1/command_publishing_blocklist.py
```diff
@@ -0,0 +1,9 @@
+from marshmallow import Schema, fields
+
+
+class CommandPublishingBlocklistSchema(Schema):
```
scott-taubman, line 4: As referred to in earlier you'll want a CommandPublishingBlocklistListSchema (as silly as that name sounds) that facilitates returning a list of things.  Again, see the UserListSchema as a reference.

src/app/beer_garden/command_publishing_blocklist.py
```diff
@@ -0,0 +1,22 @@
...
+    blocked_command.save()
+    return blocked_command
+
+
+def command_publishing_block_list_remove(command_publishing_id):
+    blocked_command = CommandPublishingBlockList.objects.get(id=command_publishing_id)
+    blocked_command.delete()
+
+
+def command_publishing_block_list_get():
```
scott-taubman, line 15: This likely goes away in favor of using the schema to generate the appropriate response.

src/app/beer_garden/router.py
```diff
@@ -156,6 +159,9 @@
     "RUNNER_RESCAN": beer_garden.local_plugins.manager.rescan,
     "PUBLISH_EVENT": beer_garden.events.publish,
     "USER_SYNC": beer_garden.user.user_sync,
+    "COMMAND_BLOCK_LIST_ADD": beer_garden.command_publishing_blocklist.command_publishing_block_list_adds,
+    "COMMAND_BLOCK_LIST_GET": beer_garden.command_publishing_blocklist.command_publishing_block_list_get,
```
scott-taubman, line 163: This one goes away.

src/app/test/api/http/unit/handlers/v1/command_publishing_blocklist_test.py
```diff
@@ -0,0 +1,394 @@
...
+        request = HTTPRequest(
+            url, method="POST", headers=headers, body=json.dumps(body)
+        )
+
+        response = yield http_client.fetch(request)
+
+        assert response.code == 201
+
+    @pytest.mark.gen_test
+    def test_post_auth_401_no_token(
```
scott-taubman, line 338: You don't actually need to test the no token scenario here.  That's a function of the `AuthorizationHandler` that the blocklist handlers are built on and is tested as part of its test suite.

src/app/test/api/http/unit/handlers/v1/command_publishing_blocklist_test.py
```diff
@@ -0,0 +1,394 @@
...
+
+        assert excinfo.value.code == 401
+
+    @pytest.mark.gen_test
+    def test_get(self, http_client, base_url, blocklist):
+        url = f"{base_url}/api/v1/commandpublishingblocklist/"
+
+        request = HTTPRequest(url, method="GET")
+
+        response = yield http_client.fetch(request)
```
scott-taubman, line 194: Mostly just an FYI, but you don't need to construct a HTTPRequest for a basic "GET".  That's the default behavior if you just pass in a URL here, so if you cared to you could just do:

```python
response = yield http_client.fetch(url)
```

src/app/beer_garden/api/http/handlers/v1/command_publishing_blocklist.py
```diff
@@ -0,0 +1,110 @@
...
+          50x:
+            $ref: '#/definitions/50xError'
+        tags:
+          - Command Block List
+        """
+        blocked_command = CommandPublishingBlockList.objects.get(
+            id=command_publishing_id
+        )
+        garden = Garden.objects.filter(
+            namespace=blocked_command["namespace"], name=blocked_command["system"]
```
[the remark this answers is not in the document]
scott-taubman: Less minor things:

* namespace is not a field on Garden.  The field you are looking for is name.
* See my comment elsewhere about the fact that you're getting back a `QuerySet` here and not a `Garden` object.

src/app/beer_garden/command_publishing_blocklist.py
```diff
@@ -0,0 +1,50 @@
...
+        )
+        blocked_command.status = "ADD_REQUESTED"
+        blocked_command.save()
+    else:
+        blocked_command = CommandPublishingBlockList(**command)
+        blocked_command.save()
+    return blocked_command
+
+
+def command_publishing_blocklist_delete(blocked_command_id):
```
scott-taubman, line 29: This was okay taking an actual `blocked_command` which is a `CommandPublishingBlocklist` instance.  That helps avoid the double db lookup.

src/app/beer_garden/command_publishing_blocklist.py
```diff
@@ -0,0 +1,50 @@
...
+            )
+        )
+        blocked_command.status = "REMOVE_REQUESTED"
+        blocked_command.save()
+    else:
+        blocked_command.delete()
+
+
+# todo remove since might not need
+def command_publishing_blocklist_remove(command_publishing_id):
```
scott-taubman, line 48: This should be what the router calls for the `COMMAND_BLOCKLIST_REMOVE` operation.  It's essentially just going to be:

```python
CommandPublishingBlockList.objects.filter(id=command_publishing_id).delete()
```

Probably with some exception handling as appropriate.

src/app/beer_garden/api/http/handlers/v1/command_publishing_blocklist.py
```diff
@@ -0,0 +1,119 @@
...
+          50x:
+            $ref: '#/definitions/50xError'
+        tags:
+          - Command Block List
+        """
+        commands = self.schema_validated_body(CommandPublishingBlocklistListInputSchema)
+        checked_gardens = []
+        for command in commands["command_publishing_blocklist"]:
+            if command["namespace"] not in checked_gardens:
+                _ = self.get_or_raise(Garden, SYSTEM_UPDATE, name=command["namespace"])
```
scott-taubman, line 105: I didn't notice this until I was testing it again, but using `self.get_or_raise` here has an unwanted side effect in that it returns a 404 if the garden doesn't exist.  In this context, the response should be a 400 if the garden doesn't exist since the user supplied bad data.  404 would indicate that the `/api/v1/commandpublishingblocklist` endpoint doesn't exist, which isn't true.

There's a few ways that you can handle this, but the easiest is probably to just catch the `NotFound` exception and re-raise as a `BadRequest`.  Something like:

```python
# This goes up top with the imports
from beer_garden.api.http.exceptions import BadRequest, NotFound

# Then here you would do
try:
    _ = self.get_or_raise(Garden, SYSTEM_UPDATE, name=command["namespace"])
except NotFound:
   raise BadRequest(reason=f"Invalid garden name: {command['namespace']}")
```

src/app/beer_garden/command_publishing_blocklist.py
```diff
@@ -0,0 +1,76 @@
+from brewtils.models import Operation
+
+import beer_garden.config as config
+from beer_garden.db.mongo.models import CommandPublishingBlockList
```
scott-taubman, line 4: The model name seems to be the last one standing in the great war over naming consistency. It's still `BlockList` camel case, rather than `Blocklist` like everything else.  It's not the end of the world honestly, but it should be an easy enough thing to change so I vote that we change it and then everything will be consistent.

PR #1287, opened 2022-04-05
#1286 - Correctly display column filters on request table

### LINKED ISSUE #1286: Request table column filtering off by one
The filtering headers on the Request table columns are off by one because the filter for "Requester" currently shows regardless of whether auth is enabled or not.  This should be fixed so that the filter columns are appropriately aligned.

scott-taubman:
Closes #1286 

This PR fixes an issue where the filter headers on the Request table were off by one when auth is disabled.  To fix this, the `$scope.dtOptions.lightColumnFilterOptions` is now constructed after `$scope.dtColumns` is finished being built.  This ensures that only columns that are actually present get added to the column filters.

## Testing Instructions
With auth enabled and disabled, navigate to the Request index page.  The filter columns should properly align and work in both cases.  On develop, the later columns are shifted by one because the filter for the "requester" field was present even if the column was not.

### DIFF
--- a/src/ui/src/js/controllers/request_index.js
+++ b/src/ui/src/js/controllers/request_index.js
@@ -82,78 +82,7 @@ export default function requestIndexController(
             },
         );
       })
-      .withLightColumnFilter({
-        0: {
-          html: 'input',
-          type: 'text',
-          attr: {class: 'form-inline form-control', title: 'Command Filter'},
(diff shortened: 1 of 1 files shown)


jlrpnbbngtn (approved):
Yep, this is good. Nice catch!

PR #1289, opened 2022-04-07
#1288 - Fix to allow local plugins to connect to rabbitmq over TLS

### LINKED ISSUE #1288: Plugins are unable to connect to rabbitmq using certificates
Certificate information is currently not included in the rabbitmq connection information provided to local plugins.  This means means that if certificates are required to connect to rabbitmq, it is not possible for the plugins to connect.

scott-taubman:
Closes #1288 

This PR makes changes to allow local plugins to properly connect to rabbitmq when it is setup to require client certificates.  Specifically, it now includes the full set of `mq_config.connections.message.ssl` settings instead of just `mq_config.connections.message.ssl.enabled` so that local plugins receive the location of the certificate file to use when connecting.

## Test Instructions
This should be tested in together with [brewtils/pull/390](https://github.com/beer-garden/brewtils/pull/390).

To test, you must setup rabbitmq to accept TLS connections and require client certificates.  You must then ensure that your `mq.message.ssl` and `mq.admin.ssl` settings are configured, with the client cert specifically pointing to a cert / key bundle.  With all of that configured, start up beergarden and make sure everything connects okay.

### DIFF
--- a/docker/docker-compose/tls/docker-compose.yml
+++ b/docker/docker-compose/tls/docker-compose.yml
@@ -100,7 +100,7 @@ services:
         restart: always
 
     rabbitmq:
-        image: rabbitmq:management-alpine
+        image: rabbitmq:3.8-management-alpine
         restart: always
         hostname: rabbitmq
         networks:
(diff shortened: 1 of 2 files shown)


docker/docker-compose/tls/docker-compose.yml
```diff
@@ -100,7 +100,7 @@ services:
         restart: always
 
     rabbitmq:
-        image: rabbitmq:management-alpine
+        image: rabbitmq:3.8-management-alpine
```
scott-taubman, line 103: The changes to this file do 2 things:

* Pin the rabbitmq version to 3.8, which is the latest version that actually works with the environment variable configuration we have defined.
* Requires client certificates to be present for rabbitmq.

These changes aren't really testable yet, but it mostly doesn't matter.  Basically this is intended to be an easy way to fire up a beergarden stack to play around with.  This file won't work because it uses the latest bg docker image release, which as of right now still wouldn't actually have the rest of the changes from here or brewtils.  Once we make the next release, we can try try using this file and further tweak if necessary.  It's already broken as-is and certainly won't get more broken by what's here.

PR #1303, opened 2022-05-06
Issue/1302 role sync

### LINKED ISSUE #1302: Allow roles to be synced across gardens
Currently the roles need to be managed on each garden individually.  In cases where the expectation is that users will be managed on a central garden and then synced to any remote gardens, it would be especially useful if the roles could be synced as well.  This would make role management across many gardens much easier and help avoid situations where a role update was not deployed to a garden, resulting in users that have been synced to downstream gardens not actually having permissions that match those of the upstream garden because the role definitions differ.

This is fundamentally very similar to the user sync and could behave and be presented the exact the same way.

scott-taubman:
Closes #1302 

This PR adds functionality for role syncing across gardens.  The backend portion very closely mirrors the workflow for the user syncing.  The `PATCH` method on the `/api/v1/gardens` endpoint now accepts a `sync_roles` operation, which sends the role definitions down to all known remote gardens.  The remote gardens import the roles and repeat the process for their own downstream gardens.

The models. functions, and eventing exist for checking the sync status of the roles on downstream gardens, but that information is not used anywhere at the moment.  If / when we create a dedicated role management UI, the information needed to display the sync status will already be present.

For the UI, the role syncing is bundled together with the user syncing, with a check box available to opt out of the role syncing.  Verbiage has been added to the user sync modal to explain the role sync behavior.

## Test Instructions
The easiest way to test is to use the user sync option in the UI.

* Stand up a local and remote garden.  Make sure auth is enabled.
* Make some changes to the remote garden so that the users and roles are not in sync with the local garden.
* Use the user sync functionality in the UI and keep the "sync roles" box checked.
* Check the remote garden and ensure that the users and roles are now in sync with the local.
* Repeat this exercise, but uncheck the "sync_roles" checkbox.
* The Users should be synced on the remote garden, but the roles should remain untouched.

### DIFF
--- a/src/app/beer_garden/api/http/handlers/v1/garden.py
+++ b/src/app/beer_garden/api/http/handlers/v1/garden.py
@@ -12,6 +12,7 @@ from beer_garden.authorization import user_has_permission_for_object
 from beer_garden.db.mongo.api import MongoParser
 from beer_garden.db.mongo.models import Garden
 from beer_garden.garden import local_garden
+from beer_garden.role import initiate_role_sync
 from beer_garden.user import initiate_user_sync
 
 GARDEN_CREATE = Permissions.GARDEN_CREATE.value
@@ -174,7 +175,7 @@ class GardenListAPI(AuthorizationHandler):
         summary: Retrieve a list of Gardens
         responses:
           200:
-            description: Garden with the given garden_name
+            description: A list of all gardens
             schema:
               type: array
               items:
@@ -247,6 +248,7 @@ class GardenListAPI(AuthorizationHandler):
 
           * sync
           * sync_users
+          * sync_roles
 
           ```JSON
           [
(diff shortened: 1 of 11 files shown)


src/app/beer_garden/api/http/handlers/v1/garden.py
```diff
@@ -247,18 +248,14 @@ async def patch(self):
...
           * sync_users
+          * sync_roles
 
           ```JSON
           [
             { "operation": "" }
           ]
           ```
         parameters:
-          - name: garden_name
```
scott-taubman, line 257: Noticed this was wrong when testing the changes via the swagger docs.

jlrpnbbngtn (changes requested):
From a setup with a parent and a single child:
* add role to user on parent, sync with 'sync roles' deselected. Expected: new role does not appear within roles on child. Actual: New role appears
* remove role from user on parent, sync with 'sync roles' deselected. Expected: existing role continues to appear on child. Actual: role is removed on child.

Could it be that the 'sync roles' checkbox does not fire the correct code?

scott-taubman:
The refactor of this is pretty substantial, so I'm going to close this PR and open a new one when ready.

PR #1307, opened 2022-05-10
#1306 - Protected roles

### LINKED ISSUE #1306: Make superuser role internally managed
The superuser role is currently subject to some confusing / inconvenient behavior.  The role is intended to contain all permissions, such that someone (such as the default admin account) assigned the superuser role has full access for the assigned domain (all access to all the things in the case of the "Global" scope).

There are, however, two caveats to this:

* The superuser role is created during garden initialization, but never automatically updated.  This means that if a new permission is introduced in a future release, the superuser role doesn't automatically receive that new permission.  The garden administrator is left having to micromanage the superuser role in order for it to continue serving its intended function.
* If a role definition file is used to define the roles (currently the only way to actually define roles), the superuser role must be present, or else it will get deleted and recreated every time the garden starts.  This is problematic because any role assignments for that role will be deleted with the role.

These problems are probably most easily solved by having a concept of "protected" roles that are intended to be internally managed.  We can make the superuser role one such role and then make sure it exists and is properly configured whenever the garden starts.  This would allow us to make sure that the role always gets any new permissions that are created and allows the users to stop including it in their role definition file.

scott-taubman:
Closes #1306 

This PR adds the concept of "protected" roles and makes "superuser" one such role.  A protected role is one that is intended to be internally managed and is therefore not modified or deleted during any role imports or syncs.

## Test Instructions
* Start your garden.  The superuser role will be created / configured as necessary.
* Set your auth -> role_definition_file setting to point to `example_configs/roles.yaml`
* Restart your garden and note that the "superuser" role stays intact, despite it not being present in the `roles.yaml`.
* Add the "superuser" role to the roles.yaml.  For example:
  ```yaml
    - name: "superuser"
      permissions:
        - "garden:read"
    ```
* Restart your garden and note that the "superuser" role again has not changed.  If you look at the beer-garden logs you should see the message "Role sync request for protected role superuser will be ignored," informing you that "superuser" was not changed even though it was present in the `roles.yaml`.


src/app/test/db/mongo/util_test.py
```diff
@@ -365,19 +365,3 @@ def test_ensure_roles_creates_superuser_role_if_none_exists(self, monkeypatch):
         superuser = Role.objects.get(name="superuser")
 
         assert len(superuser.permissions) == len(Permissions)
-
-    def test_ensure_roles_preserves_existing_superuser(self, monkeypatch):
```
scott-taubman, line 369: The things this was testing for is not intended to be true with the changes to this PR.  Thus the removed test.

jlrpnbbngtn (approved):
LGTM

PR #1308, opened 2022-05-11
issue #1229 Adds UI for Command Publishing Blocklist

### LINKED ISSUE #1229: Create UI for interacting with command block list
Need to create UI for interacting with commands to block list api.

west270:
Closes #1229 

Adds a UI for managing the command publishing blocklist.

The UI for interacting with the command publishing blocklist can be found at `<your_host_path>/commandblocklist` or using the link found in the admin drop-down button

### DIFF
--- a/src/ui/src/js/controllers/add_command_publishing_blocklist.js
+++ b/src/ui/src/js/controllers/add_command_publishing_blocklist.js
@@ -28,101 +28,112 @@ export default function AddCommandPublishingBlocklistController(
     $state,
     commandBlocklist,
 ) {
+  $scope.instanceCreated = function(_instance) {
+    $scope.dtInstance = _instance;
+  };
+
   $scope.dtOptions = DTOptionsBuilder.newOptions()
       .withOption('autoWidth', false)
       .withOption(
           'pageLength', 10,
       )
       .withOption('order', [
-        [0, 'asc'],
         [1, 'asc'],
         [2, 'asc'],
+        [3, 'asc'],
       ])
       .withLightColumnFilter({
+        0: {
+          html: 'input',
+          type: 'checkbox',
+          attr: {
+            'class': 'pull-right',
+            'title': 'Select All',
+            'id': 'select-all-checkbox',
+            'style': 'width: 1%;',
+            'onclick': 'document.getElementById("select-all-hidden-checkbox").click()',
+          },
+        },
         1: {
           html: 'input',
           type: 'text',
-          attr: {class: 'form-inline form-control', title: 'Namespace Filter'},
+          attr: {class: 'form-inline form-control', title: 'Namespace Filter', style: 'height: 1%'},
         },
         2: {
           html: 'input',
           type: 'text',
-          attr: {class: 'form-inline form-control', title: 'System Filter'},
+          attr: {class: 'form-inline form-control', title: 'System Filter', style: 'height: 1%'},
         },
         3: {
           html: 'input',
           type: 'text',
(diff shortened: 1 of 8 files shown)


scott-taubman (changes requested):

src/ui/src/js/controllers/add_command_publishing_blocklist.js
```diff
@@ -0,0 +1,130 @@
...
+          $state.reload();
+          alert('Successfully added commands to command publishing blocklist.');
+        },
+        (errorResponse) => {
+          alert(`Failure! Server returned status ${errorResponse.status}`);
+        },
+    );
+  };
+
+  $scope.instanceCreated = function(_instance) {
```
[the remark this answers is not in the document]
west270: It is for interacting with the table, which is now being used to get the filtered data so the select all functionality on selects the filtered items.

src/ui/src/js/controllers/add_command_publishing_blocklist.js
```diff
@@ -0,0 +1,141 @@
...
+        },
+    );
+  };
+
+  $scope.selectAllValue = false;
+
+  $scope.selectAll = function() {
+    const filteredData = $scope.dtInstance.DataTable.rows( {search: 'applied'} ).data();
+    $scope.selectAllValue = !$scope.selectAllValue;
+    for (let i = 0; i < filteredData.length; i++) {
```
scott-taubman, line 105: another `forEach` candidate.

src/ui/src/js/controllers/command_publishing_blocklist.js
```diff
@@ -0,0 +1,109 @@
...
+          html: 'input',
+          type: 'text',
+          attr: {class: 'form-inline form-control', title: 'Status Filter', style: 'height: 1%'},
+        },
+      })
+      .withBootstrap();
+
+  $scope.deleteBlocklistEntry = function(blockedCommandId, index) {
+    CommandPublishingBlocklistService.
+        deleteCommandPublishingBlocklist(blockedCommandId).then(
```
scott-taubman, line 66: There's currently no feedback when this delete fails.  There should be one of those red error banners at the top of the page similar to what happens on the garden admin page when the save fails.

src/ui/src/js/controllers/add_command_publishing_blocklist.js
```diff
@@ -0,0 +1,141 @@
...
+    $uibModalInstance.dismiss('cancel');
+  };
+
+  $scope.onCommandCheckChange = function(command) {
+    command.isChecked = !command.isChecked;
+    $scope.selectAllValue = false;
+    document.getElementById('select-all-checkbox')['checked'] = false;
+  };
+
+  $scope.submitBlocklistEntries = function() {
```
scott-taubman, line 85: There's currently an opportunity for confusion when no items are selected here, because the UI messaging makes it sound like an entry was created anyway.  I made a comment in #1314 in regards to improving the backend portion of this, but in the UI we could do more to help as well.  Deactivating the submit button if no items are selected would probably be the simplest way to avoid confusion.  Alternatively you could have the alert messaging inform them that no entries were selected and then keep the modal open rather than closing it.

scott-taubman (changes requested):

src/ui/src/js/controllers/add_command_publishing_blocklist.js
west270: `forEach` does not work with `filteredData` because it does not have the `forEach` function.

PR #1311, opened 2022-05-12
#1299 - Add visual feedback to garden sync actions

### LINKED ISSUE #1299: Improve Garden Sync Feedback
The feedback provided in the UI when performing a garden sync, either at the individual garden level or via a "Sync All", is poor at best and non-existent at worst.  For example:

## Top level garden admin page
* On the top level garden admin page, clicking the "Sync All" button provides no feedback of any kind.  An API call is made, but the UI does not reflect the fact that anything happened and the results, whether successful or an error, are not communicated in any way.  The page has to be reloaded to see any changes in status.

## Individual garden admin page
* On the admin page for a specific garden with a status of "RUNNING", click the "Sync" button.  The Status will update to "INITIALIZING" and then eventually just transition to "ERROR", but visually it's very easy to miss that this change ever takes place and while the garden is in "INITIALIZING" it's not really obvious that anything is happening.  A spinner or some other indicator might help make it clear that we're waiting on the final status of the garden.
* In this same scenario as above, now that the garden is in the "ERROR" state, click the "Sync" button again.  From the behavior of the UI it looks like nothing happens at all.  The status just stays in "ERROR", so the fact that the sync operation was retried is not at all clear.    In reality a new API all is indeed made, there's just no reflection of that fact in the UI.
* There is no surfacing of the ERROR that came back as a result of the "Sync" attempt.  The error that the API returns could actually be useful in some cases, so surfacing that as an alert somewhere on the page would be an improvement.
* If we fix the connection parameters and save the garden config, and then click the "Sync" button, the garden status will update to say "INITIALIZING", but it will never transition out of that state.  Instead, we have to reload the page to see that the sync worked and the garden is now "RUNNING".

jlrpnbbngtn:
Closes #1299 

Added visual feedback in the form of alerts to garden sync and garden connection update on individual garden page. Informational alerts will disappear after 30 s, whereas error alerts must be manually dismissed. The garden management page also has alerts displayed for both success and error associated with the 'sync all' action.

### DIFF
--- a/src/ui/src/js/controllers/admin_garden.js
+++ b/src/ui/src/js/controllers/admin_garden.js
@@ -19,6 +19,7 @@ export default function adminGardenController(
     EventService,
 ) {
   $scope.setWindowTitle('gardens');
+  $scope.alerts = [];
   $scope.gardenCreateSchema = GardenService.CreateSCHEMA;
   $scope.gardenCreateForm = GardenService.CreateFORM;
   $scope.successCallback = function(response) {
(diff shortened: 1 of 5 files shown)


scott-taubman (approved):
Looks good.  Performed as expected.

PR #1312, opened 2022-05-13
#1302 - Role Syncing

### LINKED ISSUE #1302: Allow roles to be synced across gardens

scott-taubman:
Closes #1302 

This PR adds functionality for role syncing across gardens.  The role syncing now happens as part of the user syncing, with the roles being imported on the remote garden first, followed by the users.  This ensures that user permissions are truly consistent across the gardens, rather than leaving open the possibility that users would have the same assigned roles, but that those role assignments might mean different things on different gardens.

## Test Instructions
The easiest way to test is to use the user sync option in the UI.

- Stand up a local and remote garden. Make sure auth is enabled.
- Make some changes to the remote garden so that the users and roles are not in sync with the local garden.
- Use the user sync functionality in the UI to sync the users and roles
- Check the remote garden and ensure that the users and roles are now in sync with the local.


src/app/beer_garden/user.py
```diff
@@ -133,26 +133,51 @@ def user_synced_with_garden(user: User, garden: Garden) -> bool:
...
+
     role_assignments = UserSyncSchema().dump(user).data["role_assignments"]
 
     return role_assignments == remote_user.role_assignments
 
 
-def user_sync_status(user: User) -> dict:
-    """Provides the sync status of the provided User with each remote garden. The
-    resulting dict formatting is:
+def user_sync_status(users: List[User]) -> dict:
```
scott-taubman, line 147: This got reworked to take a list of users rather than an individual one.  This was done mainly because the determination of whether a user is considered synced or not now factors in whether or not all of the roles they are assigned are synced as well.  Since the the status of the roles being synced is identical for all users, I wanted to avoid having to do that check over and over for every individual user.  Instead, I changed this to take a list of users so that I can get the `role_sync_status` for all roles once and only once, no matter how many users we're checking.

src/app/beer_garden/db/mongo/models.py
```diff
@@ -924,14 +929,73 @@ class CommandPublishingBlocklist(Document):
...
     permissions = ListField(field=StringField(), validation=validate_permissions)
 
     meta = {
         "indexes": [{"name": "unique_index", "fields": ["name"], "unique": True}],
     }
 
+    def __str__(self) -> str:
+        return self.name
+
+    def save(self, publish: bool = True, *args, **kwargs):
```
scott-taubman, line 943: I waffle back and forth on whether having the custom save here on the model is preferable, or if I should have mirrored what's done with the User model.  In that case, there is an `update_user` helper function that effectively wraps the save and publishes the event, rather than customizing the save method itself.  The difference there is that there's some special handling for the password field, though really that difference isn't so meaningful.

The benefit to having the custom save here is that you can just interact with the model naturally.  So the usage looks like:

```python
from beer_garden.db.mongo.models import Role
role = Role.objects.first()
role.description = "A truly great role"
role.save()
```

As opposed to for users:
```python
from beer_garden.db.mongo.models import User
from beer_garden.user import update_user
user = User.objects.first()
update_user(user, password="supersecret")
```

The User one requires the extra import and is also maybe less obvious that the save actually happened.

Maybe when in doubt go for consistency should have been the answer.  I figure I can solicit opinions before spending time changing it though.

src/ui/src/templates/sync_users.html
```diff
@@ -1,10 +1,10 @@
 <form ng-submit="sync()">
   <div class="modal-header">
     <h3 class="modal-title" id="modal-title">Sync Users</h3>
-    This will sync all users from this garden to all remote gardens, making the
-    role assignments and password for each user on the remote garden match those
-    of the the users for this garden. Any role assignments that were granted
-    directly on the remote gardens will be overwritten by this process.
+    This will sync all users and roles from this garden to all remote gardens,
+    making the role assignments and password for each user on the remote garden
+    match those of the the users for this garden. Any role assignments that were
```
jlrpnbbngtn, line 6: type *the the*

src/app/beer_garden/db/mongo/models.py
jlrpnbbngtn: I"m sympathetic to the "interact with the model naturally" argument. *A foolish consistency is the hobgoblin of little minds...*

jlrpnbbngtn (approved):
Single typo. Otherwise ready for merge.

PR #1313, opened 2022-05-17
#1301 - Remove queue connection info from system responses

### LINKED ISSUE #1301: Message Queue connection info should be filtered out of Systems API response
If you provide the query parameter `dereference_nested=1` to the /systems API endpoint, among the info that comes back is the message queue connection information, including username and password.  It's unclear why this is the case and almost certainly should not be.  For this issue we need to do the following:

* Identify what, if anything might care about that message queue connection info being in the response
* If there is anything that cares, propose an alternate way of providing that information. The systems endpoint is not the appropriate place for it.
* Update the response so that it doesn't include this info, or at a minimum does not include the especially sensitive bits (the password).

It is possible, maybe even likely, that the code changes for this actually need to happen in the schemas in brewtils.  If that ends up being the case we can open a separate issue there and then reference that issue here as appropriate.

scott-taubman:
Closes #1301 

This PR does the following:

* Removes the `queue_type` and `queue_info` properties from Instances in responses from the /systems endpoints
* Changes the permissions needed to access the /instances endpoints from `SYSTEM_*` to `INSTANCE_*`.

This is done to make it so that the /systems endpoints no longer needlessly provide sensitive message queue connection information.  Meanwhile, access to the /instances endpoints which do include the sensitive connection information (specifically needed by a plugin registering itself with the garden), can be separately provisioned.

**NOTE**: As explained in the docstring for `_remove_queue_info`, the solution applied here is fairly unfortunate, but is by far the least intrusive solution.  Anything more involved is pretty hard to justify in terms of time, effort, and risk.

## Test Instructions
* Verify that the /systems endpoints no longer include the `queue_type` and `queue_info` properties in the instances portion of the response (there are new unit tests that check for this).
* Beer garden should otherwise continue to function as expected, with plugins able to register themselves, be tasked, etc.

### DIFF
--- a/src/app/beer_garden/api/authorization.py
+++ b/src/app/beer_garden/api/authorization.py
@@ -18,6 +18,11 @@ class Permissions(Enum):
     GARDEN_UPDATE = "garden:update"
     GARDEN_DELETE = "garden:delete"
 
+    INSTANCE_CREATE = "instance:create"
+    INSTANCE_READ = "instance:read"
+    INSTANCE_UPDATE = "instance:update"
+    INSTANCE_DELETE = "instance:delete"
+
     QUEUE_CREATE = "queue:create"
     QUEUE_READ = "queue:read"
     QUEUE_UPDATE = "queue:update"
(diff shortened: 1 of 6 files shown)


jlrpnbbngtn (approved):
LGTM

Issue #1300, opened 2022-04-29
garden API response does not include systems and namespaces for the local garden


scott-taubman:
If you access either the garden list endpoint (/api/v1/gardens) or the individual garden endpoint (/api/v1/gardens/{name}), the namespaces and systems for the local garden do not get listed.

At one point, local garden information was moved to being stored in the database to simplify user permissions checks.  At that point in time we opted not to create some mechanism for keeping the garden's systems and namespaces up to date in the garden document itself, with the idea that we would come back and refactor the way garden --> system relationships were tracked across the board.  It is now clear that we are not going to tackle that refactor, but that leaves us still needing to do something to correct the API response for local gardens.

There is a helper function called `local_garden()` in `beer_garden.gardens` that will do the work of populating the system and namespace info for the local garden.  The easiest fix for this in the case of the list endpoint is likely just to update the API method to call `local_garden()` for getting the local garden, change the direct model query to filter out the local garden, and then combine those two sets of results.

So in `GardenListAPI.get` it might look something like:

```python
permitted_gardens = self.permissioned_queryset(Garden, GARDEN_READ)
permitted_gardens_list = list(permitted_gardens.filter(connection_type__ne="LOCAL").no_cache())
_local_garden = local_garden()

if user_has_permission_for_object(self.current_user, GARDEN_READ, _local_garden):
    permitted_gardens_list.append(_local_garden)

for garden in permitted_gardens_list:
   # code continues on as before
```

For the `GardenAPI.get`, we can check if the supplied name matches `config.get("garden.name")` and if so use `local_garden()` to retrieve the info.

PR #1315, opened 2022-05-23
 #1300 - Include the Systems and Namespaces for the local garden

kvig:
Closes https://github.com/beer-garden/beer-garden/issues/1300

This PR does the following:

- Populates the Namespaces and Systems for the local garden when making calls to the /api/v1/gardens endpoint.

Test Instructions:
Testing this requires a beergarden with at least one remote garden. Deploy plugins to both gardens.
- Check the Admin->Gardens page, Verify that the Namespaces and Systems isn't 0 for the local garden.
- Run the following command ```curl localhost:2337/api/v1/gardens | jq -S '.'```. Check that there is an entry for ```connection_type: LOCAL``` with systems and namespaces.


src/app/test/api/http/unit/handlers/v1/garden_test.py
```diff
@@ -360,6 +382,8 @@ def test_auth_enabled_returns_permitted_gardens(
         assert len(response_body) == 1
         assert response_body[0]["id"] == str(garden_permitted.id)
         assert "connection_params" in response_body[0].keys()
+        assert len(response_body[0]["systems"]) == 1
+        assert response_body[0]["systems"][0]["name"] == system.name
```
scott-taubman, line 386: These checks are getting a little unwieldy, but may as well also check the namespaces, since that was the other data point that was missing before this change.

scott-taubman (changes requested):
What's here looks good.  I think we also want to update the `GardenAPI.get` method.  There's a suggestion at the bottom of the issue description on how to possibly approach that.

src/app/beer_garden/api/http/handlers/v1/garden.py
```diff
@@ -189,7 +189,17 @@ async def get(self):
         permitted_gardens = self.permissioned_queryset(Garden, GARDEN_READ)
         response_gardens = []
 
-        for garden in permitted_gardens.no_cache():
+        permitted_gardens_list = list(
+            permitted_gardens.filter(connection_type__ne="LOCAL").no_cache()
+        )
+        _local_garden = local_garden()
```
scott-taubman, line 195: I only thought about this while testing it, but it turns out that for remote gardens that themselves have remote gardens (aka a "child" that has "grandchildren"), we list all known systems, not just the local ones.  I'm not particularly fond of that behavior, but the right thing to do here is probably to replicate it with the local garden.  You can do that by instead calling:

```python
_local_garden = local_garden(all_systems=True)
```

That should result in the response including not just the systems of the local garden, but all remote gardens as well.

PR #1322, opened 2022-05-25
issue# 1283 syncs the CommandPublishingBlocklist

### LINKED ISSUE #1283: Add CommandPublishingBlockList data to the garden sync
When a garden sync happens, the CommandPublishingBlockList should be updated in addition to all of the system data that is currently updated.  This would provide an easy means to get things back in sync if the local garden's idea of what is currently blocked somehow gets out of sync with reality on the remote garden.

west270:
Closes #1283 

Syncs the CommandPublishingBlocklist between the child and the parent gardens on garden sync.

Depends on beer-garden/brewtils#399

### DIFF
--- a/src/app/beer_garden/app.py
+++ b/src/app/beer_garden/app.py
@@ -17,6 +17,7 @@ from apscheduler.executors.pool import ThreadPoolExecutor as APThreadPoolExecuto
 from brewtils import EasyClient
 from brewtils.models import Event, Events
 from brewtils.stoppable_thread import StoppableThread
+from mongoengine import DoesNotExist
 from pytz import utc
 
 import beer_garden.api
@@ -29,6 +30,7 @@ import beer_garden.local_plugins.manager
 import beer_garden.namespace
 import beer_garden.queue.api as queue
 import beer_garden.router
+from beer_garden.db.mongo.models import CommandPublishingBlocklist
 from beer_garden.events.handlers import garden_callbacks
 from beer_garden.events.parent_procesors import HttpParentUpdater
 from beer_garden.events.processors import FanoutProcessor, QueueListener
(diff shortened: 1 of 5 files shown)


src/app/beer_garden/app.py
```diff
@@ -152,6 +154,27 @@ def handle_event(event):
                 and event.metadata["entry_point_type"] == "HTTP"
             ):
                 beer_garden.local_plugins.manager.rescan()
+        if (event.name == "COMMAND_PUBLISHING_BLOCKLIST_SYNC") and (
+            event.garden != config.get("garden.name")
+        ):
+            local_command_publishing_blocklist = CommandPublishingBlocklist.objects(
```
scott-taubman, line 160: All of the logic in this if block should really be in a function that lives elsewhere, presumably in `beer_garden.command_publishing_blocklist`.  Similar to how the code above this just called `beer_garden.local_plugins.manager.rescan()`, rather than having all of the rescan logic mashed in here.

scott-taubman: Also though, I don't think this is the right place for this handling to live.  It should be in [`beer_garden.events.handlers`](https://github.com/beer-garden/beer-garden/blob/3.13.0/src/app/beer_garden/events/handlers.py#L37).  In there you would tack on an entry like:

```python
(beer_garden.command_publishing_blocklist.handle_event, "CommandPublishingBlocklist event handler"),
```

And that `handle_event` method would essentially contain your `event.name` and `event.garden` check, followed by this logic.

scott-taubman (changes requested):
Made misc comments, but also be sure to add unit tests for this.

src/app/test/command_publishing_blocklist_test.py
```diff
@@ -0,0 +1,87 @@
...
+        temp_blocked_command.delete()
+        CommandPublishingBlocklist(
+            namespace=garden.name,
+            system=system_name,
+            command="should_get_removed",
+            status="REMOVE_REQUESTED",
+        ).save()
+        handle_event(event)
+
+        blocklist_data = {
```
scott-taubman, line 73: I'm not sure you need to construct this dictionary and do your comparisons against it.  I think something like this is equivalent to everything from this point on, and is a bit easier to follow:

```python
        blocked_commands = CommandPublishingBlocklist.objects.all().values_list(
            "command", "status"
        )
        assert len(blocked_commands) == 2
        assert ("should_get_added", "CONFIRMED") in blocked_commands
        assert ("should_get_removed", "REMOVE_REQUESTED") not in blocked_commands
```

scott-taubman (changes requested):

scott-taubman:
Test failures are due to an unrelated recent change.  I'm going to merge this PR and we'll fix the issue causing the tests to fail separately.

Issue #1295, opened 2022-04-11
Eliminate pathtools dependency


scott-taubman:
[pathtools](https://github.com/gorakhargosh/pathtools) is a library that has not been updated in any meaningful way in ~11 years.  We currently have it as a dependency in only [one spot in the schedule code](https://github.com/beer-garden/beer-garden/blob/3.12.0/src/app/beer_garden/scheduler.py#L27).  The function it serves is ultimately a pretty simple one, so it would be nice if we could just write our own function to meet our ends here and remove the dependency altogether.

If writing our own function somehow turns out to be harder than it appears, we could alternatively pull in the small handful of functions we'd need from pathtools and eliminate the dependency that way.  Pathtools is released under the MIT license, so as long as we properly attribute and include the copyright information, we should be allowed to use it.  Still, I think the first preference should be to duplicate the functionality as simply as possible ourselves.

scott-taubman:
I think this dependency is only a thing because of the file trigger stuff on jobs, which doesn't even work (and never has as far as I can tell).  We should probably solve this by just removing the file trigger code altogether.

PR #1324, opened 2022-05-26
#1295 - Remove file triggers and explicit dependency on pathtools

kvig:
Closes https://github.com/beer-garden/beer-garden/issues/1295

This PR does the following:
- Removes the File Trigger from the UI
- Removes the file/async scheduled task processing from the server.
- Removes the direct beer-garden dependency on pathtools

*Note*: The watchdog lib is still used to load the logging configuration files and to watch for changes, so there is still the watchdog code in the server.

Test Instructions:
- Check the UI, make sure that the File Trigger isn't a trigger type and doesn't have settings to set
- Create a scheduled job, make sure that the request is executed as expected.
- Touch the `plugin.logging.config_file`, make sure the server still reloads the file.

### DIFF
--- a/src/app/beer_garden/db/mongo/models.py
+++ b/src/app/beer_garden/db/mongo/models.py
@@ -74,7 +74,6 @@ __all__ = [
     "DateTrigger",
     "CronTrigger",
     "IntervalTrigger",
-    "FileTrigger",
     "Garden",
     "File",
     "FileChunk",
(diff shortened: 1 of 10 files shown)


src/app/beer_garden/db/mongo/models.py
```diff
@@ -705,27 +704,6 @@ class CronTrigger(MongoModel, EmbeddedDocument):
     jitter = IntField(required=False)
 
 
-class FileTrigger(MongoModel, EmbeddedDocument):
```
scott-taubman, line 708: This probably needs to stay.  By removing the model you end up with an exception getting thrown by the API if you hit the jobs endpoint (for instance, going to the scheduler page) and a job that uses `FileTrigger` is present.  Even though they never worked, it's possible that someone created a job using that trigger and has it sitting around still.  We don't want to break things in that situation.

Alternatively, we could look to add something to this [`initial_setup`](https://github.com/beer-garden/beer-garden/blob/3.13.0/src/app/beer_garden/db/mongo/api.py#L172) call (probably in `ensure_migrations`) that removes the `FileTrigger` jobs from the database.  Doing that should make it safe to remove this mode.

scott-taubman (changes requested):

src/app/beer_garden/db/mongo/models.py
kvig: Added a migration to remove the existing file triggers.

Issue #1294, opened 2022-04-11
Update to newer version of yapconf and watchdog


scott-taubman:
A new release of yapconf (1.0.0) is available that has a [fix for a potential issue](https://github.com/loganasherjones/yapconf/commit/0199d8b232fa50a590aac7dd5cc772efc1f7872c) in python >= 3.9.  I would be good to update to that release, but doing so would require that we also update to a much newer version of the watchdog library.  We are on 0.10.4 of watchdog.  yapconf suggests that for python >= 3.6 watchdog > 1 should be used.  The newest watchdog is currently 2.1.7.

We should evaluate the feasibility of updating to the latest version of watchdog, or at a minimum watchdog > 1.  Assuming it is feasible, we should make the necessary changes and then update the version of yapconf we're using as well.

scott-taubman:
I think this dependency is only a thing because of the file trigger stuff on jobs, which doesn't even work (and never has as far as I can tell).  We should probably solve this by just removing the file trigger code altogether.

PR #1325, opened 2022-05-27
Issue/1294 update yapconf and watchdog

kvig:
Closes https://github.com/beer-garden/beer-garden/issues/1294

This PR upgrades yapconf to 1.0.0 and watchdog to > 2.1.0(currently 2.1.8).

Watchdog is currently only used for re-reading the `plugin.logging.config.file`. The plugins will *not* be updated at this time, but if the plugin is reloaded or a new one is started, it will get the updated configuration without having to restart BeerGarden. 

*Note*: This PR requires https://github.com/beer-garden/beer-garden/pull/1324 to be merged first.

Test Instructions:
- Remove the old pathtools, watchdog and yapconf python packages (or make a new virtualenv)
- Reinstall the deps
- Make sure that BeerGarden still starts as expected
- Touch the local plugin logging config file(may be example_configs/local-plugin-logging.yaml). Verify that a message gets printed indicating the file was modified.

### DIFF
--- a/src/app/setup.py
+++ b/src/app/setup.py
@@ -43,8 +43,8 @@ setup(
         "ruamel.yaml<0.17",
         "stomp.py<6.2.0",
         "tornado<7",
-        "watchdog<1",
-        "yapconf==0.3.7",
+        "watchdog>2.1.0",
+        "yapconf>=1.0.0",
     ],
     classifiers=[
         "Development Status :: 4 - Beta",
--- a/src/app/requirements-dev.txt
+++ b/src/app/requirements-dev.txt
@@ -98,10 +98,6 @@ packaging==20.9
     #   sphinx
 pathspec==0.9.0
     # via black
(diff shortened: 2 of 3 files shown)


src/app/setup.py
```diff
@@ -43,8 +43,8 @@ def find_version(version_file):
         "ruamel.yaml<0.17",
         "stomp.py<6.2.0",
         "tornado<7",
-        "watchdog<1",
-        "yapconf==0.3.7",
+        "watchdog>2.1.0",
+        "yapconf==1.0.0",
```
scott-taubman, line 47: We don't need to be so strict with the version here.  We had it pinned to 0.3.7 before because we discovered 1.0.0 was going to be breaking.  We should loosen this up to be `>=1.0.0` and only pin it in the future if something breaks again.

kvig: Updated

Issue #1309, opened 2022-05-11
Create a default plugin user


scott-taubman:
We should have a separate account for running the plugins by default, rather than defaulting to the admin account.

scott-taubman:
Some notes about how to possibly go about this:

* There's an `ensure_users` (https://github.com/beer-garden/beer-garden/blob/3.13.0/src/app/beer_garden/db/mongo/util.py#L96) function that sets up the default admin user.  We can break the admin creating parts into a `_create_admin` function and then just add another similar one for `_create_pluginrunner`.  (NOTE: A better username than "pluginrunner" would be acceptable).
* The easiest version of this could just assign the "pluginrunner" the same superuser role that the admin gets.  For what we're trying to accomplish I think that would be acceptable.
* Alternatively, we could figure out which permissions an account specifically needs to be able to operate a plugin, create a new internally managed / protected role with those permissions and assign that.  My starting guess at those permission would be: INSTANCE_READ, INSTANCE_UPDATE, REQUEST_CREATE, REQUEST_READ, REQUEST_UPDATE.
* If we go that second route (I think it's probably worth attempting), you'll also want to look at the `ensure_roles`(https://github.com/beer-garden/beer-garden/blob/3.13.0/src/app/beer_garden/db/mongo/util.py#L45) function.  That's where the system managed roles get created.

PR #1327, opened 2022-06-02
#1309 - Create plugin user and role on new garden startup

kvig:
Closes https://github.com/beer-garden/beer-garden/issues/1309

This PR adds steps to the garden initialization process that will create a plugin_admin user, and create a default plugin role with limited permissions necessary to start and run a plugin.

Some notes:

The plugin_admin username and password can be customized via the app config.
The plugin_admin user will not be created if any other users already exist
The plugin role will be created if it does not exist. 

### Test Instructions
- Start a fresh instance of beergarden. 
  - You should see log messages indicating that the plugin_admin user and plugin role are created.
- Without any other setup, you should be able to login to the UI as plugin_admin / password (unless you changed the defaults in your config).
  - You will have limited permissions, for example, you cannot clear all queues in Admin->Systems

### DIFF
--- a/src/app/beer_garden/db/mongo/util.py
+++ b/src/app/beer_garden/db/mongo/util.py
@@ -21,6 +21,18 @@ from beer_garden.role import sync_roles
 
 logger = logging.getLogger(__name__)
 
+PLUGIN_ROLE_PERMISSIONS = [
+    Permissions.GARDEN_READ.value,
+    Permissions.INSTANCE_READ.value,
+    Permissions.INSTANCE_UPDATE.value,
+    Permissions.REQUEST_CREATE.value,
+    Permissions.REQUEST_READ.value,
+    Permissions.REQUEST_UPDATE.value,
+    Permissions.SYSTEM_CREATE.value,
+    Permissions.SYSTEM_READ.value,
+    Permissions.SYSTEM_UPDATE.value,
+]
+
 
 def ensure_local_garden():
     """Creates an entry in the database for the local garden
@@ -45,6 +57,7 @@ def ensure_local_garden():
 def ensure_roles():
     """Create roles if necessary"""
     _configure_superuser_role()
+    _configure_plugin_role()
     _sync_roles_from_role_definition_file()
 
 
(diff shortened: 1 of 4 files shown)


src/app/beer_garden/db/mongo/util.py
```diff
@@ -93,21 +94,72 @@ def _configure_superuser_role():
...
+        logger.info("Creating default plugin user with username: %s", username)
 
-        admin = User(username=username)
-        admin.set_password(password)
-        admin.role_assignments = [
-            RoleAssignment(role=superuser_role, domain={"scope": "Global"})
+        plugin_user = User(username=username)
+        plugin_user.set_password(password)
+        plugin_user.role_assignments = [
+            RoleAssignment(role=plugin_user_role, domain={"scope": "Local"})
```
scott-taubman, line 160: The scope here should be "Global".  The valid scopes are:

* Global
* Garden
* System

Some info about that is documented [here](https://beer-garden.io/docs/app/security/#role_assignments)

src/app/beer_garden/db/mongo/util.py
```diff
@@ -93,21 +94,72 @@ def _configure_superuser_role():
...
+        plugin_role = Role(name="plugin")
+
+    plugin_role.permissions = [
+        Permissions.GARDEN_READ.value,
+        Permissions.INSTANCE_READ.value,
+        Permissions.INSTANCE_UPDATE.value,
+        Permissions.REQUEST_CREATE.value,
+        Permissions.REQUEST_READ.value,
+        Permissions.REQUEST_UPDATE.value,
+        Permissions.SYSTEM_READ.value,
```
scott-taubman, line 112: From my tests it seems like `SYSTEM_CREATE` is needed as well.

src/app/test/db/mongo/util_test.py
```diff
@@ -365,3 +365,15 @@ def test_ensure_roles_creates_superuser_role_if_none_exists(self, monkeypatch):
...
+        """A plugin role with all permissions should be created if none exists"""
+        monkeypatch.setattr(
+            beer_garden.db.mongo.util, "_sync_roles_from_role_definition_file", Mock()
+        )
+
+        assert len(Role.objects.filter(name="plugin")) == 0
+        ensure_roles()
+        plugin_role = Role.objects.get(name="plugin")
+
+        assert len(plugin_role.permissions) == 8  # Roles defined in util.py
```
scott-taubman, line 379: This check feels a little too fragile.  Anytime we change the permissions that are associated with the role we have to come back and touch this test.  Two options:

* Just check that the role got created and call that good.
* Put the list of permissions associated with the role in a constant that can get imported here.  So you'd have something like:

```python
from beer_garden.db.mongo.util import PLUGIN_ROLE_PERMISSIONS
...
...
    assert plugin_role.permissions == PLUGIN_ROLE_PERMISSIONS
```

Then this test continues to pass no matter how that list changes.

scott-taubman (changes requested):

src/app/beer_garden/db/mongo/util.py
```diff
@@ -93,21 +94,72 @@ def _configure_superuser_role():
...
+
+    admin = User(username=username)
+    admin.set_password(password)
+    admin.role_assignments = [
+        RoleAssignment(role=superuser_role, domain={"scope": "Global"})
+    ]
+    admin.save()
+
+
+def _create_plugin_user():
```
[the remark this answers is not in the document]
scott-taubman: I guess you could also follow the pattern with the separate settings for the default account like I have with the admin account, but that's not helping the issue of the overwhelming number of settings

scott-taubman (approved):

Issue #1329, opened 2022-06-02
Remove pymongo dependencies


catattack05:
Remove pymongo from examples like:

https://github.com/beer-garden/beer-garden/blob/develop/src/app/beer_garden/db/mongo/api.py#L17

Specifically, places where we are importing pymongo errors where really we should be importing mongoengine errors.

Additionally, there are instances where we import from `bson`, which is actually provided by pymongo.  We should look to eliminate those instances as well.  Specifically:

* Anywhere that we're doing something like:
  ```python
  from bson import ObjectId
  ...
  thing = ObjectId(mystrid)
  ```
  We should instead by doing:
  ```python
  from mongoengine.fields import ObjectIdField
  ...
  thing = ObjectIdField().to_mongo(mystrid)
  ```
* There is a single usage of `from bson import DBRef`. The [place where that is used](https://github.com/beer-garden/beer-garden/blob/develop/src/app/beer_garden/db/mongo/util.py#L128) is a migration that's fixing an issue from a very specific, fairly old version of beer-garden.  I think it is safe to remove that whole `_ensure_owner_collection_migration()` function at this point, as the likelihood that anyone is specifically running beer-garden >3.0.0, <3.2.0, seems sufficiently small.  We can put a warning in the release notes if we care to.

aks14075032:
@catattack05 I want to work on this issue but I'm facing problem while setting up the project. 

```
(base) apli.ai@192 app % python bin/app.py -c example_configs/config.yaml
2022-06-04 15:57:56,128 - beer_garden.__main__ - INFO - Hi! Please give me just a minute to get set up.
2022-06-04 15:57:56,209 - beer_garden.db.mongo.util - INFO - auth.role_definition_file not defined. No roles will be synced.
2022-06-04 15:57:56,295 - beer_garden.app - INFO - All set! Let me know if you need anything else!
2022-06-04 15:57:56,583 - beer_garden.api.stomp - INFO - Stomp entry point started
2022-06-04 15:57:56,702 - beer_garden.api.http - INFO - Http entry point started
2022-06-04 15:57:56,703 - beer_garden.local_plugins.manager - ERROR - Error scanning plugin path: [Errno 2] No such file or directory: 'plugins'
Traceback (most recent call last):
  File "/Users/apli.ai/Desktop/Personal/Open_Source/beer-garden/src/app/beer_garden/local_plugins/manager.py", line 409, in scan_path
    for path in plugin_paths:
[...]

PR #1331, opened 2022-06-02
#1321 - Properly list python-box and wrapt dependencies

### LINKED ISSUE #1321: Verify all dependencies are properly listed
We've fixed one off missing dependency issues over time, but we should take a pass at all of the imports and make sure we're explicitly listing the right dependencies in the setup.py as needed.

catattack05:
Closes #1321


src/app/requirements.txt
```diff
@@ -16,6 +16,8 @@ backports-zoneinfo==0.2.1
     #   tzlocal
 brewtils==3.14.0
     # via beer-garden (setup.py)
+bson==0.5.10
```
scott-taubman, line 19: I was confused by this one.  We definitely have direct imports of bson, but the fact that this is an addition to the requirement.txt and not a change indicates that it wasn't actually being installed before.  I dug into it a bit and it looks like the bson that we are using was coming from pymongo.  The [pymongo pypi page](https://pypi.org/project/pymongo/) specifically has this warning:

> Do not install the “bson” package from pypi. PyMongo comes with its own bson package; doing “easy_install bson” installs a third-party package that is incompatible with PyMongo.

What that means for this PR is that I think we want to actually take bson out here and then figure out what to do with pymongo as part of #1329.  It may be the case that our use of bson means we have the direct pymongo dependency after all, even though that relationship isn't obvious.

scott-taubman:
I ran into some other issues when trying to install the dev dependencies that alerted me to another change that I think should happen in this PR.  In the `requirements-dev.in` we list two sphinx related documentation dependencies.  Those aren't actually needed anymore and are causing a conflict with the brewtils dependencies, so we should just get rid of them.

scott-taubman:
I ran into some other issues when trying to install the dev dependencies that alerted me to another change that I think should happen in this PR.  In the `requirements-dev.in` we list two sphinx related documentation dependencies.  Those aren't actually needed anymore and are causing a conflict with the brewtils dependencies, so we should just get rid of them.

Issue #1314, opened 2022-05-18
Polish the commandpublishingblocklist API


scott-taubman:
Some observations when using the commandpublishingblocklist API endpoint:

* Adding a CPBL entry for a garden that is not properly configured resulted in a 400.  This isn't really the right response because the data provided to the commandpublishingblocklist endpoint was correct.  The issue is that the remote garden isn't available.  503 seems like an appropriate response here, along with a message indicating that the remote garden could not be reached.  The same is true for other operations, like DELETE.
* In the above scenario, the CPBL entry still gets created, which is problematic.  It means that when the remote garden is reachable and I want to try to add the CPBL entry again, I get a 409 response because the entry already exists.  We should not create the CPBL entry until the request to the remote garden has been successfully routed.
* Posting an empty list as the creation input results in a 201 (created) response, which is pretty confusing.  It also leads to UI confusion where you can open the "Add" modal, select nothing and click "Submit", and you get the pop-up saying that the blocklist has been successfully updated.  An empty list should fail validation and return a 400.
* Adding or deleting an entry for a grandchild garden does not work.  It seemingly tries to route directly and fails rather than going through the child garden.

scott-taubman:
Regarding the issue of the entry being created locally even if it never got routed to the remote garden, in `beer_garden/command_publishing_blocklist.py:command_publishing_blocklist_add` the [save()](https://github.com/beer-garden/beer-garden/blob/develop/src/app/beer_garden/command_publishing_blocklist.py#L21) should probably come after the route call. That way, if the routing of the operation fails, we don't even create a local entry. This would require a change where the id of the blocklist entry isn't sent as part of the command in the Operation. I'm not sure of the implications of this in the current implementation.

scott-taubman:
The logic for this probably doesn't belong in the API, but when a blocklist entry is added for the local garden, the status should be populated with `CONFIRMED` (or whatever we decided the final state should be). I'm not sure if the plan was for that to get handled with the event processing work, but if it's not being done there we should tackle it here.

scott-taubman:
> The logic for this probably doesn't belong in the API, but when a blocklist entry is added for the local garden, the status should be populated with `CONFIRMED` (or whatever we decided the final state should be). I'm not sure if the plan was for that to get handled with the event processing work, but if it's not being done there we should tackle it here.

It looks like this particular thing will be addressed as part of #1283

PR #1332, opened 2022-06-03
Issue/1310 Added Events to Update Blocklist

### LINKED ISSUE #1310: Publish and process CommandPublishingBlocklist update events
When a CommandPublishingBlocklist entry is created, updated, or deleted an event needs to be published so that upstream gardens can be aware of the change.  Create and update can probably share an event type since we do not do anything to differentiate between update and creation upstream.  Delete would be its own event type.

Along with the publishing, we'll want to add the event processing for these new event types, which would result in the entries on the upstream garden's CommandPublishingBlocklist getting updated as appropriate.

west270:
Closes #1310

Added event publishing to publish event when blocklist is has an entry removed or updated, and also process the event to keep blocklist between parent and child synced.

### DIFF
--- a/src/app/beer_garden/app.py
+++ b/src/app/beer_garden/app.py
@@ -17,7 +17,6 @@ from apscheduler.executors.pool import ThreadPoolExecutor as APThreadPoolExecuto
 from brewtils import EasyClient
 from brewtils.models import Event, Events
 from brewtils.stoppable_thread import StoppableThread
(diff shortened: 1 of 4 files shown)


scott-taubman:
Per discussion I think we'll just combine these changes here with the branch for #1314 since there was going to be some necessary conflict resolution.  Closing this PR.

PR #1336, opened 2022-06-06
Issue #1329

### LINKED ISSUE #1329: Remove pymongo dependencies

catattack05:
Closes #1329 

This will probably need someone to seriously check out for any unintended side effects (plus bson was present on a number of different files so there's a lot of files that had only a few lines changed)

### DIFF
--- a/src/app/beer_garden/api/http/base_handler.py
+++ b/src/app/beer_garden/api/http/base_handler.py
@@ -20,7 +20,9 @@ from marshmallow import Schema
 from marshmallow.exceptions import ValidationError as MarshmallowValidationError
 from mongoengine.errors import DoesNotExist, NotUniqueError
 from mongoengine.errors import ValidationError as MongoValidationError
-from pymongo.errors import DocumentTooLarge
+from mongoengine.errors import InvalidDocumentError
+
+# from pymongo.errors import DocumentTooLarge
 from tornado.web import HTTPError, RequestHandler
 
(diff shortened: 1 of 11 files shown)


catattack05:
...and I also realized I included the pika commits in there, so whoops.

PR #1337, opened 2022-06-06
Issue #1329

### LINKED ISSUE #1329: Remove pymongo dependencies

catattack05:
Closes #1329 

Will probably need some serious examination for unintended consequences

### DIFF
--- a/src/app/beer_garden/api/http/base_handler.py
+++ b/src/app/beer_garden/api/http/base_handler.py
@@ -18,7 +18,7 @@ from brewtils.errors import (
 )
 from marshmallow import Schema
 from marshmallow.exceptions import ValidationError as MarshmallowValidationError
-from mongoengine.errors import DoesNotExist, InvalidDocumentError, NotUniqueError
+from mongoengine.errors import DoesNotExist, NotUniqueError
 from mongoengine.errors import ValidationError as MongoValidationError
 from tornado.web import HTTPError, RequestHandler
 
--- a/src/app/beer_garden/api/http/base_handler.py
+++ b/src/app/beer_garden/api/http/base_handler.py
@@ -20,7 +20,9 @@ from marshmallow import Schema
 from marshmallow.exceptions import ValidationError as MarshmallowValidationError
 from mongoengine.errors import DoesNotExist, NotUniqueError
 from mongoengine.errors import ValidationError as MongoValidationError
-from pymongo.errors import DocumentTooLarge
+from mongoengine.errors import InvalidDocumentError
+
+# from pymongo.errors import DocumentTooLarge
 from tornado.web import HTTPError, RequestHandler
 
 import beer_garden.api.http
@@ -70,7 +72,7 @@ class BaseHandler(RequestHandler):
         NotUniqueException: {"status_code": 409},
         NotUniqueError: {"status_code": 409, "message": "Resource already exists"},
         EndpointRemovedException: {"status_code": 410, "message": "Endpoint removed"},
-        DocumentTooLarge: {"status_code": 413, "message": "Resource too large"},
+        InvalidDocumentError: {"status_code": 413, "message": "Resource too large"},
         RequestPublishException: {"status_code": 502},
         RoutingException: {"status_code": 500},
         socket.timeout: {"status_code": 504, "message": "Backend request timed out"},
--- a/src/app/beer_garden/api/http/base_handler.py
+++ b/src/app/beer_garden/api/http/base_handler.py
@@ -21,8 +21,6 @@ from marshmallow.exceptions import ValidationError as MarshmallowValidationError
 from mongoengine.errors import DoesNotExist, NotUniqueError
 from mongoengine.errors import ValidationError as MongoValidationError
 from mongoengine.errors import InvalidDocumentError
-
-# from pymongo.errors import DocumentTooLarge
 from tornado.web import HTTPError, RequestHandler
 
 import beer_garden.api.http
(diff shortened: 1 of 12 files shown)


src/app/beer_garden/db/mongo/api.py
```diff
@@ -141,11 +141,12 @@ def check_connection(db_config: Box):
         # The 'connect' method won't actually fail
         # An exception won't be raised until we actually try to do something
         logger.debug("Attempting connection")
-        conn.server_info()
+        # conn.server_info()
```
scott-taubman, line 144: Two things:

* In general, leaving commented out lines of code isn't desirable.  We've got the git history, so if we need to resurrect something we'd do it that way.
* We need *something* to happen here.  The `conn.server_info()` call is an arbitrary one that's just a proxy for checking that we can connect to the database.  The easiest thing is probably to keep it and just break the try except into two separate blocks:

```python
    try:
        # Set timeouts here to a low value - we don't want to wait 30
        # seconds if there's no database
        conn = connect(
            alias="aliveness",
            db=db_config["name"],
            socketTimeoutMS=1000,
            serverSelectionTimeoutMS=1000,
            **db_config["connection"],
        )
    except ConnectionFailure as ex:
        logger.debug(ex)
        return False

    try:
        # The 'connect' method won't actually fail
        # An exception won't be raised until we actually try to do something
        logger.debug("Attempting connection")
        conn.server_info()
    except Exception as ex:
        logger.debug(ex)
        return False

    return True
```

The reason to do this is that you don't want to catch *all* exceptions for the `connect` call.  For instance, if someone changed the structure of `db_config` such that it no longer has a `name`, you don't want the `KeyError` that would get raised to be caught.

Catching all exceptions for the `conn.server_info()` call isn't ideal either, but by the time we get there all of the input has already been provided and it largely doesn't matter what the exception might be.

src/app/beer_garden/db/mongo/util.py
```diff
@@ -266,10 +211,10 @@ def check_indexes(document_class):
         existing = document_class._get_collection().index_information()
 
         if document_class == Request and "parent_instance_index" in existing:
-            raise OperationFailure("Old Request index found, rebuilding")
+            raise OperationError("Old Request index found, rebuilding")
```
scott-taubman, line 214: Not your fault here, but the idea that we explicitly raise exceptions from another package seems weird to me.  Catching and re-raising, sure.  But raising those exceptions ourselves doesn't seem appropriate.  This really ought to be a custom exception defined in `beer_garden.errors`, or alternatively this could be refactored so that the conditional checks control the code flow, rather than the try / except.

src/app/beer_garden/db/mongo/util.py
```diff
@@ -297,7 +242,7 @@ def check_indexes(document_class):
             db = get_db()
             db[document_class.__name__.lower()].drop_indexes()
             logger.warning("Dropped indexes for %s collection", document_class.__name__)
-        except OperationFailure:
+        except OperationError:
```
scott-taubman, line 245: This is probably true of the earlier `ensure_indexes()` call too, but switching from `pymongo.errors.OperationFailure` to `mongoengine.errors.OperationError` here is not going to work.  The `drop_indexes()` call is a pymongo one, so there's no situation where it's going to raise a mongoengine errror.

This might be the point where the goal of removing the `pymongo` dependency completely becomes a fool's errand.  Mongoengine doesn't expose a way drop indexes and instead provides these mechanism to get at the pymongo objects do it through those.

Others may have opinions or ideas, but we may need to accept the reality that using mongoengine means interacting with pymongo directly for some things.

src/app/beer_garden/db/mongo/util.py
catattack05: I can add a new error to our custom ones if that works. I sort of assume that goes in `brewtils.errors` unless there's an error file in the beergarden directory I'm not seeing.

src/app/beer_garden/api/http/base_handler.py
```diff
@@ -70,7 +69,7 @@ class BaseHandler(RequestHandler):
         NotUniqueException: {"status_code": 409},
         NotUniqueError: {"status_code": 409, "message": "Resource already exists"},
         EndpointRemovedException: {"status_code": 410, "message": "Endpoint removed"},
-        DocumentTooLarge: {"status_code": 413, "message": "Resource too large"},
```
scott-taubman, line 73: I might just remove this one to be honest.  I don't think `InvalidDocumentError` is a drop in replacement here and neither of these is actually what gets raised in the scenario where you call `save()` on an overly large document.

src/app/beer_garden/db/mongo/util.py
```diff
@@ -376,7 +321,7 @@ def check_indexes(document_class):
         try:
             document_class.ensure_indexes()
             logger.warning("%s indexes rebuilt successfully", document_class.__name__)
-        except OperationFailure:
+        except IndexOperationError:
```
scott-taubman, line 324: You don't want to change this one.  `ensure_indexes()` isn't going to raise the new `IndexOperationError` that you've introduced.

scott-taubman (changes requested):

PR #1340, opened 2022-06-21
issue #1338 - large output override

### LINKED ISSUE #1338: Provide an opt-in for viewing large request output
Currently, there is a [check on the request view page](https://github.com/beer-garden/beer-garden/blob/3.13.0/src/ui/src/js/controllers/request_view.js#L226) against the size of the output.  After a certain threshold, the output no longer gets rendered and instead we direct the user to download the output instead.

Realistically, the threshold for what could reasonably be rendered is going to depend on the resources available to the user's browser.  There may be cases where beer garden opts not to render results that may actually be viable to render for a particular user.  It would be nice to provide an opt-in to the output rendering in such a case, which would let the user essentially ignore the "output is too large" error and tell the UI to just render it anyway.

jessica-obrien:
Closes #1338 

User can click error pop-up or icon to choose to display large outputs. Added warning in pop-up text to do so at your own risk. Consolidated size checks for better handling.

### DIFF
--- a/src/ui/src/js/controllers/request_view.js
+++ b/src/ui/src/js/controllers/request_view.js
@@ -225,7 +225,7 @@ export default function requestViewController(
           ' crash the page. Downloading File might take a minute for UI to prepare. To ' +
           'display the output anyway, <i>click here</i>. <b>NOTE:</b> Displaying large ' +
           'request output could result in your browser crashing or hanging indefinitely.';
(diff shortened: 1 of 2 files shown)


jlrpnbbngtn (approved):
LGTM. Produces the desired behavior whenever
1. the return is JSON with too many nodes, or
2. the return is any type but is too large.

This is the correct approach, IMO.

PR #1344, opened 2022-07-12
Issue 249: Tests written and passing

### LINKED ISSUE #249: Mark request as cancelled
It would be nice to have something in the UI that would allow us to mark a request as CANCELLED.

Note that (and we'd have to make this explicit in the UI) this doesn't mean actually cancelling a request, but rather marking one as cancelled.

Right now we do everything possible to prevent a request from being marked as IN_PROGRESS but not existing on a queue anywhere, but sometimes it happens. In these unrecoverable cases where an admin KNOWS the request will never "finish" it would be useful to be able to mark as CANCELLED (or something) to prevent that request from contaminating the metrics forever.

catattack05:
Closes #249 
This PR gives the MongoPruner the ability to mark any requests that have been sitting in CREATED or IN_PROGRESS for longer than a specified amount of time as cancelled. There is now a setting in the configs called in_progress that holds that value

### DIFF
--- a/src/app/beer_garden/app.py
+++ b/src/app/beer_garden/app.py
@@ -82,11 +82,16 @@ class Application(StoppableThread):
         ]
 
         # Only want to run the MongoPruner if it would do anything
-        tasks, run_every = db.prune_tasks(**config.get("db.ttl"))
+        ttl_config = config.get("db.ttl")
+
+        tasks, run_every = db.prune_tasks(**ttl_config)
         if run_every:
             self.helper_threads.append(
                 HelperThread(
-                    db.get_pruner(), tasks=tasks, run_every=timedelta(minutes=run_every)
+                    db.get_pruner(),
+                    tasks=tasks,
+                    run_every=timedelta(minutes=run_every),
+                    cancel_threshold=ttl_config.in_progress,
                 )
             )
 
--- a/src/app/beer_garden/config.py
+++ b/src/app/beer_garden/config.py
@@ -757,6 +757,15 @@ _DB_SPEC = {
                     "previous_names": ["info_request_ttl"],
                     "alt_env_names": ["INFO_REQUEST_TTL"],
                 },
+                "in_progress": {
+                    "type": "int",
+                    "default": -1,
+                    "description": (
+                        "Number of minutes to wait for a request in CREATED or IN_PROGRESS"
+                        "to complete before considering timed out and marking as CANCELLED"
+                        "(negative number for never)"
+                    ),
+                },
                 "file": {
                     "type": "int",
                     "default": 15,
--- a/src/app/beer_garden/db/mongo/pruner.py
+++ b/src/app/beer_garden/db/mongo/pruner.py
@@ -10,17 +10,30 @@ from beer_garden.db.mongo.models import File, RawFile, Request
 
 
 class MongoPruner(StoppableThread):
-    def __init__(self, tasks=None, run_every=None):
+    """
+    Class to prune Mongo DB canceled or timed out Request objects.
+
+    Attributes:
+        logger: Logger to run debug messages
+        display_name: The name of the pruner
+        run_every: How often to run the pruner in minutes
+        tasks: List of tasks to prune
+        cancel_threshold: Time after which outstanding requests are marked as canceled in minutes
+    """
+
+    def __init__(self, tasks=None, run_every=None, cancel_threshold=None):
         self.logger = logging.getLogger(__name__)
         self.display_name = "Mongo Pruner"
         self._run_every = (run_every or timedelta(minutes=15)).total_seconds()
         self._tasks = tasks or []
+        self._cancel_threshold = cancel_threshold or -1
 
         super(MongoPruner, self).__init__(logger=self.logger, name="Remover")
 
     def add_task(
         self, collection=None, field=None, delete_after=None, additional_query=None
     ):
+        """Adds tasks to pruner object"""
         self._tasks.append(
             {
                 "collection": collection,
@@ -30,7 +43,23 @@ class MongoPruner(StoppableThread):
             }
         )
 
+    def _cancel_outstanding(self):
+        """
+        Helper function for run to mark requests still outstanding after a certain
+        amount of time as canceled.
+        """
+
+        timeout = datetime.utcnow() - timedelta(minutes=self._cancel_threshold)
+        outstanding_requests = Request.objects.filter(
+            status__in=["IN_PROGRESS", "CREATED"], created_at__lte=timeout
+        )
+        for request in outstanding_requests:
+            request.status = "CANCELED"
+            request.save()
+
     def run(self):
+        """Runs the pruner to delete tasks"""
+
         self.logger.debug(
             "%s is started with run interval %ss", self.display_name, self._run_every
         )
@@ -50,6 +79,9 @@ class MongoPruner(StoppableThread):
                     % (task["collection"].__name__, str(delete_older_than))
                 )
                 task["collection"].objects(query).no_cache().delete()
+
+            if self._cancel_threshold > 0:
+                self._cancel_outstanding()
 
         self.logger.debug(self.display_name + " is stopped")
 
--- a/src/app/example_configs/config.yaml
+++ b/src/app/example_configs/config.yaml
@@ -24,6 +24,7 @@ db:
   ttl:
     action: -1
     file: 15
+    in_progress: 15
     info: 15
 entry:
   http:
--- a/src/app/test/db/mongo/pruner_test.py
+++ b/src/app/test/db/mongo/pruner_test.py
@@ -1,11 +1,15 @@
 # -*- coding: utf-8 -*-
+import datetime
 from datetime import timedelta
 
 import pytest
 from mock import MagicMock, Mock, patch
+from mongomock.gridfs import enable_gridfs_integration
 
 from beer_garden.db.mongo.models import File, RawFile, Request
 from beer_garden.db.mongo.pruner import MongoPruner
+
+enable_gridfs_integration()
 
 
 @pytest.fixture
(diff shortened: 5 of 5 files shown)


src/app/example_configs/config.yaml
```diff
@@ -1,19 +1,19 @@
 auth:
-  enabled: false
```
scott-taubman, line 2: A bunch of stuff moved around here, I'm guessing because you are using the example config directly in your environment and it got rewritten when the application started.  You should clean this up so that the only diff is the thing that actually changed.

src/app/beer_garden/db/mongo/pruner.py
```diff
@@ -5,16 +5,18 @@
...
 from brewtils.stoppable_thread import StoppableThread
 from mongoengine import Q
+from mongoengine.queryset.visitor import Q as Q2
 
 from beer_garden.db.mongo.models import File, RawFile, Request
 
 
 class MongoPruner(StoppableThread):
-    def __init__(self, tasks=None, run_every=None):
+    def __init__(self, tasks=None, run_every=None, ttl=None):
```
scott-taubman, line 14: `ttl` is a bit too generic in this context.  Maybe call it `cancel_threshold` or something?

Also, the name becomes marginally less critical if there was a proper docstring for the class explaining what the attributes are.  Let's take the opportunity here and write up a docstring for this class so that you can document the new one.  It looks like we have literally never written a docstring for a class in the code base (shame on us), but for the other places where docstrings exist we strive to follow the google style guide.  See the section on class docstrings [here](https://google.github.io/styleguide/pyguide.html#384-classes).

src/app/beer_garden/db/mongo/pruner.py
```diff
@@ -30,7 +32,17 @@ def add_task(
             }
         )
 
+    def cancel_in_progress(self):
```
scott-taubman, line 35: Since this deals in both "IN_PROGRESS" and "CREATED" requests, let's rename this to `cancel_oustanding_requests`.  Similarly, the variable you have named `in_progress` below would become `outstanding_requests`.

Also, let's add a basic docstring describing the method.

src/app/beer_garden/db/mongo/pruner.py
```diff
@@ -5,16 +5,18 @@
 
 from brewtils.stoppable_thread import StoppableThread
 from mongoengine import Q
+from mongoengine.queryset.visitor import Q as Q2
```
scott-taubman, line 8: The `Q` that you're importing as `Q2` here is the same `Q` as the line above.  So there's no need to import it twice.  I don't think you need Q at all though, which I will explain below.

src/app/beer_garden/db/mongo/pruner.py
```diff
@@ -30,7 +32,17 @@ def add_task(
             }
         )
 
+    def cancel_in_progress(self):
+        in_progress = Request.objects(Q2(status="IN_PROGRESS") | Q2(status="CREATED"))
```
scott-taubman, line 36: You don't need to use `Q` to construct this query. You can just do:

`Request.objects.filter(status__in=["IN_PROGRESS", "CREATED"])`

src/app/beer_garden/db/mongo/pruner.py
```diff
@@ -30,7 +32,17 @@ def add_task(
...
+        current_time = datetime.utcnow()
+        for request in in_progress:
+            timeout = request.created_at + timedelta(minutes=self._ttl)
+            if current_time >= timeout:
```
scott-taubman, line 40: You should be able to tack on the timeout check as criteria when you're querying for the Requests.  That is, rather than getting **all** "IN_PROGRESS" and "CREATED" requests and then checking them in application space to determine if they've timed out, just query for the ones that have timed out.  Then you could iterate over them, or possibly just do a single bulk update.

src/app/beer_garden/db/mongo/pruner.py
```diff
@@ -51,6 +63,8 @@ def run(self):
                 )
                 task["collection"].objects(query).no_cache().delete()
 
+            MongoPruner.cancel_in_progress(self)
```
scott-taubman, line 66: This looks weird.  I'm assuming this is supposed to be `self.cancel_in_progress()`?

Also, worth mentioning is that by convention, if a method is intended for internal use only (you don't want / expect it to be called outside of the MongoPruner's own code), it should start with an underscore.  If that's the case (looks like it based on the usage here) and taking into account my suggested name elsewhere, then this should possibly be:

```python
self._cancel_outstanding_requests()
```

src/app/test/db/mongo/pruner_test.py
```diff
@@ -36,6 +38,60 @@ def test_prune_something(self, pruner, collection_mock):
         pruner.run()
         assert collection_mock.objects.return_value.no_cache.return_value.delete.called
 
+    def test_in_progress_ttl_index(self):
```
scott-taubman, line 41: What is this testing for exactly?

src/app/test/db/mongo/pruner_test.py
```diff
@@ -36,6 +38,60 @@ def test_prune_something(self, pruner, collection_mock):
...
+        timeout = config._DB_SPEC["items"]["ttl"]["items"]["in_progress"].get("default")
+        test_pruner = MongoPruner(tasks=[task], ttl=timeout)
+        assert test_pruner._ttl == 15
+
+    def test_created_to_canceled(self, pruner):
+        test = Request(
```
scott-taubman, line 47: We do have some safeguards to cleanup test data, but you should still explicitly clean it up in your individual tests.  Instead of saving the Request here in your test, you'll want to create a fixture and call that.  Done properly, the fixture will clean up your data for you when the test is done.  There are lots of examples of this in other tests.

src/app/test/db/mongo/pruner_test.py
```diff
@@ -36,6 +38,60 @@ def test_prune_something(self, pruner, collection_mock):
...
+            system_version="T",
+            instance_name="T",
+            namespace="T",
+            command="T",
+            created_at=datetime.datetime(2020, 5, 17),
+            status="CREATED",
+        )
+        test.save()
+        req_id = test.id
+        pruner.cancel_in_progress()
```
scott-taubman, line 58: I don't feel super strongly about this, but if `cancel_in_progress` is better suited to being an internal function (`_cancel_outstanding_requests`) then the function that we primarily want to be testing is the `run` function.  `run` is the public interface by which `_cancel_outstanding_requests` would be called, so that's the behavior you want to verify.  Said another way, confirming `_cancel_outstanding_requests` does what it's supposed to doesn't help you if the thing that actually gets called (`run`) never calls it.

More succinctly, write your tests so that you call `run` and verify that you get the intended behavior.

src/app/beer_garden/db/mongo/pruner.py
```diff
@@ -30,7 +43,22 @@ def add_task(
             }
         )
 
+    def _cancel_outstanding(self):
+        """
+        Helper function for run to mark requests still outstanding after a certain
+        amount of time as canceled.
+        """
+        timeout = datetime.utcnow() - timedelta(minutes=self._cancel_threshold)
+        outstanding_tasks = Request.objects.filter(
```
scott-taubman, line 52: `outstanding_tasks` --> `outstanding_requests`

src/app/test/db/mongo/pruner_test.py
```diff
@@ -36,6 +72,14 @@ def test_prune_something(self, pruner, collection_mock):
         pruner.run()
         assert collection_mock.objects.return_value.no_cache.return_value.delete.called
 
+    def test_run_deletes_outstanding_tasks(self, pruner, in_progress, created):
```
scott-taubman, line 75: This test name is misleading, since the desired behavior is that it cancels rather than deletes.  Also, they are "requests", rather than "tasks".

src/app/beer_garden/db/mongo/pruner.py
```diff
@@ -10,17 +10,30 @@
...
-    def __init__(self, tasks=None, run_every=None):
+    """
+    Class to prune Mongo DB canceled or timed out Request objects.
+
+    Attributes:
+        logger: Logger to run debug messages
+        display_name: The name of the pruner
+        run_every: How often to run the pruner
+        tasks: List of tasks to prune
+        cancel_threshold: Time after which outstanding requests are marked as canceled
```
scott-taubman, line 21: Should specify the units here (minutes).  Also, the config description mentions that a negative number means never, which is what the behavior should be, but I'm not sure that's properly accounted for in this code.  Setting cancel_threshold to `None` when initializing `MongoPruner` seems like it would just make things blow up and I think you need handling of the negative number option in `app.py` where you're reading the "db.ttl" setting.

It would be good to write an additional unit test for that (the `canceled_threshold=None` / `db.ttl.in_progress < 0` scenario).

src/app/beer_garden/app.py
```diff
@@ -82,11 +82,17 @@ def initialize(self):
         ]
 
         # Only want to run the MongoPruner if it would do anything
+        cancel_after = config._DB_SPEC["items"]["ttl"]["items"]["in_progress"].get(
```
scott-taubman, line 85: Meant to comment on this last time and forget to apparently.  The way you want to interact with the config is something like:

```python
cancel_after = config.get("db.ttl.in_progress")
```

You don't want to reference the specs directly and also what you're doing here seems to be getting the default value, as opposed to the value that the user actually set.

Also, I think we talked about at some point the fact that the entire "db.ttl" config section is retrieved immediately after this, which means you could slightly refactor so that you get that once and then pull the "in_progress" part of it for use with `cancel_after`.

src/app/beer_garden/config.py
```diff
@@ -757,6 +757,15 @@ def _parse_args(args: Sequence[str]) -> Tuple[YapconfSpec, dict]:
                     "previous_names": ["info_request_ttl"],
                     "alt_env_names": ["INFO_REQUEST_TTL"],
                 },
+                "in_progress": {
+                    "type": "int",
+                    "default": 15,
```
scott-taubman, line 762: Another thing I missed last time: this should default to `-1`.  That way the default behavior is what people were already used to (we don't cancel outstanding requests) and they have to opt into this new behavior by explicitly setting this.

src/app/beer_garden/db/mongo/pruner.py
```diff
@@ -30,7 +43,24 @@ def add_task(
             }
         )
 
+    def _cancel_outstanding(self):
+        """
+        Helper function for run to mark requests still outstanding after a certain
+        amount of time as canceled.
+        """
+        if isinstance(self._cancel_threshold, int):
```
scott-taubman, line 51: There's something a little bit awkward about having this check here.  I think there are a few options that would feel more natural.  An easy one might just be to set `_cance_outstanding` to '-1' if it's provided as `None`.  So in the `__init__`, instead of:

```python
self._cancel_threshold = cancel_threshold
```
you would do

```python
self._cancel_threshold = cancel_threshold or -1
```
Then you can skip the ` isinstance(self._cancel_threshold, int)` check entirely.

Another super minor thing, but I'd also move the `self._cancel_threshold > 0` check to where you are making the call to this function, rather than inside of it.  The reason for that being that someone reading or stepping through the code has to jump into this function to figure out that it's a no-op, as where if you do the check before the function call, it's clear right up front.

src/app/beer_garden/app.py
```diff
@@ -82,11 +82,16 @@ def initialize(self):
         ]
 
         # Only want to run the MongoPruner if it would do anything
+        cancel_after = config.get("db.ttl.in_progress")
```
scott-taubman, line 85: I think I commented about not doing the `config.get` twice in a row here.  Just get the "db.ttl" data once, then use it in the two places you need it.

src/app/beer_garden/app.py
scott-taubman: No, something is not setup right there in however you are trying to do this.  You will pretty much never get `None` back for those values because there are defaults defined.  So even if there was nothing in your custom config, the `get` would return the defaults.

In any case, it works something like this:

```python
In [2]: ttl_config = beer_garden.config.get("db.ttl")

In [3]: ttl_config
Out[3]: <Box: {'action': -1, 'info': 15, 'in_progress': 1, 'file': 15}>

In [4]: ttl_config.in_progress
Out[4]: 1
```

PR #1346, opened 2022-07-27
#1343 - Remove redundant system call from frontend

### LINKED ISSUE #1343: UI gardens call makes systems call redundant
In 3.11.0, the UI started making a call to the /gardens endpoint in order to collect info needed to do various permissions checks.  While that call included the systems data for remote gardens, the local garden is handled differently and no systems data was included.

As of 3.14.0, the systems data is now included for the local garden in the /gardens endpoint response.  This makes the /systems call to populate the system data in the UI completely redundant.  That data could instead be populated from the data that's returned from the /gardens call.  We should look to do that or find some other means of reducing the redundancy.

west270:
Closes #1343

Removes redundant system call and sets the system data when it makes the garden call.

### DIFF
--- a/src/ui/src/js/configs/routes.js
+++ b/src/ui/src/js/configs/routes.js
@@ -50,30 +50,22 @@ export default function routeConfig(
           ],
           systems: [
             '$rootScope',
-            'SystemService',
-            ($rootScope, SystemService) => {
-              return SystemService.getSystems().then(
-                  (response) => {
-                    $rootScope.sysResponse = response;
-                    $rootScope.systems = response.data;
-                  },
-                  (response) => {
-                    $rootScope.sysResponse = response;
-                    $rootScope.systems = [];
-                  },
-              );
-            },
-          ],
-          gardens: [
-            '$rootScope',
             'GardenService',
             ($rootScope, GardenService) => {
               return GardenService.getGardens().then(
                   (response) => {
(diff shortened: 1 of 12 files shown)


src/ui/src/js/configs/routes.js
```diff
@@ -49,31 +49,23 @@ export default function routeConfig(
...
             'GardenService',
             ($rootScope, GardenService) => {
               return GardenService.getGardens().then(
                   (response) => {
                     $rootScope.gardens = response.data;
+                    const localGarden = $rootScope.gardens.find(
+                        (item) => item.connection_type === 'LOCAL',
+                    );
+                    $rootScope.systems = localGarden.systems;
+                    response.data = localGarden.systems;
```
scott-taubman, line 62: I think these changes work, but the stuff going on here to populate `$rootScope.sysResponse` seems kinda sketchy.  I would try to track down what's using `sysResponse` and see if it's feasibly to make it just work with `rootScope.systems` rather than the response object.

If it's really impossible to detangle these things (I don't think it should be), then we at least need a comment here explaining why we're manipulating the response in this manner.

scott-taubman (changes requested):
Appears to achieve what we want, but I have some reservations about the `sysResponse` stuff.  See my comment in that section.

scott-taubman (approved):

PR #1351, opened 2022-08-08
#1347 - Fix 404 response on queue DELETE

### LINKED ISSUE #1347: Clearing queue results in 404
In some cases, seemingly related to certain characters being present in the instance name, attempting to clear the instance queue through the UI results in a 404 being returned.  A 404 in this context would generally mean that no queue by the requested name was found in rabbitmq.  What's possibly happening here is that we are not correctly generating the queue name from the instance name when certain characters are present, resulting in the queue not being found.

One example that reproduces this problem is naming your instance "This->Breaks".

scott-taubman:
Closes #1347 

This PR updates the url routes to be less strict on the allowed queue names.  Previously, only word characters plus hyphen were allowed.  This was problematic since the queue names are derived from instance names, which do not have this restriction.

Most of the code in this PR is refactoring the way the routes are defined so that they could be leveraged in the unit tests, rather than the unit tests having their own set of routes defined (a longstanding TODO).  This was necessary to allow a meaningful unit test to be written.

## Testing instructions
* The unit test pretty well covers this, so you could just look at that if you care to.
* To functionally test:
  * Create a plugin instance with a name like "this->breaks"
  * Go to the System Admin page and clear the queue for that instance.  Before this change it would result in a 404, but now should succeed.

### DIFF
--- a/src/app/beer_garden/api/http/__init__.py
+++ b/src/app/beer_garden/api/http/__init__.py
@@ -4,7 +4,7 @@ import os
 import ssl
 import types
 from copy import deepcopy
-from typing import Optional, Tuple
+from typing import List, Optional, Tuple
 
 from apispec import APISpec
 from brewtils.models import Event, Events, Principal
@@ -35,7 +35,7 @@ from brewtils.schemas import (
 )
 from tornado.httpserver import HTTPServer
 from tornado.ioloop import IOLoop
-from tornado.web import Application, RedirectHandler
+from tornado.web import Application, RedirectHandler, RequestHandler
 
 import beer_garden
 import beer_garden.api.http.handlers.misc as misc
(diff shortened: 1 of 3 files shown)


src/app/beer_garden/api/http/__init__.py
```diff
@@ -76,6 +76,117 @@
...
+        (rf"{prefix}api/v1/gardens/?", v1.garden.GardenListAPI),
+        (rf"{prefix}api/v1/namespaces/?", v1.namespace.NamespaceListAPI),
+        (rf"{prefix}api/v1/instances/(\w+)/?", v1.instance.InstanceAPI),
+        (rf"{prefix}api/v1/instances/(\w+)/logs/?", v1.instance.InstanceLogAPI),
+        (rf"{prefix}api/v1/instances/(\w+)/queues/?", v1.instance.InstanceQueuesAPI),
+        (rf"{prefix}api/v1/requests/(\w+)/?", v1.request.RequestAPI),
+        (rf"{prefix}api/v1/requests/output/(\w+)/?", v1.request.RequestOutputAPI),
+        (rf"{prefix}api/v1/systems/(\w+)/commands/(\w+)/?", v1.command.CommandAPI),
+        (rf"{prefix}api/v1/systems/(\w+)/?", v1.system.SystemAPI),
+        (rf"{prefix}api/v1/queues/(.*)/?", v1.queue.QueueAPI),
```
scott-taubman, line 105: Other than the refactor for test purposes, this is the only line that actually needed to change for this PR.

src/app/beer_garden/api/http/__init__.py
```diff
@@ -184,89 +295,11 @@ def _setup_application():
 
 
 def _setup_tornado_app() -> Application:
-    prefix = config.get("entry.http.url_prefix")
```
jlrpnbbngtn, line 187: This is a good refactor regardless of the motivation because it makes the code cleaner.

jlrpnbbngtn (approved):
GTG

PR #1353, opened 2022-08-11
#1350 - Remove brew-view/bartender mentions

### LINKED ISSUE #1350: Remove any remaining references to brew

catattack05:
Closes #1350 

I've removed almost all mentions of brew_view, brew-view, and bartender that I could find. This also included deleting unit test files that mentioned these (that were being skipped before). Some things that remain:
-Changelog mentions and anything else that looked like it was logging past changes
-Some binary files from the tests and potentially things in src/app/beergarden.egg-info we're not tracking
-docker/docker-compose/data/prometheus/prometheus.yml
-Credits for the guy who came up with the bartender name, though I did add that it's no longer being used


test/conf/config.json
```diff
@@ -1,5 +1,4 @@
 {
-  "DB_NAME": "test_bartender",
```
scott-taubman, line 2: It's not obvious to me where this file gets used, if at all.  If it were being used, I would think removing this line altogether would break something.  Let's just delete this file (and the directory it's in since it would then be empty) and if the integration tests break even worse than they're already broken, I'll decide what to do about that.

scott-taubman (changes requested):

PR #1355, opened 2022-08-19
#1354 - fix QUEUE_DELETE_ALL operation routing

### LINKED ISSUE #1354: Support QUEUE_DELETE_ALL in the router
The "Clear all queues" button in the UI currently encounters an error because the router doesn't know what to do with the QUEUE_DELETE_ALL operation type.

The underlying needed functionality appears to already exist in *queues.py*, so it *should* just be a matter of adding a case to the `if` tower in `router.py:_target_from_type()`.

scott-taubman:
Closes #1354 

This PR semi-fixes the call behind the "Clear All Queues" button on the system admin page.  It adds the necessary route for QUEUE_DELETE_ALL to the router, but only executes the operation on the local garden.  It does not attempt to pass the operation on to all downstream gardens.  Doing so would add a great deal of complexity for little gain and could also be confusing at best, and outright dangerous at worst due to the unreliable nature of the asynchronous event processing.

## Test instructions
On the System Admin page, click the "Clear All Queues" button in the top right.  The operation should succeed (though there's no feedback in the UI).  If you look at the garden logs, you should see a bunch of INFO messages for each of the queues, showing that each one got cleared.  They'll look something like:

```
INFO - Clearing Queue: parent1.file-bytes.1-0-0-dev0.default
```

### DIFF
--- a/src/app/beer_garden/queues.py
+++ b/src/app/beer_garden/queues.py
@@ -14,6 +14,7 @@ technology we use.
 
 import logging
 
+from brewtils.errors import NotFoundError
 from brewtils.models import Events, Queue, System
 
 import beer_garden.db.api as db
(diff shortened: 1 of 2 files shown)


jlrpnbbngtn (approved):
LGTM

PR #1362, opened 2022-09-13
#1361 - add missing headers for CORS

### LINKED ISSUE #1361: Add missing CORS headers
Need to add `Authorization` to `Access-Control-Allow-Headers` to allow the production build of the React UI to run properly. Typical contents for this header would be:

```
            self.set_header("Access-Control-Allow-Headers", "Content-Type, Origin, X-Requested-With, Authorization")
```

This setting is in `base_handler.py`

obr42:
Closes #1361 

Add missing headers to allow CORS requests against API from production build of React UI


scott-taubman (approved):
This handling is pretty naïve, but I believe it makes sense to go with it for now and not think too hard about it unless a specific need to do so arises.

PR #1366, opened 2022-10-04
#1363 - Add react-ui to RPM build process

### LINKED ISSUE #1363: React UI in the RPM
We should add installing the React UI to the main BG RPM, alongside the Angular UI. Update the nginx config so that there's a separate path (/beta or some such thing) that gets you to the React UI.  The idea being that you just install the BG RPM as you normally would, and now you can try out the React UI right alongside the current UI.

One day, when the React stuff is ready to replace the Angular UI, we just stop including the Angular UI in the RPM and update the nginx confix so that the default path points to React, rather than the /beta.

The RPM build process is presumably nice and straightforward currently because everything is in this repo.  We'll have to modify that process so that it can pull in and build the React app.  That should be doable though.

Split from react-ui ticket [#114](https://github.com/beer-garden/react-ui/issues/114)

scott-taubman:
Closes #1363 

This PR updates the build script and github action to include the latest react-ui release in the RPM that gets built for the beer-garden release.  The result is that with the provided nginx config, the react version of the ui gets served up under the path `/preview`.

## Test Instructions

Manually testing this one is non-trivial, so I don't particularly suggest it.  You can see a successful run of the github action in my fork [here](https://github.com/scott-taubman/beer-garden/actions/runs/3184814284/jobs/5193659617), with the actual react portion of the build beginning [here](https://github.com/scott-taubman/beer-garden/actions/runs/3184814284/jobs/5193659617#step:5:177).  The release with the actual rpm artifact is found [here](https://github.com/scott-taubman/beer-garden/releases/tag/3.15.0).

You can download the rpm and install it in a centos container if you really want to see it work.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -182,7 +182,7 @@ jobs:
       - name: Make Deps
         run: |
           python -m pip install --upgrade pip
-          pip install bandit
+          pip install "importlib-metadata<5" bandit
 
       - uses: actions/checkout@v2
         with:
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -120,6 +120,15 @@ jobs:
         id: get_version
         uses: battila7/get-version-action@v2
 
+      - name: Download react-ui
+        uses: robinraju/release-downloader@v1.5
+        with:
+          repository: "beer-garden/react-ui"
+          latest: true
+          tarBall: true
+          zipBall: false
+          out-file-path: "src"
+
       - name: Build RPM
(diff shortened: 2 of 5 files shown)


.github/workflows/pull_request.yml
```diff
@@ -182,7 +182,7 @@ jobs:
       - name: Make Deps
         run: |
           python -m pip install --upgrade pip
-          pip install bandit
+          pip install "importlib-metadata<5" bandit
```
scott-taubman, line 185: Not related to this PR really, but a new release of importlib-metadata happened to come out and break bandit.  This fixes that so the actions can pass.

kvig:
Changes look good.

PR #1450, opened 2024-02-14
Fixed stop button bug

TheBurchLog:
Fixed bug in JS where list eval was done in valid Python instead of Javascript format.


1maple1 (approved):
LGTM

PR #1453, opened 2024-02-15
Support Read Logs as Routable

TheBurchLog:


src/app/beer_garden/router.py
```diff
@@ -610,8 +609,20 @@ def handle_event(event):
                 pass
 
 
+def _operation_converstion(operation: Operation) -> Operation:
```
1maple1: check spelling

PR #1465, opened 2024-03-01
Add Topic Subscriber api and mongo models

1maple1:
brewtils: dynamic_topics


src/app/beer_garden/db/mongo/models.py
```diff
@@ -580,6 +582,49 @@ def clean_update(self):
...
+                    "namespace",
+                    "system",
+                    "version",
+                    "instance",
+                    "command",
+                ],
+                "unique": True,
+            }
+        ],
+    }
```
TheBurchLog: Embedded documents do not need the `meta` section populated. It is not a primary table or indexed. The unique constraints should be handled elsewhere in the code.

test/integration/local_plugins/plugins/publisher_test.py
```diff
@@ -12,8 +14,39 @@ def system_spec():
     }
 
 
+@pytest.fixture
+def topic():
+    return Topic(
+        name="newtopic",
+        subscribers=[
+            Subscriber(garden="default",
+                       namespace="default",
```
TheBurchLog: Change these from "default" to "docker" and it will work. That is the garden name that we use in the testing environment.

src/app/beer_garden/db/mongo/models.py
1maple1: Removed meta section

test/integration/local_plugins/plugins/publisher_test.py
1maple1: Appreciate it. Changed garden and namespace to None to get around BG naming inconsistencies.

src/app/beer_garden/publish_request.py
```diff
@@ -35,18 +38,26 @@ def handle_event(event: Event):
...
-            event.paylaod.instance_name = None
+            event.payload.instance_name = None
             event.payload.command = None
             del event.payload.metadata["_publish"]
 
+        subscribers = []
+        for topic in get_all_topics():
+            if re.findall(topic.name, event.metadata["topic"]):
+                # get a list of subscribers for matching topic
+                subscribers = topic.subscribers
```
TheBurchLog: You need to do a union of `subscribers` with `topic.subscribers`, because you could match on multiple topics. Especially since they can use regex in their topic.

src/app/beer_garden/publish_request.py
```diff
@@ -65,6 +76,21 @@ def process_publish_event(garden: Garden, event: Event):
...
+                if not match:
+                    event_subscriber = Subscriber(
+                        garden=garden.name,
+                        system=system.name,
+                        namespace=system.namespace,
+                        version=system.version,
+                        instance=instance.name,
+                        command=command.name,
+                    )
+                    subscribers = subscribers if subscribers else []
```
TheBurchLog: Instead, it would be easier to do:
```python
if subscribers:
    for subscriber in subscribers:
```

This would check for both empty array and None

src/app/beer_garden/topic.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+def subscriber_match(
+    first_subscriber: Subscriber, second_subscriber: Subscriber
+) -> bool:
+    match = False
+    for item in ["garden", "namespace", "system", "version", "instance", "command"]:
+        first_value = getattr(first_subscriber, item)
+        second_value = getattr(second_subscriber, item)
+        if first_value and second_value:
```
TheBurchLog, line 135: Not a comment, but I thought this was a slick way of doing the comparison. It took me a minute to realize how simple it was to do it this way.

src/app/test/publish_request_test.py
```diff
@@ -39,107 +48,116 @@ def localgarden(localgarden_system):
 
 
 class TestSubscriptionEvent(object):
-    def test_topic_one(self, monkeypatch, localgarden):
+    def test_newtopic(self, monkeypatch, localgarden):
```
TheBurchLog, line 51: Can you make a test where it matches on both the commands topic list and the new Topic/Subscriber list, then prove that it only triggers one request? I don't see anything in your code stating that it wouldn't only do one request, but I want to future-proof us just in case we tinker with that code down the line.

src/app/beer_garden/publish_request.py
1maple1: Fixed

src/app/beer_garden/publish_request.py
1maple1: Agree and implemented as outlined

src/app/test/publish_request_test.py
1maple1: Added new test that will match on topic commands and the new topic subscribers. Triggers on commands topics first.

TheBurchLog:
Can you update the change log with the overview of these changes. This will be a 3.25.0 release. Anything that is not released already will be merged into the 3.25.0 change log.

TheBurchLog (approved):

PR #1478, opened 2024-03-15
Fixed STOMP preview format

TheBurchLog:


src/ui/src/js/controllers/admin_garden.js
```diff
@@ -244,11 +244,17 @@ export default function adminGardenController(
 
     if (isReceiving){
       if (connection.config["subscribe_destination"] !== undefined){
-        url = url + " " + connection.config["subscribe_destination"];
+        if (connection.config["subscribe_destination"].charAt(0) != "/" && url.charAt(url.length - 1) != "/"){
```
1maple1: What about "/dest" and "url/"? This won't add a / to the end of url but will combine it into "url//dest". Is this possible?

Would it be better to strip any leading "/" from dest and any tailing "/" from url and manually add the "/" when doing the combine?

1maple1:
```
dest = dest.replace(/^\/+/, "")
url = url.replace(/\/+$/, "")
url = url.concat("/", dest)
```

TheBurchLog: I tried giving this a try and the regex wasn't supported in the way it was written. So I broke it down to a simple example for testing:
```javascript
const topic = "/topic/subtopic";
const regex = /^/+/;
console.log(topic.replace(regex, ''));
// Expected output: "topic/subtopic"
```
This gets an error of:
```
Error: Invalid regular expression: missing /
```

1maple1: Sorry, the escapes were lost because I didn't use the code feature. Fixed it.

src/ui/src/js/controllers/admin_garden.js
```diff
@@ -242,23 +242,20 @@ export default function adminGardenController(
       url = url + "/" + connection.config["url_prefix"];
     }
 
+    //const regex = /^/+/;
+    //const regex2 = //+$/;
+
```
1maple1: Probably want to remove the commented out lines

1maple1 (approved):
Double check the changelog prior to merging

PR #1514, opened 2024-04-24
Fixing name of max-concurrent plugin

1maple1:
Example plugin name fixed so fixing misspelling of name here.


TheBurchLog:
OBE from Jmeter overhaul

PR #1522, opened 2024-05-01
Refactoring pruner to use AP Scheduler

### LINKED ISSUE #891: Add file to pruner test
A prune task was added to pruner to delete files. The new task should be added to testing in the pruner test.

1maple1:


src/app/beer_garden/app.py
```diff
@@ -93,39 +93,36 @@ def initialize(self):
...
-            else:
-                self.helper_threads.append(
-                    HelperThread(
-                        db.get_pruner(),
-                        tasks=tasks,
-                        run_every=timedelta(minutes=run_every),
-                        cancel_threshold=ttl_config.in_progress,
-                    )
-                )
+        # Add scheduled jobs for Mongo Pruner
```
TheBurchLog, line 96: Should do an overall check if the pruner interval is greater than 0, then do the same for each TTL to determine if it should be scheduled.

src/app/beer_garden/db/mongo/pruner.py
```diff
@@ -6,251 +6,237 @@
...
-    def run(self):
-        """Runs the pruner to delete tasks"""
+            task["collection"].objects(query).no_cache().delete()
 
-        self.logger.debug(
-            "%s is started with run interval %ss", self.display_name, self._run_every
-        )
+    if cancel_threshold > 0:
+        cancel_outstanding(cancel_threshold)
+    logger.debug(display_name + " is stopped")
```
TheBurchLog: OBE logger?

src/app/beer_garden/db/mongo/pruner.py
```diff
@@ -6,251 +6,237 @@
...
                         )
-                        request.delete()
+                    )
+                    task["collection"].objects(query).limit(
+                        task["batch_size"]
+                    ).no_cache().delete()
 
-    def run(self):
-        """Runs the pruner to delete tasks"""
+            task["collection"].objects(query).no_cache().delete()
```
TheBurchLog: For both of the delete commands, we should see if something like this works:
```python
totalDeleted = task["collection"].objects(query).no_cache().delete().count()
logger.info(f"Pruned {totalDeleted} records")
```

TheBurchLog: https://docs.mongoengine.org/apireference.html#mongoengine.Document.delete

src/app/beer_garden/db/mongo/pruner.py
```diff
@@ -6,57 +6,208 @@
...
+                    )
+                    task["collection"].objects(query).limit(
+                        task["batch_size"]
+                    ).no_cache().delete()
+
+            num = task["collection"].objects(query).no_cache().delete()
+            if num:
+                logger.debug(
+                    "Deleted %s %s from %ss"
+                    % (num, ttl_name, task["collection"].__name__)
```
1maple1, line 60: This only works for files. Delete embedded document in mongoengine is the only delete method that returns a count.

src/app/beer_garden/app.py
1maple1: Resolved

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ TBD
 - Added auto refresh checkbox to requests page
 - Added support for cross-server parent/child requests
 - Reuse existing rabbit admin queue by purging messages instead of creating new queue
+- Refactored Mongo pruner to use APScheduler
```
TheBurchLog: Since we cut the release, need to move this to a new version. Then it is good to go

TheBurchLog:
Need to remove the ttl.multithread option from the config since it shouldn't be used any more
Can you update admin TTL default to 15?

1maple1:
Removed  ttl.multithread option and updated admin TTL to 15

TheBurchLog (approved):

PR #1532, opened 2024-05-16
Fix large output handling on parent requests

1maple1:


src/app/beer_garden/requests.py
```diff
@@ -1003,6 +1003,11 @@ def handle_event(event):
             filter_params={"id": event.payload.id},
             include_fields=[
```
TheBurchLog: Drop `include_fields`, so we don't miss any thing during the update_request function

src/app/beer_garden/requests.py
```diff
@@ -1052,6 +1027,7 @@ def handle_event(event):
 
                     if getattr(existing_request, field) != new_value:
                         request_changed[field] = new_value
```
TheBurchLog, line 1029: Instead, you could use `request_changed` as a boolean value. Then when there is a change, set it to `True`

src/app/beer_garden/requests.py
```diff
@@ -1052,6 +1027,7 @@ def handle_event(event):
...
+                        setattr(existing_request, field, new_value)
 
                 # Add output fields only if the status changes to a compelted state
                 if "status" in request_changed:
```
1maple1, line 1033: The if status in request_changed statement is why I did not change request_changed to a boolean.

TheBurchLog: If it is set to `True` then it will execute. The same as if the dictionary was populated.

1maple1: ```if "status" in True``` will throw an error. It needs to either be a dict or a list to work like this.

TheBurchLog: I concede

TheBurchLog (approved):

PR #1538, opened 2024-05-23
Add Topic Upsert

TheBurchLog:


src/app/beer_garden/topic.py
```diff
@@ -9,16 +9,24 @@
...
     """
-    topic = db.create(topic)
-    return topic
+    try:
+        topic = db.query_unique(Topic, name=new_topic.name, raise_missing=True)
+        # If there are new subscribers, combine them
+        if new_topic.subscribers:
+            for subscriber in new_topic.subscribers:
+                if subscriber not in topic.subscribers:
+                    topic.subscribers.append(subscriber)
```
1maple1, line 26: This will always do a union between old and new subscribers. If the new topic has less subscribers, shouldn't it remove subscribers?

1maple1: What about wrapping the create in a try? If the create succeeds, then there is no reason to do an update. Currently, there's always going to be the hit of doing the query first in both create and update scenarios.

PR #1550, opened 2024-05-30
Loads roles and users from config

1maple1:


src/app/beer_garden/role.py
```diff
@@ -54,14 +55,26 @@ def delete_role(role: Role = None, role_name: str = None, role_id: str = None) -
...
+    roles_yaml = "example_configs/roles.yaml"
+    with open(roles_yaml, "r") as config:
+        return yaml.safe_load(config)
+
 def rescan():
     """ Rescan the roles configuration file"""
-    pass
+    roles_config = load_roles_config()
+    for role in roles_config:
+        role = Role(name=role.get("name"), permission=role.get("permission"))
```
TheBurchLog: Missing the various scopes for roles. These are lists of string values.
https://github.com/beer-garden/brewtils/blob/User_Model/brewtils/models.py#L1666

src/app/beer_garden/role.py
```diff
@@ -54,14 +55,32 @@ def delete_role(role: Role = None, role_name: str = None, role_id: str = None) -
...
+    roles_config = load_roles_config()
+    for role in roles_config:
+        role = Role(name=role.get("name"), permission=role.get("permission"),
+                    description=role.get("description"),
+                    scope_gardens=role.get("scope_gardens"),
+                    scope_namespaces=role.get("scope_namespaces"),
+                    scope_systems=role.get("scope_systems"),
+                    scope_instances=role.get("scope_instances"),
+                    scope_versions=role.get("scope_versions"),
+                    scope_commands=role.get("scope_commands"))
```
TheBurchLog: Should add a default value for all of these, so they are not required to be populated. The only required fields are `name` and `permission`. I think nulls will work, but if not, we can make scopes empty arrays and description an empty string.

1maple1: Get was used for the very reason that some of these will not be populated and will default to None much like in the brewtils model.

TheBurchLog: Yeah, I forgot you didn't need to provide defaults. This is all good then.

TheBurchLog (approved):

TheBurchLog (approved):

PR #1554, opened 2024-06-26
LDAP user authentication

1maple1:


src/app/beer_garden/api/http/authentication/login_handlers/ldap.py
```diff
@@ -0,0 +1,122 @@
...
+        Args:
+            request: tornado HTTPServerRequest object
+
+        Returns:
+            User: The User object matching the supplied username and password
+            None: If no User was found or the supplied password was invalid
+        """
+        authenticated_user = None
+
+        if request.body:
```
TheBurchLog, line 92: We should check if it is enabled before we do any checks. Probably should do this in Trusted Headers as well.

`if config.get("ldap.enabled") and request.body:`

src/app/beer_garden/api/http/authentication/login_handlers/ldap.py
```diff
@@ -0,0 +1,122 @@
...
+                                authenticated_user = get_user(username=username)
+                            except DoesNotExist:
+                                authenticated_user = User(
+                                    username=username, is_remote=True
+                                )
+                                authenticated_user = create_user(authenticated_user)
+                            authenticated_user.metadata["last_authentication"] = (
+                                datetime.now(timezone.utc).timestamp()
+                            )
+                            authenticated_user.upstream_roles = self.get_user_roles(
```
TheBurchLog: `authenticated_user.roles=self.get_user_roles(`

Since we are using the BG managed roles, we should assign it to the `roles` field. Trusted Headers can get away with providing upstream roles because it could provide JSON blobs for them but LDAP doesn't have the ability to generate a object, unless we do some crazy parsing.

src/app/example_configs/config.yaml
```diff
@@ -199,3 +201,12 @@ ui:
   icon_default: fa-beer
   name: Beer Garden
   auto_refresh: false
+ldap:
```
TheBurchLog: Unless we have another use for LDAP, we should move this under `auth.authentication_handlers.ldap`

TheBurchLog:
For the unit testing, take a look at this page to mock the LDAP connection. So you can confirm that only roles in the database are mapped. Also, to confirm that default roles are added as expected.

https://ldap3.readthedocs.io/en/latest/mocking.html

src/app/beer_garden/api/http/authentication/login_handlers/ldap.py
1maple1: There is already an enabled_login_handlers() function so this seems a bit redundant. What do you think? https://github.com/beer-garden/beer-garden/blob/438ab0a2c716d851747ce359644fab57dce2b609/src/app/beer_garden/api/http/authentication/login_handlers/__init__.py#L13C5-L13C29

src/app/example_configs/config.yaml
1maple1: complete

src/app/beer_garden/api/http/authentication/login_handlers/ldap.py
1maple1: complete

src/app/beer_garden/api/http/authentication/login_handlers/ldap.py
TheBurchLog: yeah I couldn't remember how we made the list, so you are good

src/app/beer_garden/api/http/authentication/login_handlers/ldap.py
```diff
@@ -0,0 +1,125 @@
...
+            groups.add(entry["cn"].value)
+
+        for group in groups:
+            try:
+                roles.append(get_role(role_name=group))
+            except DoesNotExist:
+                pass
+
+        logger.info(
+            f"Updating {username} upstream roles to {[role.name for role in roles]}"
```
TheBurchLog: f"Updating {username} local roles to {[role.name for role in roles]}"

src/app/test/api/http/unit/authentication/login_handlers/ldap_test.py
```diff
@@ -0,0 +1,509 @@
...
+        handler = LdapLoginHandler()
+
+        request = HTTPServerRequest(
+            body=b'{"username":"sbrown20","password":"sbrown20"}'
+        )
+        authenticated_user = handler.get_user(request)
+
+        assert authenticated_user is not None
+        assert authenticated_user.username == "sbrown20"
+        assert len(authenticated_user.local_roles) == 2
```
TheBurchLog, line 507: Along with asserting the length, we want to confirm the expected roles were added. Could you add this to the section after the len assertion:
```python
        for local_role in authenticated_user.local_roles:
            if local_role.name == role2.name or local_role.name == role3.name:
                # Confirm that new roles were added
                assert True
            elif  local_role.name == role1.name:
                # Confirm that old role was removed
                assert False 
            else:
                # Unknown role added
                assert False
```

TheBurchLog (approved):

src/app/beer_garden/api/http/authentication/login_handlers/ldap.py
1maple1: done

src/app/test/api/http/unit/authentication/login_handlers/ldap_test.py
1maple1: done

PR #1557, opened 2024-07-05
Blocklist removal

1maple1:


TheBurchLog:
Instead of deleting the API entry points, we should mark them as deprecated. Take a look in the commands API how we are handling it for `CommandAPIOld`:
```python
class CommandAPIOld(AuthorizationHandler):
    async def get(self, command_id):
        """
        ---
        summary: Retrieve a specific Command
        deprecated: true
        parameters:
          - name: command_id
            in: path
            required: true
            description: The ID of the Command
            type: string
        responses:
          200:
            description: Command with the given ID
            schema:
              $ref: '#/definitions/Command'
          404:
            $ref: '#/definitions/404Error'
          50x:
            $ref: '#/definitions/50xError'
        tags:
          - Deprecated
        """
        raise EndpointRemovedException(
            message=(
                "This endpoint has been removed as Commands no longer have IDs. "
                "Please use /systems/<system_id>/commands/<command_name> instead."
            )
        )
```

1maple1:
Deprecated API endpoints instead. They will show up under Deprecated section in Swagger docs.

TheBurchLog:
image

Error on swagger docs

CHANGELOG.md
```diff
@@ -1,10 +1,11 @@
 # Beer Garden Changelog
 
-# 3.26.3
+# 3.27.0
```
TheBurchLog: Let's keep this 3.26.3 for the moment, so we can do a release without all of the other major changes.

PR #1558, opened 2024-07-05
Update NPM packages to support Node.js 21

TheBurchLog:
Ran `npm audit fix` and now we can support up to 21 with no errors in the logs, but 22 throws an error due to `http-proxy` not being supported. The error thrown is:
`DeprecationWarning: The `util._extend` API is deprecated. Please use Object.assign() instead.`

Caused by `webpack-dev-server`
```
├─┬ webpack-dev-server@3.11.3
│ ├─┬ http-proxy-middleware@0.19.1
│ │ ├─┬ http-proxy@1.18.1
```


1maple1 (approved):
Make setup working fine with this change.

PR #1559, opened 2024-07-08
Fixed Command Index Breadcrumbs

TheBurchLog:


1maple1 (approved):
Cool solution. Based on what I can tell this approach will iterate the full list even if there are any duplicate names.

PR #1571, opened 2024-08-05
Make request auto refresh configurable

1maple1:
brewtils: 3.27.0_staged


src/app/beer_garden/config.py
```diff
@@ -620,6 +620,11 @@ def load_child(config_file):
             "description": "The title to display on the GUI",
             "previous_names": ["application_name"],
         },
+        "auto_refresh": {
+            "type": "bool",
+            "default": False,
+            "description": "Auto refresh user interface",
```
TheBurchLog, line 633: "Auto refresh request user interface", I think everywhere else is fine with just auto_refresh but we can spell it out more here. Only comment I could find.

TheBurchLog (approved):

PR #1574, opened 2024-08-06
Db model validation

TheBurchLog:
Adds validation to ensure models can be properly loaded at startup

brewtils: 3.27.0_staged


src/app/test/db/mongo/util_test.py
```diff
@@ -59,6 +66,9 @@ def test_same_indexes(self, model_mocks):
             model_mock._get_collection = Mock(
                 return_value=Mock(index_information=Mock(return_value={"index1": {}}))
             )
+            # model_mock.objects = Mock(
```
1maple1: Remove if not needed

1maple1:
Add changelog

PR #1589, opened 2024-08-19
Apm service name fix

TheBurchLog:


src/app/beer_garden/metrics.py
```diff
@@ -170,7 +171,7 @@ def initialize_elastic_client(label: str):
     if config.get("metrics.elastic.enabled"):
         Client(
             {
-                "SERVICE_NAME": f"{config.get('garden.name')}-{label}",
+                "SERVICE_NAME": f"{re.sub(r'[^a-zA-Z0-9 _-]+$', '', config.get('garden.name'))}-{label}",
```
1maple1: Could this be simplified to [^a-zA-Z0-9 _-] to catch bad characters anywhere in the string or am I missing the intent? This will only catch those at the end because of the $.

1maple1 (approved):
See inline comment before merging.

PR #1594, opened 2024-08-20
Beer.conf Plugin requires

1maple1:
brewtils: develop


src/ui/src/js/controllers/admin_system.js
```diff
@@ -158,6 +158,9 @@ export default function adminSystemController(
             if (instance.status == "UNRESPONSIVE"){
               return $rootScope.getIcon('fa-triangle-exclamation');
             }
+            if (instance.status == "AWAITING_SYSTEM"){
+              return $rootScope.getIcon('fa-face-grin-beam-sweat')
```
TheBurchLog: Might want to look for other options, but it is a funny icon for this

1maple1: fa-spinner is probably more appropriate unless there is a better idea

1maple1: Decided to go with fa-hourglass so no animation was required.

PR #1605, opened 2024-08-26
Fixed remove topic call

TheBurchLog:


src/app/beer_garden/topic.py
```diff
@@ -182,7 +182,7 @@ def prune_topics(garden):
             if update_subscribers:
                 if len(valid_subscribers) == 0:
                     logger.debug(f"Removing Topic {topic.name}")
-                    remove_topic(topic)
+                    remove_topic(topic_name=topic)
```
justinryan1: So right now you're giving the topic object in as the topic_name so I think it would be topic=topic OR topic_name=topic.name

justinryan1 (approved):
LGTM

PR #1608, opened 2024-08-27
Updating logging messages for garden rescan

TheBurchLog:


src/app/beer_garden/garden.py
```diff
@@ -665,9 +665,14 @@ def rescan():
                 garden = db.query_unique(Garden, name=garden_name)
 
                 if garden is None:
+                    logger.info(f"Loading new configuration file for {garden.name}")
                     garden = create_garden(
                         Garden(name=garden_name, connection_type="Remote")
                     )
+                else:
+                    logger.debug(
```
justinryan1: For now I think we should make this INFO level

justinryan1 (approved):

PR #1616, opened 2024-09-03
Topic propagate bug

TheBurchLog:
Child Topic Publish events were being treated as local events


src/app/test/publish_request_test.py
```diff
@@ -121,6 +121,7 @@ def topics():
 
 
 class TestSubscriptionEvent(object):
+
```
1maple1: Blank line needed here?

1maple1 (approved):

PR #1630, opened 2024-09-09
Update instance error

1maple1:


src/app/beer_garden/plugin.py
```diff
@@ -546,9 +546,26 @@ def _from_kwargs(
...
+                route(
+                    Operation(
+                        operation_type="GARDEN_SYNC",
+                        target_garden_name=config.get("garden.name"),
+                        kwargs={"sync_target": config.get("garden.name")},
+                    )
+                )
+                raise NotFoundException(
+                    f"Unable to find system matching instance id {instance_id}"
+                )
```
TheBurchLog: I think this except should be the first exception caught in the handle_event try/except statement. We should also target the `event.garden` for the sync.

TheBurchLog (approved):

PR #1641, opened 2024-09-12
Check System instead of Namespace for Local Request

### LINKED ISSUE #1632: Namespace != Garden Name
There are a few places where the Namespace is being compared to the Garden name. While this is the default value for Systems not providing a Namespace, local systems can provide a Namespace and this will no longer work.

The only way to determine if it is a local request is to check the System that executed it and see if the `local` flag is set.

TheBurchLog:
fixes: #1632

brewtils: fix_namespace_usage

### DIFF
--- a/src/app/beer_garden/router.py
+++ b/src/app/beer_garden/router.py
@@ -713,17 +713,6 @@ def _pre_route(operation: Operation) -> Operation:
         if operation.model.namespace is None:
             operation.model.namespace = config.get("garden.name")
 
-    elif operation.operation_type == "SYSTEM_READ_ALL":
-        # what looks like a sensible edit is to change the next line to:
-        #   operation.kwargs.get("filter_params", {}).get("namespace", "") == "":
-        # but that will break everything, as it turns out. So, unless
-        # operation.kwargs has a key called filter_params whose value is a dictionary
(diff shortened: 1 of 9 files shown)


src/app/beer_garden/router.py
```diff
@@ -713,17 +713,6 @@ def _pre_route(operation: Operation) -> Operation:
...
 
-    elif operation.operation_type == "SYSTEM_READ_ALL":
-        # what looks like a sensible edit is to change the next line to:
-        #   operation.kwargs.get("filter_params", {}).get("namespace", "") == "":
-        # but that will break everything, as it turns out. So, unless
-        # operation.kwargs has a key called filter_params whose value is a dictionary
-        # that has a key called namespace and its value is an empty string, the follow-
-        # ing never runs. Intuition says that's not what was intended here, so
-        # TODO: look into this
-        if operation.kwargs.get("filter_params", {}).get("namespace") == "":
```
1maple1, line 724: So this is no longer needed after all?

1maple1 (approved):

src/app/beer_garden/router.py
TheBurchLog: No, because I updated the logic on the brewtils side

Issue #1584, opened 2024-08-16
Beer Garden support loading a Schedule Jobs file from a configuration file at startup


TheBurchLog:
Support loading the Schedule Jobs from a static file at startup through a config.yaml property. This file should be the same file that is exported and imported via the Scheduler page.

PR #1642, opened 2024-09-12
Load jobs file from configuration file at startup

1maple1:
Addresses #1584


src/app/beer_garden/app.py
```diff
@@ -284,6 +285,10 @@ def _startup(self):
         if config.get("plugin.local.logging.config_file"):
             self.plugin_local_log_config_observer.start()
 
+        if config.get("scheduler.job_startup_file"):
```
TheBurchLog: `if config.get("scheduler.job_startup_file") and os.path.isfile(config.get("scheduler.job_startup_file"):`

TheBurchLog (approved):

PR #1645, opened 2024-09-17
Add API support of Base64 File MD5 Tracking

### LINKED ISSUE #1640: File Requests have MD5 Check
When a file is uploaded via Base64, include in the metadata the MD5 Checksum of the File so it can be validated when the child receives the file.

```python
from hashlib import md5

# Do something like this when we get load the parameter data
md5sum = md5(parameter.getbuffer())
return current_request.metadata[f"{parameter_name}_MD5_SUM"] == md5sum 
```

TheBurchLog:
brewtils: md5_sum_file_chunks

fixes: #1640


TheBurchLog:
Once Add-On is released, we need to update that dependency for the UI

1maple1 (approved):
Very slick to use addons to do the bulk of the work to populate the md5.

PR #1649, opened 2024-09-24
Plugin groups

### LINKED ISSUE #1595: beer.conf/env support of Groups
Add support in the beer.conf for setting `groups` or `group`, and support it in the environment variables for just `group`

1maple1:
fixes: #1595

### DIFF
--- a/src/app/beer_garden/local_plugins/manager.py
+++ b/src/app/beer_garden/local_plugins/manager.py
@@ -743,7 +743,7 @@ class PluginManager(StoppableThread):
 
         # System info comes from config file
         for key in _SYSTEM_SPEC:
-            key = key.upper():q
+            key = key.upper()
 
             if key in plugin_config:
                 env["BG_" + key] = plugin_config.get(key)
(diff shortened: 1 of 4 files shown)


src/app/beer_garden/local_plugins/manager.py
```diff
@@ -696,6 +696,11 @@ def _process_args(plugin_config: Dict[str, Any], instance_name: str):
         if plugin_args is not None:
             process_args += plugin_args
 
+        if plugin_config.get("GROUPS"):
+            plugin_groups = plugin_config["GROUPS"]
+            if plugin_groups is not None:
+                process_args += ["--groups=" + arg for arg in plugin_groups]
+
```
TheBurchLog, line 703: You specified `group` and `groups` in the config options but you only used `groups`. We should check if either is present, then append `group` to the `groups` list.

1maple1: I'm not entirely sure `group` is needed here as it works fine without it, but added it anyway. However, adding `group` to the `groups` list will cause duplicates.

TheBurchLog (approved):

PR #1650, opened 2024-09-25
Cancel/Delete Individual request

### LINKED ISSUE #1586: Cancel/Delete Individual Requests
Need the ability to cancel a running request that will cascade the cancellation request to all children. This doesn't stop the anything that a plugin is already processing, but it will prevent it from kicking off the request if it is still on queue.

Need the ability to delete a single request.

TheBurchLog:
fixes: #1586

Delete Request:
image

Cancel Request:
image


1maple1 (commented):
Please add a changelog entry.

1maple1 (approved):

PR #1661, opened 2024-10-04
Topic add sub patch

TheBurchLog:
brewtils: topic_add_sub_patch


1maple1 (approved):
OK - I see that raise_missing defaults to False now negating the try/except

PR #1665, opened 2024-10-07
Prune legacy requests with no command type

### LINKED ISSUE #1657: Pruning Legacy Requests with no Command Type
In legacy BG code, Requests didn't need to be populated with a Command Type. The default value was `ACTION`. So for the pruner task for Action we should also include `None` lookups.

Something like this should cover it.
```python
(
    Q(command_type="ACTION") | 
    Q(command_type=None) | 
    Q(command_type__exists=False)
)
```

1maple1:
Fixes: #1657

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 TBD
 
+- Added support to prune legacy request with no command type
 - Added support for group/groups in beer.conf and group environment variable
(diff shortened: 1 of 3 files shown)


TheBurchLog:
Tested this by running the following and then waited for the pruner to run:
```
db.request.updateOne({}, {$unset:{command_type:1}})
```

TheBurchLog (approved):

PR #1668, opened 2024-10-08
Missing systems alert msg

### LINKED ISSUE #1597: Create Request checks System Requires warning
When a Request is being created, if a required system is not in an operational status, throw a warning message to the requester.

Either: "Unable to determine the status of required system N, proceed with caution" or "Required system N is in status X, proceed with caution"

Or something to the effect you can make the request, but it could fail due to this dependency.

1maple1:
Fixes: #1597

### DIFF
--- a/src/ui/src/js/controllers/command_view.js
+++ b/src/ui/src/js/controllers/command_view.js
@@ -116,7 +116,7 @@ export default function commandViewController(
           missingSystems.push(system.name);
         }
       });
-      helptext = 'Unable to determine the status of required system(s): ' +
+      helptext = 'Required system(s) are not running: ' +
       missingSystems.join(', ') +
       '. Proceed with caution.';
     }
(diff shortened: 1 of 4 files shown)


src/ui/src/js/controllers/command_view.js
```diff
@@ -98,6 +102,27 @@ export default function commandViewController(
...
+      const requires = $scope.system.requires;
+      const missingSystems = [];
+      requires.forEach((req) => {
+        const system = SystemService.findSystem($scope.system.namespace, req, 'latest');
+        const notRunning = system.instances.every((instance) => instance.status != 'RUNNING');
+        if (notRunning) {
+          missingSystems.push(system.name);
+        }
+      });
+      helptext = 'Unable to determine the status of required system(s): ' +
```
TheBurchLog: Slightly different wording? They might have been found, but just were not running. So we did determine the status of the required system.

TheBurchLog (approved):

PR #1678, opened 2024-10-10
New user model migration from 3.26 and Garden Model from 3.23

### LINKED ISSUE #1666: Expand Exception Handling for Document Validation
Currently we are only catching `IndexOperationError` in the primary try/except statement. When we have a model migration like UserToken as we did in 3.26.5 to 3.27.2, it will throw an `OperationError` when it attempts to `get_collection()`. If we catch this exception, then it should all continue to operate as excepted. 

https://github.com/beer-garden/beer-garden/blob/1eec3e83bd9de9cefa7f3d2fb524159475bb548b/src/app/beer_garden/db/mongo/util.py#L184

TheBurchLog:
Adds Model Migration from 3.26 and 3.23
fixes: #1666

### DIFF
--- a/src/app/beer_garden/db/mongo/util.py
+++ b/src/app/beer_garden/db/mongo/util.py
@@ -152,9 +152,9 @@ def ensure_v3_24_model_migration():
 
         logger.warning(
             "Encountered an error loading Gardens. This is most likely because"
-            " the database is using the old (v3.23) style of storing in the database. To"
-            " fix this the Garden collections mapped to yaml files, then"
-            " drop the Garden collection."
+            " the database is using the old (v3.23 or prior) models. Migration"
+            " strategy is to map all records in the Garden collection to yaml"
+            " files, then drop the Garden collection to be rebuilt."
         )
 
         db = get_db()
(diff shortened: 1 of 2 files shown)


src/app/beer_garden/db/mongo/util.py
```diff
@@ -127,11 +127,49 @@ def ensure_v2_to_v3_model_migration():
         db.drop_collection("system")
 
 
+def ensure_v3_27_model_migration():
+    """Ensures that the Role model is consolidated
+
+    In Version 3.6 and earlier the utilized role assignments to determine the
```
justinryan1: Think that's supposed to be 2 and 3 and not 3.6 and 3.7 respectively

TheBurchLog:
These are the expected log statements when we do the migrations:
```
2024-10-11 15:25:16,637 - beer_garden.db.mongo.util - WARNING - Encountered an error loading Gardens. This is most likely because the database is using the old (v3.23) style of storing in the database. To fix this the Garden collections mapped to yaml files, then drop the Garden collection.
2024-10-11 15:25:16,638 - beer_garden.db.mongo.util - WARNING - Mapping Child Config: ./children/http.yaml
2024-10-11 15:25:16,640 - beer_garden.db.mongo.util - WARNING - Mapping Child Config: ./children/stomp.yaml
2024-10-11 15:25:16,673 - beer_garden.db.mongo.util - WARNING - Encountered an error loading Roles or Users or User Tokens. This is most likely because the database is using the old (v3.26) style of storing in the database. To fix this the roles, remote_roles, role_assignment, user, remote_user, and user_token collections will be dropped.
```

src/app/beer_garden/db/mongo/util.py
```diff
@@ -127,11 +127,142 @@ def ensure_v2_to_v3_model_migration():
...
+
+        import yaml
+
+        db = get_db()
+
+        logger.warning(
+            "Encountered an error loading Gardens. This is most likely because"
+            " the database is using the old (v3.23) style of storing in the database. To"
+            " fix this the Garden collections mapped to yaml files, then"
+            " drop the Garden collection."
```
1maple1: Wording? will be mapped before dropping the Garden collection?

PR #1686, opened 2024-10-23
Prune topics when system removed

### LINKED ISSUE #1663: Evaluate Topic Pruning
Currently topics are pruned when a Garden Sync event is seen. We should also look at pruning system/instance specific subscribers when they are deleted.

1maple1:
fixes: #1663


TheBurchLog:
When deleting a system, it deletes all the other topics and leaves just the topics for the system being deleted.

src/app/beer_garden/topic.py
```diff
@@ -184,14 +184,26 @@ def subscriber_match(
     return match
 
 
-def prune_topics(garden):
+def topic_has_system_subscribers(topic: Topic, system: System):
+    for subscriber in topic.subscribers:
+        if subscriber.system == system.name:
```
TheBurchLog: This should check `system.name`, `system.namespace`, and `system.version`. Because if you have two versions of a System running, it could delete the dynamic topics from the other one.

TheBurchLog (approved):

PR #1688, opened 2024-10-31
Add support for shutdown and startup functions

TheBurchLog:


TheBurchLog:
Example Plugin

```yaml
PLUGIN_ENTRY="-m shutdown"
CLIENT_SHUTDOWN_FUNCTION="shutdown_two"
CLIENT_STARTUP_FUNCTION="startup_two"
```

```python
from brewtils import Plugin, command, shutdown, startup

__version__ = "3.0.0.dev0"

class ShutdownClient:

    @command
    def my_function(self):
        return "my_function"
    
    @shutdown
    def shutdown_one(self):
        print("Internal Shutdown One Function")

    def shutdown_two(self):
        print("Internal Shutdown Two Function")

    @startup
    def startup_one(self):
        print("Internal Startup One Function")

    def startup_two(self):
        print("Internal Startup Two Function")

def main():
    def other_shutdown_function():
        print("External Shutdown Function")

    def other_startup_function():
        print("External Startup Function")

    Plugin(
        ShutdownClient(),
        name="shutdown_example",
        version=__version__,
        shutdown_function=other_shutdown_function,
        startup_function=other_startup_function,
    ).run()


if __name__ == "__main__":
    main()
```

src/app/beer_garden/local_plugins/manager.py
```diff
@@ -706,6 +706,38 @@ def _process_args(plugin_config: Dict[str, Any], instance_name: str):
             if plugin_group is not None:
                 process_args += ["--group=" + str(plugin_group)]
 
+        if plugin_config.get("CLIENT_SHUTDOWN_FUNCTIONS") or plugin_config.get(
+            "CLIENT_SHUTDOWN_FUNCTIONS"
```
1maple1, line 710: This is the same condition

src/app/beer_garden/local_plugins/manager.py
```diff
@@ -706,6 +706,38 @@ def _process_args(plugin_config: Dict[str, Any], instance_name: str):
...
+                shutdown_functions.append(
+                    plugin_config.get("CLIENT_SHUTDOWN_FUNCTIONS")
+                )
+
+            process_args += [
+                "--client_shutdown_functions=" + arg for arg in shutdown_functions
+            ]
+
+        if plugin_config.get("CLIENT_STARTUP_FUNCTIONS") or plugin_config.get(
+            "CLIENT_STARTUP_FUNCTIONS"
```
1maple1: This is the same condition

1maple1 (approved):
Just the two comments, otherwise looks good.

PR #1708, opened 2024-12-09
Update Authorization Filtering for Connection configurations

### LINKED ISSUE #1707: Garden Connection Params Filtering
In `authorization.py` for the Model filter `_get_garden_filter` we need to remove `connection_params` if the user does not have `GARDEN_ADMIN` for the Garden.

TheBurchLog:
Fixes: #1707

### DIFF
--- a/src/app/beer_garden/authorization.py
+++ b/src/app/beer_garden/authorization.py
@@ -162,15 +162,17 @@ class QueryFilterBuilder:
                 if role.permission in permission_levels:
                     filter = {}
                     if len(role.scope_systems) > 0:
-                        filter["request_template__system__in"] = role.scope_systems
+                        filter["request_template__system__in"] = (
+                            role.scope_systems
+                        )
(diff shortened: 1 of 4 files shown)


src/app/beer_garden/authorization.py
```diff
@@ -809,6 +809,38 @@ def _get_garden_filter(
...
+                            "scope_instances",
+                            "scope_versions",
+                            "scope_commands",
+                        ],
+                    )
+                ):
+                    allow_connection_params = True
+                    break
+            if allow_connection_params:
+                break
```
1maple1, line 833: It seems this break would never run.

1maple1 (approved):

src/app/beer_garden/authorization.py
TheBurchLog: It would break if you are currently evaluating `local_roles`, then it skip over evaluating `upstream_roles`

PR #1712, opened 2024-12-12
Prepopulate filename for pour again file uploads

### LINKED ISSUE #1691: Pour Again File Uploads
For file_uploads if you pour it again the contents of the uploaded doc are correctly populated in the params and the upload input field is validated with a green box. However the text inside the input box doesn't show the filename like when you first uploaded it.

Expected behavior is to load the file name inside the field.

We should also see what happens with you pour it again when the file has already been pruned to make sure that it doesn't give you the green check mark for a valid file.

1maple1:
fixes: #1691

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 TBD
 
+- Prepopulate filename on pour again file uploads by adding filename to request parameters on submit.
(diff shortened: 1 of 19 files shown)


TheBurchLog:
Missing mongo models update?

TheBurchLog (approved):

PR #1713, opened 2024-12-17
Capture streams config

1maple1:


src/app/beer_garden/local_plugins/manager.py
```diff
@@ -627,7 +627,9 @@ def _create_runners(self, plugin_path: Path) -> List[ProcessRunner]:
 
         for instance_name in plugin_config["INSTANCES"]:
             runner_id = self._get_runner_id()
-            capture_streams = plugin_config.get("CAPTURE_STREAMS")
```
TheBurchLog, line 630: We have a hierarchy of where we get our configuration values. We always default to variable set closer to the actual code. So in this scenario, if a developer attempted to explicitly set CAPTURE_STREAMS to `False` where Beer Garden has it set to `True`, then it will be marked as `True`, overriding the closer configuration.

Instead, it should be something like:
```python
plugin_config.get("CAPTURE_STREAMS", config.get("plugin.local.capture_streams"))
```

So we grab the default value if it doesn't exist.

1maple1: done

src/app/example_configs/config.yaml
```diff
@@ -175,6 +175,7 @@ plugin:
       username: plugin_admin
     directory: ./plugins
     host_env_vars: []
+    capture_streams: true
```
TheBurchLog: In the example, let's make this `false`. That way we don't have extra files being generated in our development environment.

TheBurchLog (approved):

src/app/example_configs/config.yaml
1maple1: done

PR #1717, opened 2025-01-02
MIGRATED Add deprecated kwarg to parameter and command

### LINKED ISSUE #280: Parameter deprecations
We should have something in the parameter definition that show a parameter is deprecated.

1maple1:
fixes: #280 
brewtils: parameter-deprecation


TheBurchLog:
Verified that the Brewtils side is populating the Deprecated flags properly. 

The `(Deprecated)` info was not presented on the Command View page. Parameters did not indicate if they were deprecated and the description of the command wasn't updated either.

I wonder if we should treat deprecated commands the same as hidden commands. So once deprecated, we automatically hide them unless they toggle the `show hidden` feature. Another visual indication that is used by tools like Swagger is a slash through the description.

TheBurchLog (approved):

PR #1718, opened 2025-01-03
MIGRATED  Update Orphan Pruner for files

TheBurchLog:


src/app/beer_garden/db/mongo/pruner.py
```diff
@@ -66,6 +66,7 @@ def prune_by_name(ttl_name):
     match_keys = [ttl_name, "batch_size"]
     new_ttl_config = {k: ttl_config[k] for k in match_keys if k in ttl_config}
     tasks = determine_tasks(**new_ttl_config)
+    logger.error(f"Pruning {ttl_name}")
```
1maple1: This seems more like a debug or info message

src/app/beer_garden/db/mongo/pruner.py
```diff
@@ -211,6 +212,25 @@ def prune_orphans():
...
+        updated_at__lte=timeout,
+    )
+
+    for file in orphaned_files:
+        try:
+            if file.owner_type == "JOB" and file.job is not None:
+                Job.objects.get(id=file.job.id)
+            elif file.owner_type == "REQUEST" and file.request is not None:
+                Request.objects.get(id=file.request.id)
+        except DoesNotExist:
```
1maple1, line 230: What if file.job is None and/or file.request is None? What if owner_type is None? These fields are not required.

src/app/beer_garden/db/mongo/pruner.py
TheBurchLog: I removed it. That was there for testing only

src/app/beer_garden/db/mongo/pruner.py
TheBurchLog: In the normal scenario, a Job or Request gets deleted. The File object utilized `reverse_delete_rule=NULLIFY`, so the field gets nulled out instead of deleting the File record. 

We probably only need this for the `job` field on File because a Job could be running when it is deleted so the file is still required. The `request` field could be `reverse_delete_rule=CASCADE`. But we are keeping them consistent and letting the file pruner handle it.

`owner_type` is something we enforce in code.

1maple1: Ok - so we only want to prune the files that have an invalid job or request. Under a normal scenario where job or request is deleted the field will be set to None and the file will be retained. I see that this scenario is handled in the test.

1maple1 (approved):

PR #1723, opened 2025-01-10
MIGRATED Add certificate login handler

### LINKED ISSUE #1709: Certificate Login
Create an authentication handler that when a trusted cert matches a user in the database, they are logged in.

1maple1:
fixes: #1709

### DIFF
--- a/src/app/beer_garden/api/http/authentication/login_handlers/certificate.py
+++ b/src/app/beer_garden/api/http/authentication/login_handlers/certificate.py
@@ -6,11 +6,8 @@ from brewtils.models import User
 from mongoengine import DoesNotExist
 from tornado.httputil import HTTPServerRequest
 
-from beer_garden import config
 from beer_garden.api.http.authentication.login_handlers.base import BaseLoginHandler
-from beer_garden.api.http.schemas.v1.token import TokenInputSchema
-from beer_garden.role import get_role
-from beer_garden.user import create_user, get_user, update_user
+from beer_garden.user import get_user, update_user
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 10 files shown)


1maple1:
I'm thinking the client_cert config option could go with in plugin.auth instead of entry.http.ssl?

src/app/example_configs/config.yaml
```diff
@@ -1,9 +1,11 @@
 auth:
-  enabled: false
+  enabled: true
```
TheBurchLog: Need to revert back features to disabled.

src/app/beer_garden/api/http/authentication/login_handlers/certificate.py
```diff
@@ -0,0 +1,56 @@
...
+                    user = get_user(username=username)
+
+                    authenticated_user = user
+                    authenticated_user.metadata["last_authentication"] = datetime.now(
+                        timezone.utc
+                    ).timestamp()
+                    authenticated_user = update_user(user=authenticated_user)
+
+                except DoesNotExist:
+                    pass
```
TheBurchLog: Should confirm if we should add a feature to create users for when common name is found but doesn't exist. This is how Trusted Header currently works.

src/app/example_configs/config.yaml
```diff
@@ -71,7 +73,8 @@ entry:
     ssl:
       ca_cert: null
       ca_path: null
-      client_cert_verify: NONE
+      client_cert_verify: OPTIONAL
+      client_cert: null
```
TheBurchLog, line 78: I agree this should go under `plugin.local.auth` for `client_cert`

src/app/example_configs/config.yaml
1maple1: Done

src/app/example_configs/config.yaml
1maple1: Done

src/app/beer_garden/api/http/authentication/login_handlers/certificate.py
1maple1: Done

src/app/beer_garden/config.py
```diff
@@ -648,6 +648,22 @@ def load_child(config_file):
...
+            "type": "dict",
+            "items": {
+                "enabled": {
+                    "type": "bool",
+                    "default": True,
+                    "description": "Enable authentication via certificate",
+                },
+                "create_users": {
+                    "type": "bool",
+                    "default": True,
```
TheBurchLog: Default this to False. That way they can selectively add this.

TheBurchLog (approved):
Approved after the default value of Created User is set to `False`

PR #1730, opened 2025-01-16
MIGRATED Add filter for command display name migration

1maple1:


src/app/beer_garden/db/mongo/util.py
```diff
@@ -259,18 +259,19 @@ def ensure_v3_27_model_migration():
 
 def ensure_v3_29_model_migration():
     db = get_db()
-    if not contains_field("request", ["command_display_name"]):
+    if db.request.find({"command_display_name": {"$exists": True}}).count() == 0:
```
justinryan1: I think this should go into the contains_field function so that all previous and future migrations benefit from the speedup. Also does this conditional actually handle the situation where only some of the documents within a collection do not contain that field? Since we want all of the documents within a collection to be migrated to the newest model.

src/app/beer_garden/db/mongo/util.py
```diff
@@ -259,18 +259,19 @@ def ensure_v3_27_model_migration():
...
         logger.warning(
             "Command display name was not found in Requests and will be added. This is most"
             " likely because the database is using the old (v3.29) style of storing in"
             " the database."
         )
         request_collection = db.get_collection("request")
         for legacy_request in request_collection.find():
```
justinryan1: It might be a good idea to use the above idea here on line 269 `for legacy_request in request_collection.find({"command_display_name": {"$exists": False}}):` this will only return the legacy requests that do not contain `command_display_name`. Now we are only doing database calls to update the requests that needed the migration verses performing the operation on all documents in the request collection.

1maple1: This is a great idea and I implemented it exactly as described.

src/app/beer_garden/db/mongo/util.py
1maple1: For contains_field, if the field is contained in any record then it returns True. Original use case was to look for legacy fields. The command_display_name field is a bit different because it needs to be added if it is not there.  I can't simply invert the contains_field result to run when there are no documents with that field or only some documents without the field. However, I could write another function that does exactly that.

The contains_field is actually a little odd because it allows multiple fields. If there are multiple fields then the logic needs to be (field1 exists OR field2 exists) to match up with the way it is currently written. I can find no place in the code where more than one field is being used with contains_field. Does it make sense to allow contains_field to allow more than one field?

1maple1: Latest version has contains_field (exists True > 0) and missing_field (exists False > 0). Fields has been replaced with field.

TheBurchLog:
Tested the migration scripts from 3.29 to current and it worked. Then tested removing the new field manually in the database off one record to ensure the migration still ran.

Example removal, I canceled the one request so I could have an easy field to flag off.

`db.request.update({status:"CANCELED"},{$unset: {command_display_name:1}})`

justinryan1 (approved):
LGMT

PR #1732, opened 2025-01-21
MIGRATED Adding System Topic Creation from DB Models

TheBurchLog:


1maple1 (approved):
Verified that topics are created/removed for systems.

PR #1744, opened 2025-01-24
MIGRATED Garden  Scan Optimization

TheBurchLog:
fixes: 1734

### DIFF
--- a/src/app/beer_garden/garden.py
+++ b/src/app/beer_garden/garden.py
@@ -518,57 +518,11 @@ def update_garden_receiving(
     return db.update(garden)
 
 
-def load_garden_new_configuration(
-    garden: Garden,
-    receiving_connection: Connection = None,
-    publishing_connection: Connection = None,
-):
-    if receiving_connection:
-        connection_set = False
-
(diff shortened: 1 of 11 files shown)


src/app/beer_garden/garden.py
```diff
@@ -520,123 +529,159 @@ def update_garden_receiving(
...
+                "PUBLISHING" if garden_config.get("publishing") else "DISABLED"
+            )
+
+            for key in config_map:
+                http_publishing_connection.config.setdefault(
+                    config_map[key], config.get(key, garden_config)
+                )
+
+        else:
+            http_publishing_connection.status = "NOT_CONFIGURED"
```
1maple1, line 590: Can this go disabled? Http not enabled and publishing false.

TheBurchLog: I think it should only go disabled if it was configured. But if they never configured it, then mark is as `NOT_CONFIGURED`. So there is a clearer distinction on what is going on.

1maple1 (approved):

PR #1745, opened 2025-01-27
MIGRATED Add config for Tornado max body size

### LINKED ISSUE #1719: Add Tornado Max Size Configuration
Add the ability to configure the max size for Tornado.
https://github.com/tornadoweb/tornado/issues/2435
https://www.tornadoweb.org/en/stable/httpserver.html

Use this total to check files before they are uploaded (bytes) to ensure they do not exceed the max size we can transfer in a single HTTP request.

1maple1:
fixes: #1719
brewtils: file_chunk_size

### DIFF
--- a/src/app/beer_garden/files.py
+++ b/src/app/beer_garden/files.py
@@ -23,7 +23,6 @@ import beer_garden.config as config
 import beer_garden.db.api as db
 from beer_garden.errors import NotUniqueException
 
-MAX_CHUNK_SIZE = 1024 * 1024 * 15  # 15MB
 OWNERSHIP_PRIORITY = {
     "JOB": 1,
     "REQUEST": 2,
@@ -200,12 +199,12 @@ def create_chunk(
     Raises:
         NotFoundError: File with the requested ID doesn't exist and is expected to
     """
-    if len(data) > MAX_CHUNK_SIZE:
+    if len(data) > File.MAX_CHUNK_SIZE:
         return FileStatus(
             operation_complete=False,
             message=(
                 "Chunk data length exceeds the maximum "
-                f"allowable length of {MAX_CHUNK_SIZE}."
+                f"allowable length of {File.MAX_CHUNK_SIZE}."
             ),
             file_id=file_id,
             offset=offset,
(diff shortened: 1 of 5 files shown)


src/app/beer_garden/files.py
```diff
@@ -359,6 +359,12 @@ def create_file(
             f"Cannot create a file with chunk size greater than {MAX_CHUNK_SIZE}."
         )
 
+    max_body_size = config.get("entry.http.max_body_size")
+    if max_body_size and file_size > max_body_size:
+        raise ValueError(
+            f"Cannot create a file with size greater than max body size of {max_body_size}"
+        )
+
```
TheBurchLog: I think this isn't an issue since these values are hard coded in our code. But it is worth leaving it in there to protect our future selves.

What would be more important to know is if you decreased the max body size to smaller than the chunk size, this could be checked during the app startup process. In that scenario, you couldn't post any file chunks to the server.

1maple1: Agree with adding that check. However, the actual chunk size is hard-coded in EasyClient (255 * 1024). I'll have to look into a good way to pull that value out for the comparison.

TheBurchLog: We could put it in the File model object as a static value. Then make everywhere in our code reference that model size. That way the only places we need to worry about it is in brewtils and the UI code.

src/app/beer_garden/files.py
```diff
@@ -354,9 +353,15 @@ def create_file(
...
+    if chunk_size > File.MAX_CHUNK_SIZE:
         raise ValueError(
-            f"Cannot create a file with chunk size greater than {MAX_CHUNK_SIZE}."
+            f"Cannot create a file with chunk size greater than {File.MAX_CHUNK_SIZE}."
+        )
+
+    max_body_size = config.get("entry.http.max_body_size")
+    if max_body_size and file_size > max_body_size:
+        raise ValueError(
+            f"Cannot create a file with size greater than max body size of {max_body_size}"
```
TheBurchLog, line 364: I think we need this logic in the app.py `_startup()` section. Where we do something like:

```python
if config.get("entry.http.max_body_size") < File.MAX_CHUNK_SIZE:
    logger.error(f"some warning about file chunk sizing")
```

TheBurchLog: This won't stop the application from running, but they at least need a warning that they are misconfiguring themselves

1maple1: I went back and forth on this one and decided to put it in the api/http code. If you truly believe it should be in app then I can update.

TheBurchLog (approved):

PR #1746, opened 2025-01-27
MIGRATED Setting up config file to event handler control

TheBurchLog:


src/app/beer_garden/topic.py
```diff
@@ -351,6 +351,9 @@ def handle_event(event: Event) -> None:
 
     All topic handling is done at the Mongo level or scheduled jobs
 
+    This is not setup in the handlers logic. Once implemented then
+    add the logic to the list
+
```
1maple1, line 356: Holdover from topics PR?

src/app/beer_garden/config.py
```diff
@@ -1766,12 +1766,43 @@ def load_child(config_file):
...
+    "type": "dict",
+    "items": {
+        "enabled": {
+            "type": "bool",
+            "default": True,
+            "description": ("If event handler should process the event"),
+        },
+        "unique_data": {
+            "type": "bool",
+            "default": False,
```
1maple1, line 1779: Did you backtrack on setting this to true for applicable event types from the previous PR?

1maple1 (approved):

src/app/beer_garden/config.py
TheBurchLog: Yes, this is a backtrack from it. I figured we should use these values to help improve performance, but if the system isn't having that issue then why not let them process everything.

src/app/beer_garden/topic.py
TheBurchLog: Yup, I left it just in case we needed it in the future. But I wanted to be clear that it wasn't setup for it's own thread.

PR #1752, opened 2025-02-03
Upstream downstream refactor

### LINKED ISSUE #1699: Upstream/Downstream Naming Convention
For Garden connections, we need to have it be Upstream/Downstream instead of Parent/Child. This will make it easier conceptually to understand what way data is flowing.

1maple1:
fixes: #1699 
brewtils: upstream_downstream


rpm/shared-scripts/after_install.sh
```diff
@@ -61,7 +61,7 @@ if [ ! -f "$CONFIG_FILE" ]; then
         --plugin-local-directory "$PLUGIN_HOME" \
         --plugin-local-logging-config-file "$LOCAL_PLUGIN_LOG_CONFIG" \
         --plugin-remote-logging-config-file "$REMOTE_PLUGIN_LOG_CONFIG" \
-        --children-directory "$CHILDREN_CONFIG_HOME" \
+        --children-directory "$DOWNSTREAM_CONFIG_HOME" \
```
TheBurchLog: Did the cli arg get updated in the config?

src/app/beer_garden/db/mongo/models.py
```diff
@@ -1120,7 +1115,7 @@ def _get_system_triple(system: System) -> Tuple[str, str, str]:
         for bad_system_id in child_systems_already_known.values():
```
TheBurchLog: `downstream_systems` instead of `child_systems`

src/app/beer_garden/requests.py
```diff
@@ -1178,9 +1178,9 @@ def clean_command_type_temp(request: Request, is_remote: bool):
     # Delete any children that are TEMP once the current request is completed
     request.children = db.query(Request, filter_params={"parent": request})
 
-    for child in request.children:
-        if child.command_type == "TEMP":
-            db.delete(child)
+    for downstream_garden in request.children:
+        if downstream_garden.command_type == "TEMP":
+            db.delete(downstream_garden)
```
TheBurchLog: These are child requests, not downstream gardens

src/app/beer_garden/requests.py
```diff
@@ -1193,10 +1193,10 @@ def cancel_request_children(request: Request):
...
-    for child in request.children:
-        if child.status in [
+    for downstream_garden in request.children:
+        if downstream_garden.status in [
             "IN_PROGRESS",
             "CREATED",
             "RECEIVED",
         ]:
-            cancel_request(request=child)
+            cancel_request(request=downstream_garden)
```
TheBurchLog: Again, not downstream gardens, these are requests

test/conf/docker-compose-grandparents-test.yml
```diff
@@ -180,8 +178,8 @@ services:
             <<: *env
 
             BG_GARDEN_NAME: child
-            BG_PARENT_HTTP_ENABLED: "true"
-            BG_PARENT_HTTP_HOST: beer-garden-parent
+            BG_UPSTREAM_HTTP_ENABLED: "true"
+            BG_UPSTREAM_HTTP_HOST: beer-garden-parent
```
TheBurchLog: I think this will break testing older frameworks. Because they will require the legacy mappings.

TheBurchLog (commented):
First pass of comments

TheBurchLog:
Attempted to migrate over the configuration file from the previous version and nothing mapped over. Each value needs the previous name parameter like the other PR.

src/app/beer_garden/requests.py
1maple1: fixed

src/app/beer_garden/db/mongo/models.py
1maple1: fixed

rpm/shared-scripts/after_install.sh
1maple1: Right, this is now --downstream-directory.

test/conf/docker-compose-grandparents-test.yml
1maple1: Reverted this back to BG_PARENT_. PARENT should still work since it was added as an alternate env name in the config.

1maple1:
> Attempted to migrate over the configuration file from the previous version and nothing mapped over. Each value needs the previous name parameter like the other PR.

Previous names was added to the config for proper migration.

1maple1:
OBE. See #1802

PR #1754, opened 2025-02-07
MIGRATED Ensure request pruning includes gridfs files

TheBurchLog:


1maple1 (approved):
Mongoengine implementation of only() is weird how it uses defaults for fields not retrieved. Using exclude looks ok to me.

PR #1756, opened 2025-02-10
MIGRATED Pruner config update

### LINKED ISSUE #1727: Pruner Configuration Update
The TTL format doesn't make sense in the config for its purpose. Since we have expanded the scope and implementation of the pruner, the config values are placed in TTL but it doesn't truly fit there. 

Admin and Orpan pruner are cleanup functions and shouldn't be something the system admin needs to configure. Potentially, we could add a flag to disable cleanup.

config.yaml proposed changes

```yaml
# ORIGINAL
db:
  connection:
    host: localhost
    password: null
    port: 27017
    username: null
  name: beer_garden
  prune_interval: 15
  ttl:
    action: -1
    file: 15
    in_progress: 15
    info: 15   
    temp: 15
    admin: 15
    
    orphan: -1

    batch_size: -1

# CHANGES
db:
  connection:
    host: localhost
    password: null
    port: 27017
    username: null
  name: beer_garden
  prune:    
    batch_size: 1000
    execution_interval: 15
    in_progress_request_expiration: 180
    type:
      action: -1
      file: 15
      info: 15   
```

1maple1:
fixes: #1727


src/app/beer_garden/db/mongo/pruner.py
```diff
@@ -89,35 +91,28 @@ def prune_files():
...
-    info_ttl = kwargs.get("info", -1)
-    action_ttl = kwargs.get("action", -1)
-    file_ttl = kwargs.get("file", -1)
-    admin_ttl = kwargs.get("admin", -1)
-    temp_ttl = kwargs.get("temp", -1)
-    batch_size = kwargs.get("batch_size", -1)
-
     prune_tasks = []
-    if info_ttl > 0:
+    batch_size = config.get("db.prune.batch_size")
```
TheBurchLog, line 108: Instead of checking in each statement, could you just check once for length:
```python
if ttl_length <= 0:
    return []
```

src/app/beer_garden/config.py
```diff
@@ -920,83 +915,67 @@ def load_child(config_file):
...
-                "orphan": {
-                    "type": "int",
-                    "default": -1,
-                    "description": (
-                        "Number of minutes to wait before deleting "
-                        "orphaned requests (negative number for never)"
-                    ),
-                },
-                "file": {
+                "prune_interval": {
```
TheBurchLog: This feels weird when writing it out for an environment variable
`BG_DB_PRUNE_PRUNE_INTERVAL`

We should either reduce it down to `interval` or `execution_interval` to make the enviromnet variables make sense.

src/app/beer_garden/config.py
```diff
@@ -920,83 +915,67 @@ def load_child(config_file):
...
+                    "type": "dict",
+                    "items": {
+                        "action": {
+                            "type": "int",
+                            "default": -1,
+                            "description": (
+                                "Number of minutes to wait before deleting "
+                                "ACTION requests (negative number for never)"
+                            ),
+                            "previous_names": ["action_request_ttl"],
```
TheBurchLog: The way to do migration is with periods to separate the levels:
```yaml
"previous_names": ["action_request_ttl","db.ttl.action"],
```
We should add this to all of the migrated fields

1maple1: Cool. There was no example of that in yapconf documentation.

src/app/beer_garden/config.py
1maple1: Went with interval

src/app/beer_garden/db/mongo/pruner.py
1maple1: I like it. Done.

TheBurchLog (approved):
Wait to merge it in

PR #1758, opened 2025-02-11
MIGRATED Fix Garden connections lookup on UI

TheBurchLog:


src/ui/src/js/controllers/admin_garden.js
```diff
@@ -311,10 +311,15 @@ export default function adminGardenController(
...
-        return true;
+  $scope.hasConfiguredConnection = function(garden, connection_type) {
+    if (connection_type == "publishing_connections"){
+      if (garden.publishing_connections !== undefined && garden.publishing_connections != null){
+        return garden.publishing_connections.some(connection => connection.status != 'NOT_CONFIGURED');
+      }
+    }
+    if (connection_type == "receiving_connections"){
+      if (garden.receiving_connections !== undefined && garden.receiving_connections != null){
+        return garden.receiving_connections.some(connection => connection.status != 'NOT_CONFIGURED');
```
1maple1, line 322: I like the use of some over looping through each connection to check the status but I don't think hard-coding the connection type was needed.

```
$scope.hasConfiguredConnection = function(connections) {
    if (connections !== undefined && connections != null){
      return connections.some(connection => connection.status != 'NOT_CONFIGURED');
    }
```

PR #1760, opened 2025-02-11
Implemented Trigger Event Processor to offload Garden UI Updates

TheBurchLog:


src/app/beer_garden/events/processors.py
```diff
@@ -178,6 +178,95 @@ def queue_depth(self):
         return self._queue.qsize()
 
 
+class TriggerListener(BaseProcessor):
+    """Listener ignore data sent, and
+    triggers the action with no input
```
1maple1: This comment could be a bit clearer about what it is listening for and what it is doing. Basically this triggers the action using the handle_event_for_api_only instead of handle_event which as far as I can tell skips a bunch of unnecessary processing.

1maple1 (approved):
Update comments so it is more clear what this new trigger is doing

1maple1 (approved):

1maple1:
Fix change log

PR #1761, opened 2025-02-11
Optimized the Garden Topic Loop to preload Topics

TheBurchLog:


src/app/beer_garden/topic.py
```diff
@@ -308,26 +312,26 @@ def sync_garden_topics_loop(garden: Garden):
...
+                    subscriber_type="GENERATED",
                 )
 
+                if topic_generated in topics_dict:
+                    if subscriber not in topics_dict[topic_generated].subscribers:
+                        topics_dict[topic].subscribers.append(subscriber)
+                        update_topic(topics_dict[topic])
+                else:
+                    create_topic(Topic(name=topic_generated, subscribers=[subscriber]))
+
```
1maple1, line 346: Wild idea here because this is more or less the same code as above. How about something like:

```
def add_subscriber(subscriber, topic, topics_dict):
	if topic in topics_dict:
		if subscriber not in topics_dict[topic].subscribers:
			topics_dict[topic].subscribers.append(subscriber)
			update_topic(topics_dict[topic])
		else:
			create_topic(Topic(name=topic, subscribers=[subscriber]))
			
add_subscriber(subscriber, topic_generated, topics_dict)
```

src/app/beer_garden/topic.py
```diff
@@ -308,26 +312,26 @@ def sync_garden_topics_loop(garden: Garden):
...
+                    version=system.version,
+                    instance=instance.name,
+                    command=command.name,
+                    subscriber_type="GENERATED",
                 )
 
+                if topic_generated in topics_dict:
+                    if subscriber not in topics_dict[topic_generated].subscribers:
+                        topics_dict[topic].subscribers.append(subscriber)
+                        update_topic(topics_dict[topic])
```
1maple1: topics_dict[topic_generated]???

1maple1 (commented):
Looks fine using dict to avoid adding topic name more than once. See comments.

PR #1794, opened 2025-02-26
Garden remove status and status_info

### LINKED ISSUE #1788: Garden remove Status and Status_Info
In the top level of the Garden model, remove Status and Status Info. The status is now derived from the connections instead. These fields will just cause confusion going forward.

1maple1:
brewtils: garden-remove-status
fixes: #1788


TheBurchLog:
Can this be expanded to also remove Namespace?

PR #1800, opened 2025-02-28
3.30.0 Add garden to topic name

TheBurchLog:


src/app/beer_garden/db/mongo/models.py
```diff
@@ -715,53 +715,72 @@ def clean(self):
...
-                        if len(db_topic.subscribers) == 0:
-                            db_topic.delete()
+        try:
+            if len(self.instances) > 0:
+                if self.local:
+                    garden_name = config.get("garden.name")
+                else:
+                    garden_name = Garden.objects.get(
+                        systems__in=System.objects.filter(id=self.id)
+                    ).name
```
1maple1, line 725: If there will only ever be 1 system with that id, systems_contains=System.objects.get(id=self.id) might be more efficient.

TheBurchLog: It is expecting a list, so the alternative is this:
```python
garden_name = Garden.objects.get(
    systems__in=[System.objects.get(id=self.id)]
).name
```

1maple1 (approved):

PR #1804, opened 2025-03-03
3.30.0 Use Thread Pool to forward operations

TheBurchLog:


src/app/beer_garden/router.py
```diff
@@ -232,7 +232,21 @@ def route(operation: Operation):
...
+        except RuntimeError:
+            pass
+        if loop:
+            result = asyncio.get_event_loop().run_in_executor(
+                t_pool,
+                partial(
+                    initiate_forward,
+                    operation,
+                ),
+            )
```
1maple1, line 247: Can't you just use the loop you already got on line 237? Then line 241 becomes:

```
result = loop.run_in_executor(...)
```

TheBurchLog: Fixed

PR #1826, opened 2025-03-25
OBE: Topics UI

1maple1:


src/ui/src/js/controllers/admin_topic.js
```diff
@@ -0,0 +1,296 @@
...
+            var thd = document.createElement('thead')
+            let columns = ["Garden", "Namespace", "System", "Version", "Instance", "Command", "Consumer Count", "Type"]
+            var tr = document.createElement('tr');
+              columns.forEach((column) => {
+                var th = document.createElement('th');
+                var columnText = document.createTextNode(column);
+                th.append(columnText);
+                tr.append(th);
+              });           
+              thd.append(tr);
```
TheBurchLog, line 127: Is it possible to put a table width on just `Consumer Count` and `Type`? That way those columns always line up at the end. That could help a little with the flow of the page when there is a lot on there.

src/app/beer_garden/api/http/handlers/v1/topic.py
```diff
@@ -138,6 +145,14 @@ async def patch(self, topic_id):
                     )
                 )
 
+            elif operation == "reset_consumer_count":
+                response = await self.client(
+                    Operation(
+                        operation_type="TOPIC_RESET_CONSUMER_COUNT",
+                        kwargs={"topic_id": topic_id, "subscriber": subscriber},
+                    )
+                )
```
TheBurchLog: Can you make subscriber optional, then if no subscriber if provided then it resets the topics counters instead and leaves the subscriber counters alone?

TheBurchLog:
On the UI side, we need to add in permission logic. The Topics page should be allowed for anyone to read. But to access any button you need to have GARDEN_ADMIN. The has permission function should be loaded in by default.

JS:
```
ng-show="hasPermission('GARDEN_ADMIN', true)"
```
Examples on Garden Admin page

We also need this on the HTTP API side of the house. Which that API doesn't have it at all for any of the calls. But we should make a ticket for the API side because it will be a larger lift outside of these changes

src/app/beer_garden/api/http/handlers/v1/topic.py
```diff
@@ -340,3 +553,203 @@ async def post(self):
...
+                            }
+                        )
+                        | Q(
+                            **{
+                                column["data"]
+                                + "__subscriber_type__contains": column["search"][
+                                    "value"
+                                ]
+                            }
+                        )
```
TheBurchLog, line 724: Add one for Consumer Count

src/ui/src/js/controllers/admin_topic.js
```diff
@@ -0,0 +1,296 @@
...
+            });
+            tbl.appendChild(tbdy);
+            return tbl.outerHTML;
+          } else {
+            return "";
+          }
+        }),
+    DTColumnBuilder.newColumn('name').withTitle('').withOption('width', '50px')
+        .renderWith(function(data, type, full) {
+          let subscribers = full.subscribers;
```
TheBurchLog: Name does fit with the intent of the column.

Also, add this to prevent the API from trying to generate a filter for it.
JS:
```javascript
.withOption('searchable', false)
```

src/app/beer_garden/api/http/handlers/v1/topic.py
1maple1: Is there value in searching for consumer count? If someone searches on a random number like 0 or 1, they are more likely to get matches for version.

src/ui/src/js/controllers/admin_topic.js
1maple1: Changed column name to null and disabled search and sort.

src/ui/src/js/controllers/admin_topic.js
1maple1: Added `min-width` so buttons line up as expected for `Consumer Count` and `Type`. Also added div with `overflow:auto` so table stays inside column.

src/app/beer_garden/api/http/handlers/v1/topic.py
1maple1: Done

src/app/beer_garden/api/http/handlers/v1/topic.py
1maple1: Add consumer count, but it has to be an exact match.

TheBurchLog:
OBE, but won't delete until we migrate functionality to React UI

PR #1828, opened 2025-03-26
3.30.0 Remove the configuration option for event handlers

TheBurchLog:
brewtils: migrate_handler_optimization

### DIFF
--- a/src/app/beer_garden/events/handlers.py
+++ b/src/app/beer_garden/events/handlers.py
@@ -3,7 +3,6 @@ import logging
 
 from brewtils.models import Events
 
-import beer_garden.config as config
 import beer_garden.files
 import beer_garden.garden
 import beer_garden.local_plugins.manager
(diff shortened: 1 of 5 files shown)


src/app/beer_garden/events/handlers.py
```diff
@@ -95,7 +90,6 @@ def add_internal_events_handler(event_manager):
             "Requests wait events",
             False,
             [Events.REQUEST_COMPLETED, Events.REQUEST_CANCELED, Events.GARDEN_STOPPED],
-            False,
             True,
```
justinryan1, line 93: Think this one might be False and we cannot unique these events

TheBurchLog: I have gone back and forth on this one. If someone is listening for this event, they are only going to get one response. So we should give them the latest. We are also only returning a completed status, so the in progress and created events don't matter.

1maple1: The explanation on this one makes sense to me.

1maple1 (approved):

PR #1835, opened 2025-04-01
3.30.0 Create script for beerctl

1maple1:


TheBurchLog:
Putting it in the profile limits us to only ssh/user interactions with it. I believe you can't schedule scripts to run against stuff managed this way. Instead, we should have a script that we can pass in different arguments. Then drop that in `/usr/bin`. 

First pass should support:
- `beerctl start`
- `beerctl stop`
- `beerctl restart`

After that, we can look at other options but that would be outside the scope of this.

TheBurchLog (approved):

PR #1841, opened 2025-04-08
3.30.0 Fixed Garden Delete Calls

TheBurchLog:
brewtils: migrate_handler_optimization


1maple1 (approved):
Nice fix. This skips the recursion in remove_remote_systems because it will already get the child when remove_garden calls itself. Extra bonus for eliminating the name lookup too.

PR #1845, opened 2025-04-11
3.30.0 Accept Target Garden for multi hops

TheBurchLog:
Now that we know of the entire path, we can accept the provided Target Garden from the Upstream Garden.

This resolves the issues seen with: https://github.com/beer-garden/beer-garden/issues/1076


1maple1 (approved):
Looks good. It seems straightforward to use the target_garden provided outright instead of trying to calculate it.

PR #1857, opened 2025-04-24
3.30.0 Fix topic subscriber checks

TheBurchLog:


1maple1 (commented):
This should fix the Subscriber model `__eq__` method instead

1maple1:
```def __eq__(self, other):
        if isinstance(other, self.__class__):
            return (
                self.subscriber_type == other.subscriber_type
                and self.garden == other.garden
                and self.namespace == other.namespace
                and self.system == other.system
                and self.version == other.version
                and self.instance == other.instance
                and self.command == other.command
            )
        return False
```

1maple1 (approved):

PR #1863, opened 2025-05-01
3.30.0 Use Delete Many for Request pruning

TheBurchLog:


src/app/beer_garden/db/mongo/pruner.py
```diff
@@ -21,199 +20,334 @@
...
+):
+    """
+    Helper function to prune a cursor of requests
+    """
 
-            try:
+    batch_ids = []
+    for request in request_cursor:
+        try:
+            if request.id in request_ids:
```
1maple1, line 104: Need to be careful with this. Passing in request_ids and resetting request_ids when number of request_ids > batch_size.

src/app/beer_garden/db/mongo/pruner.py
```diff
@@ -21,199 +20,334 @@
...
-                            ttl_name,
-                            task["collection"].__name__,
-                            str(delete_older_than),
-                        )
-                    )
+        except DoesNotExist:
+            logger.error(
+                f"DoesNotExist: Attempted to delete request {request.id} "
+                "but does not exist in database"
+            )
```
1maple1, line 152: Where is the delete for a single request? See only delete_requests for batch_size.

TheBurchLog: That happens line 89, we only delete in the loop if we exceed the batch size.

PR #1869, opened 2025-05-08
3.30.0 Use command_display_name for Request indexes

1maple1:


src/app/beer_garden/db/mongo/models.py
```diff
@@ -381,7 +381,10 @@ class Request(MongoModel, Document):
             # Used for Gridfs File Pruning
             {"name": "gridfs_index", "fields": ["output_gridfs", "parameters_gridfs"]},
             # These are for sorting parent requests
-            {"name": "parent_command_index", "fields": ["has_parent", "command"]},
+            {
+                "name": "parent_command_index",
+                "fields": ["has_parent", "command_display_name"],
+            },
```
TheBurchLog: The names have to change as well for all of them, else we won't know to auto reindex them at startup.

1maple1: done

TheBurchLog (approved):

PR #1882, opened 2025-05-21
3.30.0 Upstream Event filtering

TheBurchLog:


src/app/beer_garden/api/stomp/manager.py
```diff
@@ -181,15 +181,36 @@ def _event_handler(self, event):
...
-                conn = value["conn"]
-                if conn:
-                    if conn.is_connected():
-                        if value["headers_list"]:
-                            for headers in value["headers_list"]:
-                                conn.send(event, headers=headers)
-                        else:
-                            conn.send(event)
+            # Keep filter list in sync with HTTP Parent Updater
+            if event.name in (
```
justinryan1: This list as well as the one in app.py should probably be extracted out and centralized since a comment to keep them both in sync seems to be prone to errors down the road.

PR #1885, opened 2025-05-22
3.30.0 Add event filtering for Entry Points

TheBurchLog:


src/app/beer_garden/api/http/__init__.py
```diff
@@ -419,13 +419,33 @@ def _event_callback(event):
...
+
+        elif event.name in [
+            Events.REQUEST_COMPLETED.name,
+            Events.REQUEST_CANCELED.name,
+            Events.GARDEN_STOPPED.name,
+        ]:
+            try:
+                beer_garden.requests.handle_wait_events(deepcopy(event))
+            except Exception as ex:
+                logger.exception(f"Error executing callback for {event!r}: {ex}")
```
1maple1: Some duplication of code. Why not wrap the whole thing in a try - except?

src/app/beer_garden/router.py
```diff
@@ -686,10 +684,16 @@ def handle_event(event):
             Events.GARDEN_UPDATED.name,
         ]
     ):
+
         # Only store the garden if it's 1 hop of the local garden
         if not event.payload.has_parent or event.payload.parent == config.get(
             "garden.name"
         ):
+            # To save memory, we need to remove children
+            if event.payload_type == "Garden":
```
1maple1, line 693: Is there a situation where payload_type will not be Garden for Garden events? Otherwise looks good to have this with the Garden events.

src/app/beer_garden/router.py
```diff
@@ -713,9 +717,10 @@ def handle_event(event):
                                 )
 
                             elif connection.status == "DISABLED":
-                                stomp_garden_connections[
-                                    event.payload.name
-                                ].disconnect()
+                                if event.payload.name in stomp_garden_connections:
```
1maple1, line 720: No action required, just a good defensive check prior to accessing.

TheBurchLog: If you have a STOMP connection that is disabled that has not been seen enabled, then it will throw an error.

src/app/beer_garden/router.py
TheBurchLog: All of them should be Garden at this point, it was just me being defensive.

1maple1 (approved):

PR #1888, opened 2025-05-28
Request event optimization

TheBurchLog:


src/app/beer_garden/requests.py
```diff
@@ -1072,6 +1073,49 @@ def handle_event_filter(event):
...
+
+        # User mappings back to local usernames
+        if event.payload.requester and config.get("auth.enabled"):
+            foundUser = False
+
+            # First try to grab requester from Parent Request
+            if event.payload.has_parent:
+                parent_requests = db.query(
+                    Request,
+                    filter_params={"id": event.payload.parent.id},
```
1maple1: Does it make sense to filter on id vs query unique? Don't really need a list here and this takes the first element anyway.

src/app/beer_garden/requests.py
```diff
@@ -1099,119 +1143,107 @@ def handle_event(event):
...
-
-                if getattr(existing_request, field) != new_value:
-                    request_changed[field] = new_value
-                    setattr(existing_request, field, new_value)
-
-            # Add output fields only if the status changes to a compelted state
-            if "status" in request_changed:
-                if event.payload.status in (
+                    metadata[f"{event.payload.status}_{config.get('garden.name')}"] = (
+                        int(datetime.datetime.utcnow().timestamp() * 1000)
```
1maple1: Is there even a remote chance that this could overwrite an existing value? Or is updating the timestamp desired in all cases? Also stop using utcnow() in new code.

PR #1890, opened 2025-06-02
Garden Event Handler Performance

TheBurchLog:


src/app/beer_garden/db/mongo/models.py
```diff
@@ -1213,10 +1213,15 @@ def _update_associated_systems(self):
 
         def _get_system_triple(system: System) -> Tuple[str, str, str]:
             namespace = getattr(system, "namespace", None)
+            name = getattr(system, "name", None)
+            version = getattr(system, "version", None)
+            if not name or not version:
+                # dbref doesn't exist
+                return str(system)
```
1maple1: This doesn't return a Tuple as indicated

PR #1892, opened 2025-06-02
Garden update connection and Request Update Gridfs

TheBurchLog:
Removed duplicate db.update for Garden API status changes

Ensure gridfs files are passed forward to a save of Request, but only during Updates.

Updated event filtering


src/app/beer_garden/db/mongo/models.py
```diff
@@ -464,149 +467,151 @@ class Request(MongoModel, Document):
...
-                # Unable to find parent, remove object to allow brewtils serializing
-                self.parent = None
+            if self.parameters_gridfs:
+                logger.debug("Retrieving parameters from GridFS")
+                self.parameters = json.loads(
+                    self.parameters_gridfs.read().decode(encoding)
+                )
+                self.parameters_gridfs = None
+
+            if self.parent:
```
justinryan1: Copy logic from line 541

src/app/beer_garden/db/mongo/models.py
```diff
@@ -1218,7 +1237,7 @@ def _get_system_triple(system: System) -> Tuple[str, str, str]:
             version = getattr(system, "version", None)
             if not name or not version:
                 # dbref doesn't exist
-                return (str(system), None, None)
+                return (str(system), "unknown", "unknown")
```
1maple1: This seems a bit arbitrary. Why can't the receiving function take the None and use that to make a determination?

src/app/beer_garden/requests.py
```diff
@@ -1073,8 +1087,26 @@ def handle_event_filter(event):
...
+    # If no parent are set, we only want to publish to the UI events handler
+    if not config.get("parent.stomp.enabled") and not config.get("parent.http.enabled"):
+        metadata["API_ONLY"] = True
+    publish(
+        Event(
+            name=event_name, payload=request, payload_type="Request", metadata=metadata
+        )
+    )
+
+
```
1maple1, line 1102: Should this always publish?

TheBurchLog: Yes, the event handler will process the downstream events, but this will republish them as a local event for the UI and Request Wait Event handler to process it

PR #1894, opened 2025-06-04
Fixed 3.30 migration script for missing command type

TheBurchLog:


src/app/beer_garden/db/mongo/util.py
```diff
@@ -336,7 +336,7 @@ def find_root_command_type_and_expiration(request):
                 minutes=config.get("db.prune.ttl.info", default=-1)
             )
 
-    return request["command_type"], expiration_at
+    return getattr(request, "command_type", "ACTION"), expiration_at
```
1maple1: Why not just get the command type at the beginning if the intent is to always return it?

src/app/test/db/mongo/util_test.py
```diff
@@ -91,6 +91,34 @@ def test_3_30_request_migration(self, request_dict, ts_dt):
...
 
+    @patch("mongoengine.connect", Mock())
+    @patch("mongoengine.register_connection", Mock())
+    def test_3_30_request_migration_no_command_type(self, request_dict, ts_dt):
+
+        config._CONFIG = {"db": {"prune": {"batch_size": -1, "ttl": {"action": 0}}}}
+
+        del request_dict["id"]
+        del request_dict["root_command_type"]
+        # del request_dict["command_type"]
```
1maple1: Remove if not needed

PR #1904, opened 2025-06-30
Update expiration at for TEMP children on completed status

1maple1:


src/app/beer_garden/db/mongo/models.py
```diff
@@ -578,6 +578,8 @@ def _pre_save(self):
                 self.root_command_type = self.command_type
 
     def _set_child_expiration(self):
+        if self.command_type == "TEMP":
+            self.expiration_at = datetime.datetime.now(datetime.timezone.utc)
```
TheBurchLog: This won't update the the `self` instance of the request for `expiration_at`. We need to update the `_pre_save` function to check the `has_parent` or is `TEMP`.

src/app/beer_garden/db/mongo/models.py
```diff
@@ -601,13 +603,13 @@ def _post_save(self):
...
-            and self.expiration_at
-            and self.status in BrewtilsRequest.COMPLETED_STATUSES
-        ):
-
-            self._set_child_expiration()
+        if self.status in BrewtilsRequest.COMPLETED_STATUSES:
+            if (
+                not self.has_parent
+                and self.expiration_at
+                or self.command_type == "TEMP"
```
TheBurchLog: This added check for TEMP won't be needed if we fix it in pre_save

src/app/beer_garden/db/mongo/models.py
1maple1: I can agree with you on this point. Adding TEMP command_type to pre-save will do the same thing.

src/app/beer_garden/db/mongo/models.py
1maple1: Pre-save will only work on the current request and will not update the children expiration the way this does. If TEMP has any children they will not be updated. Should they? We could change to `expiration_at OR ...` so existing expiration datetimes filter to children.

src/app/beer_garden/db/mongo/models.py
```diff
@@ -603,13 +605,12 @@ def _post_save(self):
...
                     pass
 
-        if self.status in BrewtilsRequest.COMPLETED_STATUSES:
-            if (
-                not self.has_parent
-                and self.expiration_at
-                or self.command_type == "TEMP"
-            ):
-                self._set_child_expiration()
+        if (
```
TheBurchLog, line 601: I think this needs to be done as:
```
if (
            (
                self.expiration_at
                or not self.has_parent
             )
            and self.status in BrewtilsRequest.COMPLETED_STATUSES
        ):
```

src/app/beer_garden/db/mongo/models.py
```diff
@@ -547,7 +547,11 @@ def _pre_save(self):
 
         if not self.expiration_at and self.status in BrewtilsRequest.COMPLETED_STATUSES:
```
TheBurchLog, line 542: It might be slightly out of scope of the original thing, but could you add this to the end of the block. So if the parent is already completed or set, that it goes ahead and grabs it's expiration date.
```python
            if self.has_parent and not self.expiration_at:
                parent = Request.objects(id=self.parent.id).only("expiration_at")
                if parent["expiration_at"]:
                    self.expiration_at = parent["expiration_at"]
```

TheBurchLog (approved):

PR #1908, opened 2025-07-02
Garden sync heartbeat

TheBurchLog:


src/app/beer_garden/garden.py
```diff
@@ -248,6 +255,12 @@ def check_garden_receiving_heartbeat(
         for connection in garden.receiving_connections:
             if connection.api == api:
                 connection_set = True
+                if connection.status != "DISABLED":
```
1maple1: Should this also check for NOT_CONFIGURED?

PR #1911, opened 2025-07-08
Multiprocessing Event Handler

TheBurchLog:


src/app/beer_garden/events/processors.py
```diff
@@ -227,6 +228,173 @@ def queue_depth(self):
...
+
+    def __init__(
+        self,
+        handler,
+        local_only=False,
+        filters=None,
+        filter_func=None,
+        allow_api_only=False,
+        unique_data=False,
+        num_workers=5,
```
1maple1, line 297: Should this be consistent with the 10 workers for the event manager?

1maple1 (commented):
Looks fine but not sure about removal of deduping.

PR #1913, opened 2025-07-09
Moved Orphan Check for Requests

TheBurchLog:


src/app/beer_garden/db/mongo/models.py
```diff
@@ -533,33 +533,40 @@ def _pre_save(self):
 
         if self.has_parent:
             try:
-                self.parent
-            except DoesNotExist:
+                if not self.parent or Request.objects(id=self.parent.id).count() == 0:
+                    # Request is an Orphan, removing parent
+                    self.has_parent = False
+                    self.parent = None
+            except (DoesNotExist, TypeError):
```
1maple1: Duplicated code removing parent. This could just throw a DoesNotExist, but will count() throw a DoesNotExist or give 0 for an invalid id? Where does TypeError fit in?

PR #1920, opened 2025-07-17
Direct DB Create/Update to Collections for Request Handler

TheBurchLog:


justinryan1:
Still throws duplicate key errors when doing events of the type REQUEST_STARTED and REQUEST_CREATED

PR #1929, opened 2025-07-25
Broadcast Remote syncs as UPSERT for routing logic

TheBurchLog:
brewtils: upsert_garden


src/app/requirements-dev.txt
```diff
@@ -4,7 +4,7 @@
 #
 #    pip-compile requirements-dev.in
 #
--e git+https://github.com/beer-garden/brewtils@develop#egg=brewtils
+-e git+https://github.com/beer-garden/brewtils@upsert_garden#egg=brewtils
```
TheBurchLog, line 7: Don't forget to remove this!

src/app/beer_garden/garden.py
```diff
@@ -517,6 +517,14 @@ def update_garden(garden: Garden) -> Garden:
     return db.update(garden)
 
 
+@publish_event(Events.GARDEN_UPSERT)
+def upsert_downstream_garden(garden: Garden):
+
+    upsert_garden(garden)
+
+    return get_garden(garden.name, exclude_fields=["systems__instances__status_info"])
```
1maple1: Just an idea, upsert_garden already returns the garden so is the extra query necessary?

Not as concise but I think this would do the same:
```
garden = upsert_garden(garden)

for system in garden.systems:
  for instance in system.instances:
    del instance.status_info

return garden
```

src/app/beer_garden/router.py
```diff
@@ -671,29 +669,43 @@ def add_routing_garden(garden: Garden):
...
         for child in garden.children:
             add_routing_garden(child)
+    del garden.children
+
+    gardens[garden.name] = garden
+
+
+def handle_event_filter(event):
+
+    # Only accept locally processed garden updates
```
1maple1: Locally processed? Downstream?

1maple1 (approved):

PR #1938, opened 2025-07-30
Fix null check for children for isRoutable

TheBurchLog:


1maple1 (approved):
Looks good wrapping those in an if statement

PR #1940, opened 2025-08-06
Index Management and Mongo TTL pruning

TheBurchLog:
brewtils: file_created_at


1maple1 (approved):
Pipelines and calling the db inside models were significant changes, but seem to be implemented ok. Worth noting that I had difficulty upgrading mongo 4.2 to mongo 6.0 but this can probably be mitigated with a proper upgrade strategy.

src/app/beer_garden/db/mongo/models.py
```diff
@@ -507,17 +511,6 @@ def _pre_save(self):
         # we opt to just pull the Request as it exists in the database so that we can
         # check those gridfs field.
 
-        if self.parameters_gridfs.grid_id:
```
justinryan1, line 510: Moving the parameters_gridfs and output_gridfs code listed below further down now makes the code comment about issue 1216 not line up with the correct section

src/app/beer_garden/db/mongo/models.py
```diff
@@ -561,11 +554,94 @@ def _pre_save(self):
...
+            self.output = None
+
     def _post_save(self):
 
         if self.status == "CREATED":
             self._update_raw_file_references()
 
+        self._update_raw_file_pruning()
+
+    def _update_raw_file_pruning(self):
```
justinryan1: Think it would be worthwhile to rename this function

src/app/beer_garden/db/mongo/models.py
```diff
@@ -1317,8 +1412,8 @@ class File(MongoModel, Document):
 
     owner_id = StringField(required=False)
     owner_type = StringField(required=False)
-    request = LazyReferenceField(Request, required=False, reverse_delete_rule=NULLIFY)
-    job = LazyReferenceField(Job, required=False, reverse_delete_rule=NULLIFY)
+    request = LazyReferenceField(Request, required=False)
```
justinryan1: I'm not sure we can immediately remove this functionality while there is still support for the legacy_pruner code to deal with systems still utilizing MongoDB versions below 6.x

src/app/beer_garden/db/mongo/models.py
```diff
@@ -561,11 +545,108 @@ def _pre_save(self):
...
+                )
+
+                get_db()["fs.chunks"].update_many(
+                    {"files_id": self.output_gridfs.grid_id},
+                    {
+                        "$set": {
+                            "status": self.status,
+                            "created_at": self.created_at,
+                            "root_command_type": self.root_command_type,
+                            "uploadDate": self.created_at,
```
justinryan1: This field name still needs to be changed

src/app/beer_garden/db/mongo/util.py
```diff
@@ -411,12 +417,164 @@ def reset_last_configuration():
...
         action_ttl=config.get("db.prune.ttl.action", default=-1),
-        info_ttl=config.get("db.prune.ttl.info", default=-1),
+        info_ttl=config.get("db.prune.ttl.info", default=15),
+        file_ttl=config.get("db.prune.ttl.info", default=15),
         version=str(beer_garden.__version__),
     )
     configuration.save()
 
 
+def update_request_ttl_indexes(command_type, ttl, previous_ttl=-1):
```
justinryan1: The previous_ttl default of -1 here and also within update_file_ttl_indexes seem a little counter intuitive

src/app/beer_garden/db/mongo/util.py
```diff
@@ -411,12 +417,164 @@ def reset_last_configuration():
...
+                name=file_raw_index,
+                expireAfterSeconds=ttl * 60,
+                partialFilterExpression={
+                    "request": None,
+                },
+            )
+
+        if file_index not in db["file"].index_information():
+            db["file"].create_index(
+                [("updated_at", 1)],
```
justinryan1, line 907: Just making the comment so we don't forget. Index name is created_at verse the field we are triggering on being updated_at. We will have to decide the best way to normalize this across the RawFile, File, and FileChunk models in both BG and brewtils before this is completed.

src/app/beer_garden/db/mongo/util.py
```diff
@@ -616,3 +797,163 @@ def prune_topics():
...
+                file_ids_str.append(str(doc["_id"]))
+
+    if len(file_ids) > 0:
+        batch_size = config.get("db.prune.batch_size")
+
+        if batch_size > 0:
+            for i in range(0, len(file_ids), batch_size):
+                File._get_collection().update_many(
+                    {"_id": {"$in": file_ids[i : i + batch_size]}},
+                    {"$unset": {"job": "", "request": ""}},
```
justinryan1: For completeness should we also unset the owner_id and owner_type fields? These File and FileChunk models have some messy extra keys that are either legacy or completely unused right now.

src/app/beer_garden/db/mongo/util.py
```diff
@@ -616,3 +797,163 @@ def prune_topics():
...
+
+        if batch_size > 0:
+
+            outstanding_requests = (
+                Request.objects.filter(
+                    status__in=["IN_PROGRESS", "CREATED"],
+                    updated_at__lte=timeout,
+                )
+                .order_by("-updated_at")
+                .batch_size(batch_size)
```
justinryan1: The batch_size usage here is incorrect and needs to be updated.

src/app/beer_garden/db/mongo/util.py
```diff
@@ -616,3 +797,163 @@ def prune_topics():
...
+            try:
+                request.status = "CANCELED"
+                request.status_updated_at = datetime.now(timezone.utc)
+                request.save()
+                serialized = MongoParser.serialize(request, to_string=True)
+                parsed = SchemaParser.parse_request(
+                    serialized, from_string=True, many=False
+                )
+
+                publish(
```
justinryan1, line 1287: Just making a note that #1946 has the changes necessary in to publish the minimum number of fields for a canceled request.

src/app/beer_garden/scheduler.py
```diff
@@ -388,97 +390,25 @@ def add_schedule(self, func, interval=None, **kwargs):
...
-            beer_garden.db.mongo.pruner.prune_orphan_command_type_admin,
-            interval=config.get("db.prune.interval", default=15),
-            max_instances=1,
-            name="prune_orphan_command_type_admin",
-        )
+            self.add_schedule(
+                beer_garden.db.mongo.util.unassign_files,
+                interval=prune_interval,
+                max_instances=1,
+                name="prune_files",
```
justinryan1: Name update

src/app/beer_garden/scheduler.py
```diff
@@ -507,6 +437,84 @@ def internal_scheduled_jobs(self):
...
+            )
+
+            self.add_schedule(
+                beer_garden.db.mongo.legacy_pruner.prune_orphan_command_type_admin,
+                interval=config.get("db.prune.interval", default=15),
+                max_instances=1,
+                name="prune_orphan_command_type_admin",
+            )
+
+            if prune_interval > -1:
```
justinryan1, line 482: Looking at all the interval parameters above we are taking the configuration value of the db.prune.interval to set the scheduling for temp, admin, etc. We can technically set the interval to -1 in the config so all of those scheduled tasks above would end up getting their interval set as -1 by default.

src/app/beer_garden/db/mongo/util.py
TheBurchLog: We can just not provide a default for them since we always pass in some sort of value.

src/app/beer_garden/db/mongo/models.py
TheBurchLog: By BG unsetting the values, the legacy pruner logic will still find these records to prune them out (`prune_files()`). So we should be covered on both sides of the releases.

Except for `FileChunk`, we unset the owner. So if we make that conditional based on the version of Mongo then both can operate.

TheBurchLog: I made File Chunk conditional based on Mongo Version to allow Legacy support to work

src/app/beer_garden/scheduler.py
```diff
@@ -507,6 +435,81 @@ def internal_scheduled_jobs(self):
                 name="publish_garden",
             )
 
+        if beer_garden.db.mongo.util.is_legacy_mongodb() and prune_interval > -1:
```
justinryan1: I believe the expected behavior we need here is to still run the following commands regardless if a pruning interval is set: prune_admin_requests, prune_temp_requests, prune_missed_temp_command, and prune_orphan_command_type_admin. Because the framework no longer exposes a way to define their pruning and we said these were internal request types the framework would handle. So maybe just have them run at a default of 15 minutes?

src/app/beer_garden/db/mongo/util.py
TheBurchLog: We can just drop the batch feature because all of the requests are handled sequentially and individually. Unlike pruners where BG just batch deletes them.

PR #1943, opened 2025-08-07
Parent none check

TheBurchLog:


1maple1 (commented):
There is another call to parent.id in requests.py. Should that have a parent none check?

PR #1951, opened 2025-08-19
Remove job counter update event handling from ui

1maple1:


TheBurchLog (approved):
The Change log if kinda weird since it is something we added in 3.30 that we are removing before it was officially released. So we might want to bring up the lower change log to the current line items and list both PRs unto the original change since it does not mention the Job Counters

PR #1953, opened 2025-08-20
Fix blocking bug for Processor Exceptions

TheBurchLog:


src/app/beer_garden/publish_request.py
```diff
@@ -210,6 +211,13 @@ def route_request(create_request):
                 model_type="Request",
             )
         )
+    except ModelValidationError as ex:
+        logger.error(
+            (
+                f"Invalid request for topic '{create_request.metadata['topic']}' "
```
1maple1: Probably should check for metadata['topic'] first in case it doesn't exist.

PR #1966, opened 2025-08-26
Split apart DB Util functions and fixed previous config saving

TheBurchLog:
- Splits apart the mongo util class
- Added logging for index creation
- Moved previous config saving to after all migration logic is completed


src/app/beer_garden/db/mongo/indexes.py
```diff
@@ -0,0 +1,361 @@
...
+            logger.warning(f"Created new {file_chunk_index} index")
+
+
+def update_file_ttl_indexes(ttl, previous_ttl):
+
+    from mongoengine.connection import get_db
+
+    db = get_db()
+
+    file_index = "created_at_file_index_ttl"
```
justinryan1: These two index names should start with updated_at for consistency

PR #1968, opened 2025-08-29
Fixed bug assigning parent to publish topic and handle save check for Parents

TheBurchLog:


1maple1 (approved):
Looks good wrapping the logic in a try except and defaulting has_parent to True.

PR #1971, opened 2025-09-08
Garden sync refactor

### LINKED ISSUE #1822: Garden Admin Page Sync Directories
From the Garden Admin Page add features to rescan downstream directories for Plugins and Downstream Gardens.

### LINKED ISSUE #1847: Forward Admin Actions to Downstream Gardens
Now that we can pass forward the Target Garden in operations, for admin actions add the `target_garden` arguments to the HTTP calls so they can invoke actions against downstream gardens instead of local.

Actions:
Rescan System Directory
Clear All Queues
Rescan Downstream Directory

1maple1:
fixes: #1822 
fixes: #1847


src/ui/src/js/controllers/admin_garden.js
```diff
@@ -401,6 +404,10 @@ export default function adminGardenController(
         if (event.payload.name != $scope.config.gardenName) {
           $scope.pushAlert('info', 'Processing Garden sync event from ' + event.payload.name);
         }
+      case 'GARDEN_RESCAN':
+        if (event.payload.name != $scope.config.gardenName) {
+          $scope.pushAlert('info', 'Processing Garden rescan event from ' + event.payload.name);
+        }
```
TheBurchLog: `GARDEN_RESCAN` is not an event type that we generate

1maple1: Right not an event - Removed

src/app/beer_garden/router.py
```diff
@@ -905,10 +905,6 @@ def _determine_target(operation: Operation) -> str:
 def _target_from_type(operation: Operation) -> str:
     """Determine the target garden based on the operation type"""
 
-    # If target garden is provided then return that
-    if operation.target_garden_name:
-        return operation.target_garden_name
-
     # Certain operations are ASSUMED to be targeted at the local garden
     if operation.operation_type not in routable_operations:
```
TheBurchLog, line 911: Why was this removed? Feels like this defeats the point of providing it.

1maple1: This line is duplicated in _determine_target and _target_from_type. Would you rather have the other one removed so _target_from_type is always called?

TheBurchLog: I see that now. This makes sense to remove it from here because it was code that would never be executed.

src/ui/src/partials/admin_garden_index.html
```diff
@@ -102,8 +102,16 @@ <h1 class="page-header">
...
+          <button type="button" class="btn btn-default fa fa-bars dropdown-toggle" data-toggle="dropdown"
+            title="Garden Menu"></button>
+          <ul class="dropdown-menu">
+            <li>
+              <a href="" ng-click="syncGarden(garden)" title="Sync Garden {{garden.name}}">Sync Garden</a>
+            </li>
+            <li>
+              <a href="" ng-click="rescanGarden(garden)" title="Rescan Plugins {{garden.name}}">Rescan Plugins</a>
+            </li>
+          </ul>
```
TheBurchLog: These are two different rescans. `GARDEN_RESCAN` is to evaluate the children directory and `SYSTEM_RESCAN` is to evaluate the plugins directory. It looks like you got them mixed up here.

1maple1: Added both rescans to Garden Admin page.

src/ui/src/js/services/garden_service.js
```diff
@@ -51,6 +51,16 @@ export default function gardenService($rootScope, $http) {
...
 
+  GardenService.rescanGardenPlugins = function(name) {
+    return $http.patch('api/v1/gardens/' + encodeURIComponent(name), {
+      operation: 'rescan_plugins',
+      path: '',
+      value: '',
+    },{
+      headers: {'Target-Garden': name}
+    });
+  };
```
TheBurchLog: This currently exists in `admin_services.js` and `api/v1/admin`, so we should pick one entry point for rescanning plugin directory

1maple1: The rescan in admin_services.js does not take a garden name. It corresponds to the button on the Garden Admin page.

The rescan in garden.services.js applies to a specific garden using garden name in the dropdown Garden Menu.

What functions should go in admin vs garden?

1maple1: They are all different. Admin has RUNNER_RESCAN. Garden has GARDEN_RESCAN and SYSTEM_RESCAN.

TheBurchLog:
The 3 operations added to Garden API already exist via other services. The primary difference is that the new operations support the ability to target the garden. Instead of adding new operation support in Garden API, update the other APIs to support accepting a Garden Name via query parameters. That way we don't have multiple points of entry for the same operation causing confusion down the road.

Then in those associated Javascript Services add in the ability to pass the Garden Name, if a garden name is not provided then it doesn't forward one, so the API will default to local. I have done that else where when we create an empty dict for the query args, then if it isn't null or unidentified then we add it to the query args dict to pass it to the call.

Then the Garden Admin controller just needs to import those services to call the correct APIs.

src/app/beer_garden/api/http/handlers/v1/admin.py
```diff
@@ -49,7 +50,7 @@ async def patch(self):
           50x:
             $ref: '#/definitions/50xError'
         tags:
-          - Admin
+          - Deprecated
```
justinryan1, line 53: This is not currently listed as deprecated in develop so for the scope of this feature it would be better to not include it.

src/app/beer_garden/api/http/handlers/v1/queue.py
```diff
@@ -82,9 +88,21 @@ async def delete(self):
...
+
+        self.verify_user_permission_for_object(garden)
 
-        await self.process_operation(Operation(operation_type="QUEUE_DELETE_ALL"))
+        await self.process_operation(
+            Operation(
+                operation_type="QUEUE_DELETE_ALL",
+                target_garden_name=garden.name,
+            )
+        )
```
TheBurchLog, line 104: I would re-write it as this because `get_or_raise` already does the permission check:
```python
        if garden_name:
            self.get_or_raise(Garden, name=garden_name)
        else:
            self.verify_user_permission_for_object(local_garden())

        await self.process_operation(
            Operation(
                operation_type="QUEUE_DELETE_ALL",
                target_garden_name=garden_name,
            )
        )
```

src/ui/src/js/controllers/admin_garden.js
```diff
@@ -130,7 +136,7 @@ export default function adminGardenController(
   };
 
   $scope.deleteQueues = function(garden) {
-    GardenService.deleteQueues(garden.name);
+    QueueService.clearQueues(garden.name);
```
TheBurchLog, line 139: Should this be `deleteQueues`?

1maple1: QueueService already has clearQueues. Renamed deleteQueues to clearQueues in garden admin to match.

src/app/beer_garden/api/http/handlers/v1/queue.py
1maple1: Done

TheBurchLog (approved):

PR #1972, opened 2025-09-09
Bug fix for update associated systems after sync

1maple1:
Fixes ObjectId comparison causing errors similar to`Removing System <parent, complex, 3.0.0.dev0> with ID=68c02b947565a299ef996d95; doesn't match ID=68c02b947565a299ef996d95 for known system with same attributes`


TheBurchLog (approved):
Add Change Log

PR #1983, opened 2025-11-12
Updated plugin read logs UI

TheBurchLog:
brewtils: read_logs_update

### DIFF
--- a/src/app/beer_garden/api/http/handlers/v1/instance.py
+++ b/src/app/beer_garden/api/http/handlers/v1/instance.py
@@ -258,8 +258,11 @@ class InstanceLogAPI(AuthorizationHandler):
 
         if self.get_query_argument("logs_only", default="").lower() == "true":
             if response.output:
-                output = json.loads(response.output)
-                self.write(output["logs"])
+                try:
+                    output = json.loads(response.output)
+                    self.write(output["logs"])
+                except json.JSONDecodeError:
+                    self.write(response.output if response.output else "")
             else:
                 self.write("")
         else:
--- a/src/app/beer_garden/api/http/handlers/v1/instance.py
+++ b/src/app/beer_garden/api/http/handlers/v1/instance.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+import json
 from asyncio import Future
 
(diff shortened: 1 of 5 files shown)


TheBurchLog:
image

image

src/app/beer_garden/api/http/handlers/v1/instance.py
```diff
@@ -211,6 +212,12 @@ async def get(self, instance_id):
             required: false
             description: End line of logs to read from instance
             type: int
+          - name: logs_only
+            in: query
+            required: false
+            description: Return only the log content
+            type: boolean
+            default: false
```
1maple1, line 220: Is this nececessary? I understand the format changed to return a dict/json as a string. But it is only used once below and makes the following code more complicated than necessary.

src/app/beer_garden/api/http/handlers/v1/instance.py
```diff
@@ -248,7 +255,18 @@ async def get(self, instance_id):
 
         self.set_header("request_id", response.id)
         self.set_header("Content-Type", "text/plain; charset=UTF-8")
-        self.write(response.output if response.output else "")
+
+        if self.get_query_argument("logs_only", default="").lower() == "true":
+            if response.output:
```
1maple1, line 260: If logs_only condition were not here then shouldn't the try/except catch it if it's not proper json? The assumption is that it is always true in the new code. I believe this will only come into play if brewtils is lagging behind beer-garden.

1maple1 (commented):
Not sure about logs_only and the utility it provides

src/app/beer_garden/api/http/handlers/v1/instance.py
TheBurchLog: We need this to support download logs file so the `.log` file isn't actually a json file

1maple1: Ignore comments above then

src/app/beer_garden/api/http/handlers/v1/instance.py
1maple1: Ignore comments

1maple1 (approved):
Good with logs_only

PR #1985, opened 2025-11-19
3.31 Release Candidate 

### LINKED ISSUE #291: Open API v3
We are on Open API v2.0 because 3.0 didn't exist before. We should update to the latest open API spec because there are better ways to describe the authentication schema and there are new ways to handle file uploads.

TheBurchLog:
fixes: #291


TheBurchLog:
TODO: Remove brewtils hard coded from requirements-dev.txt

PR #1987, opened 2025-11-20
Improved UI handling for missing permissions

TheBurchLog:


src/ui/src/partials/system_index.html
```diff
@@ -1,6 +1,12 @@
...
 
-<div ng-if="responseState(response) === 'success'">
+<div ng-if="!hasPermission('READ_ONLY')">
+  <h1>
+    <span>401: Unauthorized to access</span>
+  </h1>
+</div>
+
+<div ng-if="hasPermission('READ_ONLY') && responseState(response) === 'success'">
   <h1 class="page-header">Systems</h1>
```
1maple1: Should this page-header be moved up so it shows the heading even if a user is unauthorized? This only impacts Systems.

1maple1 (commented):
I think `401: Unauthorized to access` sounds too robotic. How about something like`401: Unauthorized. You are not authorized to access this page.`.

Also, admin_user_index.html does not display the 401 message.

1maple1 (approved):

PR #1989, opened 2025-11-24
Better unresponsive garden handling

TheBurchLog:


1maple1 (approved):
Verified that warning triangle appears for child garden if problem with parent connection. Garden in unresponsive will sync at sync interval.

PR #1995, opened 2025-11-26
Add User Preference Support

TheBurchLog:
brewtils: user_preferences


1maple1 (approved):
Confirmed that preferences are saved in db

PR #1996, opened 2025-12-02
Remove Six imports

TheBurchLog:


1maple1 (approved):
Confirmed that all the six references in beer-garden were removed minus the requirements file for yapconf.

PR #2006, opened 2025-12-08
Upgrading Test Deps and imports

TheBurchLog:


test/integration/scheduler/conftest.py
```diff
@@ -1,9 +1,6 @@
 import pytest
 
-try:
-    from helper import RequestGenerator, setup_easy_client
-except (ImportError, ValueError):
-    from ..helper import RequestGenerator, setup_easy_client  # noqa
+from helper import setup_easy_client  # noqa
```
1maple1: `# noqa` not needed because there are no linter warnings

1maple1 (approved):

1maple1 (approved):

PR #2043, opened 2026-01-15
3.32 Check dead runners

1maple1:


src/app/beer_garden/systems.py
```diff
@@ -499,6 +499,21 @@ def remove_instance(
...
+            system = db.query_unique(System, instances__metadata__runner_id=runner.id)
+            if system:
+                db.modify(
+                    system,
+                    query={"instances__metadata__runner_id": runner.id},
+                    **{
+                        "set__instances__S__status": "ERROR",
+                    },
+                )
+
```
TheBurchLog: We should take the response object and publish a System Update event. That way the UI gets dynamically updated. Something like this:
```python
publish (
        Event(
            name=Events.SYSTEM_UPDATED.name,
            payload=system
        )
    )
    ```

File: src/app/beer_garden/systems.py [hunk 2]
```diff
@@ -499,6 +499,21 @@ def remove_instance(
     return instance
 
 
+def check_dead_runners() -> None:
+    runners = lpm.runners()
+    for runner in runners:
+        if runner.dead:
+            system = db.query_unique(System, instances__metadata__runner_id=runner.id)
+            if system:
+                db.modify(
```
TheBurchLog: Should do a quick spot check to make sure the system instance isn't already in an error status

src/app/beer_garden/scheduler.py
```diff
@@ -436,6 +437,14 @@ def internal_scheduled_jobs(self):
                 name="publish_garden",
             )
 
+        # Add scheduled job for checking dead runners
+        self.add_schedule(
+            beer_garden.systems.check_dead_runners,
+            interval=15,
```
TheBurchLog: 15 minutes feels like a long interval. Maybe we can do 5 minutes? So in that window the local UI will have the dead icon next to the status, then it will be converted over to Error status once this finally kicks off.

src/app/beer_garden/systems.py
```diff
@@ -499,6 +499,32 @@ def remove_instance(
     return instance
 
 
+def check_dead_runners() -> None:
+    runners = lpm.runners()
+    for runner in runners:
+        if runner.dead:
+            system = db.query_unique(
```
TheBurchLog, line 506: Do this instead so we only match on the fields on the instance and not across all of them:

```python
system = db.query_unique(
                System,
                instances__match={'metadata__runner_id': runner.id, 'status__ne': "ERROR"}
            )
```

TheBurchLog (approved):

PR #2052, opened 2026-02-09
508 Checks with linting

TheBurchLog:


src/react/src/components/SchedulerForm.tsx
```diff
@@ -100,7 +100,7 @@ function FileForm({
   return (
     <div>
       <div className="p-field">
-        <label className="ml-2">Path</label>
```
1maple1, line 103: Generally want labels for form elements so not sure why label was removed. Could use the `for` attribute though.

PR #2053, opened 2026-02-09
Scratchpad system

1maple1:


TheBurchLog (approved):

src/react/src/components/scratchPadComponents/SystemViewCard.tsx
```diff
@@ -0,0 +1,315 @@
...
+  });
+
+  system?.instances?.forEach((instance) => {
+    if (instance.status) {
+      statusCounts.set(
+        instance.status,
+        (statusCounts.get(instance.status) || 0) + 1,
+      );
+    }
+  });
```
TheBurchLog, line 110: This works because the HTML gets re-rendered, but since it is a displayed value should we make statusCounts a state object? Then we can update it in the useEffect for system updates?

src/react/src/components/scratchPadComponents/SystemViewCard.tsx
```diff
@@ -0,0 +1,315 @@
...
+              </Button>
+              <Button className="mr-2">
+                <FontAwesomeIcon icon="file-lines" />
+              </Button>
+            </div>
+          </div>
+        </div>
+      </div>
+    );
+  };
```
TheBurchLog, line 250: Since this is identical to the system cards page, we should move this over to a small component that both can share. We could probably do this with most of the shared code. The primary difference between the two is one button to push to scratch pad. Then this code would just handle events.

PR #2055, opened 2026-02-17
React: Garden index

TheBurchLog:


src/react/src/layouts/GardenIndex.tsx
```diff
@@ -1,155 +1,537 @@
...
+  const formatDate = (value: string) => {
+    if (value === null || value === undefined || value === "") {
+      return "UNKNOWN";
+    }
+    const date = new Date(value);
+    return date.toLocaleDateString() + " " + date.toLocaleTimeString();
+  };
+
+  const columns = useMemo(() => {
+    const cols = [
```
1maple1, line 456: What happened to the upstream field that used to be under metrics?

PR #2056, opened 2026-02-20
Combine Request Creates and Code examples

TheBurchLog:


src/react/src/layouts/RequestCreate.tsx
```diff
@@ -237,29 +133,141 @@ function RequestCreate() {
...
+          return printParams.join(", ");
+        }
+        return "";
+      };
+
+      return `
+from brewtils import SystemClient
+
+request = SystemClient(
+  system_name='${request?.system}',
```
1maple1: whitespace doesn't line up here

src/react/src/layouts/RequestCreate.tsx
```diff
@@ -237,29 +133,141 @@ function RequestCreate() {
...
+	system_namespace='${request?.namespace}',
+	version_constraint='${request?.system_version}',
+	default_instance='${request?.instance_name}',
+	bg_host='${getHostName()}',
+	bg_url_prefix='${getPrefix()}',
+	bg_port=${getPort()},
+	blocking=True,
+	ssl_enabled=${getSslEnabled()},
+	ca_cert=None,
+	ca_verify=None
```
1maple1: missing comma

src/react/src/layouts/RequestCreate.tsx
```diff
@@ -237,29 +133,141 @@ function RequestCreate() {
...
+	client_cert=None).command(${generateParams()})
```
1maple1: This isn't working. For something like system echo this should be `.say(params)` instead of `.command(params)`. Also what about when the user hasn't selected a command? Should this part be left off so a user can still create a valid system client? Or should the code examples be conditionally displayed?

src/react/src/layouts/RequestCreate.tsx
```diff
@@ -237,29 +133,141 @@ function RequestCreate() {
...
+  --header=Content-Type:application/json \\
+  ${getHostName()}:${getPort()}${getPrefix()}/api/v1/requests?blocking=true
+`;
+    };
+
+    const curlCode = () => {
+      return `
+curl -X POST ${getHostName()}:${getPort()}${getPrefix()}/api/v1/requests?blocking=true \\
+  -H "Content-Type: application/json" \\
+  -d '${JSON.stringify(request)}'
```
1maple1, line 183: This request is hanging on to previous values (notably command and params) if the request is changed. For example a child system remembering `"parameters":{"the_bytes":null}}'` after changing from deploy.

1maple1 (commented):
I'm not sure what I was supposed to test but I found the following. Ad-hoc request submission working. Pour again is working. It does not appear download output is working. Scratchpad components are working fine. There is an odd issue where the scratchpad status goes to CREATED after being SUCCESS and it happens when clicking the top menu and the whole page reloads. KeyError: 'trigger_type' with interval type job.

PR #2059, opened 2026-02-23
Support file uploading for parameters

TheBurchLog:


src/react/src/components/CommandForm.tsx
```diff
@@ -105,7 +107,10 @@ function CommandForm({
   };
   const [parametersFields, setParameterFields] = useState(buildDefaults());
 
+  const [resetForm, setResetForm] = useState(0);
+
   const resetRequest = () => {
+    setResetForm(resetForm + 1);
```
1maple1: This works, but I think it is safer to do`setResetForm(currentForm => { return currentForm + 1 });`. Calls to counters can get stale values otherwise.

1maple1 (commented):
Screen flickering advanced request parameter details json on first load after submit. Otherwise works much like the old UI. Still has the same large file size limitations.

PR #2070, opened 2026-03-05
React: Supporting Prefix Paths

TheBurchLog:


1maple1 (approved):
Works fine. Wonder if NavLink should be used over Link in Nav menu.

PR #2071, opened 2026-03-05
React: Fix recursive loop with System Versions

TheBurchLog:


src/react/src/components/CommandSelect.tsx
```diff
@@ -162,8 +172,8 @@ function CommandSelect({
 
     if (
       selectedVersion !== "latest" &&
-      selectedVersion !== null &&
-      !systemVersionList.includes(selectedVersion) &&
+      (selectedVersion === null ||
```
1maple1, line 173: I'm not convinced this will handle multiple versions well.

src/react/src/components/CommandSelect.tsx
1maple1, line 138: Could this be refactored to remove repeated code?

1maple1 (approved):
Fixes version loop

PR #2073, opened 2026-03-05
React: Add JSON example code for Create Request

TheBurchLog:


1maple1 (commented):
I like how JSON doesn't show initially until the command is selected, but it allows for copying the JSON code anyway as `undefined`. It seems to only update when the the command is selected. It would be a lot better if a conditional render was used so that if the request is not fully fleshed out the the code examples are hidden.

Random idea would be to put the code examples button down near reset form. Not in the bottom right, though, because I really want to click Submit all the time in the bottom right.

TheBurchLog:
Switched the Reset button to be a split button for the code examples:
image

1maple1 (approved):

PR #2074, opened 2026-03-06
React: Add Cancel and Download features

TheBurchLog:


1maple1 (approved):
This seems a little clunky but it works and we can revisit later.

PR #2075, opened 2026-03-09
React: Support large outputs and missing output type

TheBurchLog:


src/react/src/components/RequestOutput.tsx
```diff
@@ -1,8 +1,27 @@
+import { Button } from "primereact/button";
 import { Skeleton } from "primereact/skeleton";
+import { useState } from "react";
 
 import { Request } from "../models/brewtils-types";
 
-function displautOutput(request: Request) {
+function largeOutputCheck(request: Request): boolean {
```
1maple1, line 7: If largeOutputCheck is true then my expectation is that true means it has large output. Maybe call it `displayOutputCheck` or `showOutputCheck` instead? Otherwise maybe `!isLargeOutput`.

src/react/src/components/RequestOutput.tsx
```diff
@@ -1,8 +1,27 @@
...
+  if (request.output_type === "JSON") {
+    try {
+      if (Object.keys(JSON.parse(request.output || "{}")).length > 1000) {
```
1maple1: The key length is not the best indicator of output size. What if each key has numerous nested properties? Similar to blob you probably want to use bytes for consistency.

1maple1 (approved):

PR #2078, opened 2026-03-10
React: Add Garden Name to Navigation

TheBurchLog:


1maple1 (commented):
Works as intended. It will need css to ensure that it does not push navigational elements around on screen. 

Should this use garden_name or application_name?

1maple1 (approved):

PR #2081, opened 2026-03-12
React: user login

TheBurchLog:


src/react/src/components/UserLogin.tsx
```diff
@@ -0,0 +1,86 @@
...
+        onClick={() => setVisible(false)}
+        className="p-button-text"
+      />
+      <Button label="Login" icon="pi pi-check" onClick={handleLogin} />
+    </div>
+  );
+
+  return (
+    // <div className="card flex justify-content-center">
+    // <Button label="Show Login" icon="pi pi-external-link" onClick={() => setVisible(true)} />
```
1maple1: Still needed?

1maple1 (commented):
Ideally this should prompt for a login when auth is enabled. I had to click the login button. 

That mentioned, login worked well. Pages will eventually need to be hidden if user is not logged in when auth is enabled. Login button was not present when auth was disabled as expected.

1maple1 (approved):
Prompts on initial page load and login works fine

PR #2082, opened 2026-03-17
React: Has Access Permissions

TheBurchLog:


src/react/src/services/user_service.tsx
```diff
@@ -15,11 +19,16 @@ export const GetCurrentUser = (token: string) => {
...
-export const GetCurrentRoles = (token: string) => {
+export const GetCurrentRoles = (): Array<Role> | undefined => {
+  const token = GetToken();
   if (token !== null) {
     const decode = jwtDecode<CustomJwtPayload>(token);
     if (decode.roles) {
-      return decode.roles;
+      const userRoles = [] as Array<Role>;
+      for (const role of decode.roles) {
+        userRoles.push(JSON.parse(role) as Role);
```
1maple1: I'm getting an error on this line. I think this should be `userRoles.push(role as Role)`

1maple1 (commented):
Current page doesn't refresh after login.

PR #2087, opened 2026-03-19
Role index page

1maple1:


src/react/src/layouts/RoleIndex.tsx
```diff
@@ -0,0 +1,448 @@
...
+    }
+
+    return (
+      <div>
+        <Message
+          severity="error"
+          text="Warning - Beergarden authorization is currently disabled. Changes made here
+          will be persisted, but permissions will not be enforced. Contact your
+          administator to enable this feature."
+        />
```
TheBurchLog: Conditional statement based on `config?.auth_enabled === false`

src/react/src/layouts/RoleIndex.tsx
```diff
@@ -0,0 +1,448 @@
...
+          <label htmlFor="rolePermission">Permission</label>
+          <Dropdown
+            required
+            id="rolePermission"
+            options={permissions}
+            value={rolePermission}
+            optionLabel="label"
+            placeholder="Select One"
+            onChange={(e) => {
+              // e.originalEvent?.stopPropagation();
```
TheBurchLog: comment needed?

src/react/src/components/RoleScopeCard.tsx
```diff
@@ -0,0 +1,82 @@
...
+            key={index}
+            className="card flex flex-column gap-2"
+            header={() => header(index)}
+          >
+            <label htmlFor={`${scopeName}Scope-${index}`}>Scope</label>
+            <InputText
+              id={`${scopeName}Scope-${index}`}
+              value={item}
+              onChange={(e) => handleUpdateScope(e.target.value, index)}
+            />
```
TheBurchLog: Check out AutoComplete as an option where we can load in all of the unique options for the scope.

https://primereact.org/autocomplete/

1maple1:
Updated to address feedback. However, it still does not validate scope inputs like before. 

Edit: It should not validate inputs like before.

TheBurchLog:
The Add buttons are hard to tell which scope you are going to add. We should have the button say "Add System" or which ever scope it is adding.

TheBurchLog (approved):

PR #2088, opened 2026-03-19
React: Change Nav to Buttons

TheBurchLog:


1maple1 (commented):
Some of the styling is improved, but scrap the buttons. The buttons make the nav taller than before, the text is not centered in the button, and it has an odd shaped outline on hover.

PR #2091, opened 2026-03-26
Add system card component

1maple1:


TheBurchLog (approved):
Looks good, I think we could do some overall styling for it but it maps what we previously had to a component.

PR #2092, opened 2026-03-27
Scheduler Page Updates

TheBurchLog:


src/react/src/layouts/JobIndex.tsx
```diff
@@ -33,19 +71,119 @@ function JobIndex() {
     return job.next_run_time;
   };
 
-  const nameTemplate = (job: Job) => {
+  const editJob = (jobId: string) => {
+    window.open(`${GetBaseURL()}/job/${jobId}`, "_self");
```
1maple1: Use navigate / Link here?

1maple1 (commented):
Looks pretty good. One thing I noticed is that the Associated Requests moved down to the bottom in modal likely because of spacing.

1maple1 (approved):

PR #2093, opened 2026-03-30
React: System Card formatting

1maple1:


src/react/src/components/SystemCard.tsx
```diff
@@ -18,11 +18,17 @@ import { DeleteSystem, ReloadSystem } from "../services/system_service";
...
   system: System;
+  selectedGarden?: string;
   toast?: RefObject<Toast | null>;
   PushToPad?: any;
 }
 
-function SystemCard({ system, toast, PushToPad }: SystemCardProps) {
+function SystemCard({
+  system,
+  selectedGarden = "default",
```
TheBurchLog: I think we should let the default value for selectedGarden be undefined. Then we just put the requirement on what ever calls the System Card to pass that information properly, or it will always show namespace

TheBurchLog (approved):

PR #2095, opened 2026-03-31
Clear tokens if auth gets disabled

TheBurchLog:


1maple1 (approved):
Logged in with auth enabled then disabled auth and verified that tokens were cleared.

PR #2099, opened 2026-04-01
Extending default timeouts for NGINX

TheBurchLog:


1maple1 (approved):
Confirmed those are valid nginx directives and are defined as expected.

PR #2105, opened 2026-04-03
Import and Export Jobs Support

TheBurchLog:


src/react/src/services/job_service.tsx
```diff
@@ -208,3 +208,72 @@ export const DeleteJob = async (job: Job, headerData?: any) => {
...
+    );
+    document.body.appendChild(link);
+    link.click();
+    document.body.removeChild(link);
+    window.URL.revokeObjectURL(url);
+
+    return blob;
+  } catch (error) {
+    // Handle network errors or the error thrown above
+    console.error("Error fetching Requests:", error);
```
1maple1: Fix Requests it is misleading.

1maple1 (commented):
Import should prompt for a file on every click. After successful import the file dialog does not open and shows the success toaster. Maybe look into adding a ref and using an uploadHandler to clear the file after upload.

Should this allow export if there are no jobs?

1maple1 (commented):
There is still no way to import more than one jobs file.

1maple1 (approved):

PR #2107, opened 2026-04-07
React: Menubar navigation

1maple1:


src/react/src/Navigation.tsx
```diff
@@ -30,69 +29,82 @@ function NavigationMenu({
...
-      root: true,
-      template: itemRenderer,
-      route: "/jobs",
+      template: (item: any) => {
+        return (
+          <NavLink to="/scheduler" className="p-menuitem-link">
+            <span>{item.label}</span>
+          </NavLink>
+        );
+      },
```
TheBurchLog, line 51: Either change the router logic to point to /scheduler or update this to /job

PR #2113, opened 2026-04-10
Add Choice Validation Flag support for requests

TheBurchLog:


src/app/beer_garden/api/http/handlers/v1/request.py
```diff
@@ -562,6 +568,9 @@ async def post(self):
         ):
             request_model.requester = self.current_user.username
 
+        choice_validation_enabled = self.get_argument(
+            "choice_validation_enabled", default=None
```
1maple1, line 574: default=None and default above is true

1maple1 (approved):

PR #2114, opened 2026-04-15
User Index Page

TheBurchLog:


src/react/src/layouts/UserIndex.tsx
```diff
@@ -0,0 +1,372 @@
...
+                    console.error("Error attempting to delete user", error),
+                  );
+              }
+            }}
+            tooltip={`Delete User ${rowData.username}`}
+            tooltipOptions={{ position: "bottom" }}
+            data-testid={`delete-user-${rowData.id}`}
+          >
+            <FontAwesomeIcon icon="trash" />
+          </Button>
```
1maple1, line 461: This needs to show a confirm dialog before actually deleting and a toast when the user has been removed.

src/react/src/layouts/UserIndex.tsx
```diff
@@ -0,0 +1,372 @@
...
+
+  useEffect(() => {
+    if (
+      !showAccountMappingDialog &&
+      !showRolesDialog &&
+      !showCreateUserDialog
+    ) {
+      loadUsers();
+    }
+  }, [showAccountMappingDialog, showRolesDialog, showCreateUserDialog]);
```
1maple1: Opening one of these dialogs and hitting Close triggers a rerender and it shouldn't as there was nothing changed. Create user can refresh users on submit or just append the new user to the users in state. The others could just modify the users in state if there is something in the table that needs to change.

src/react/src/services/token_service.tsx
```diff
@@ -156,3 +156,17 @@ export const GetAuthHeaders = () => {
   }
   return headers;
 };
+
+export const RevokeToken = async (username: string) => {
+  const headers = GetAuthHeaders();
+
+  const response = await fetch(`${GetBaseURL()}/api/v1/token/'${username}`, {
```
1maple1: Stray `'` before username. Seeing requests like `/'user3`

src/react/src/layouts/UserIndex.tsx
```diff
@@ -0,0 +1,395 @@
...
+          <Button
+            onClick={() => {
+              const accept = () => {
+                if (rowData.username) {
+                  DeleteUser(rowData.username)
+                    .then(() => {
+                      toast.current?.show({
+                        severity: "info",
+                        summary: "Delete User",
+                        detail: `Delete ${rowData.username} complete`,
```
1maple1: How about `Successfully deleted ${username}`?

src/react/src/components/UserChangePassword.tsx
```diff
@@ -0,0 +1,198 @@
...
+              onChange={(e: ChangeEvent<HTMLInputElement>) =>
+                setCurrentPassword(e.target.value)
+              }
+            />
+          </>
+        )}
+        <label htmlFor="newPassword" className="font-bold">
+          New Password
+        </label>
+        <Password
```
1maple1, line 168: Tab highlights the toggle mask. Play around with tabIndex so it goes to confirm input.

1maple1 (commented):
Should all buttons display a toast confirmation? For example updating user roles has no confirmation.

1maple1 (approved):

PR #2116, opened 2026-04-16
React: Error handling example

1maple1:


src/react/src/types/errors.ts
```diff
@@ -0,0 +1,10 @@
+class HttpError extends Error {
+  code: number;
+  constructor(message: string, code: number) {
+    super(message);
+    Object.setPrototypeOf(this, HttpError.prototype);
+    this.code = code;
+  }
+}
+
+export default HttpError;
```
TheBurchLog: Going away?

src/react/src/components/ErrorPage.tsx
```diff
@@ -0,0 +1,46 @@
...
+        return "Error";
+    }
+  }
+
+  return (
+    <div>
+      {!isCard && <h1 className="flex m-2">{errorType(errorCode)}</h1>}
+      <div className="flex">
+        <Panel
+          header={isCard ? errorType(errorCode) : "Details"}
```
TheBurchLog, line 32: Default could be:
```javascript
`(${errorCode}) ${errorType(errorCode)}`
```

src/react/src/components/ErrorPage.tsx
```diff
@@ -0,0 +1,46 @@
...
+    switch (errorCode) {
+      case 400:
+        return "Bad Request";
+      case 401:
+        return "Unauthorized";
+      case 404:
+        return "Not Found";
+      default:
+        return "Error";
+    }
```
TheBurchLog, line 24: Or do something like this:
```javascript
switch (errorCode) {
      case 400:
        return "Bad Request";
      case 401:
        return "Unauthorized";
      case 404:
        return `Not Found ${errorCode}`;
      case undefined:
        return "Unknown Error";
      default:
        return "Error";
    }
```

TheBurchLog (approved):

PR #2118, opened 2026-04-20
Create Request Nav Button

TheBurchLog:


src/react/src/components/Navigation.tsx
```diff
@@ -213,6 +263,31 @@ function NavigationMenu({
...
+      {showCreateRequest && requestItem && (
+        <Dialog
+          visible={showCreateRequest}
+          style={{ width: "70%", overflowY: "auto" }}
+          modal
+          onHide={() => {
+            setShowCreateRequest(false);
+            setRequestItem(undefined);
+          }}
+          content={() => (
```
1maple1: I understand that this is running in headless mode so the card can be used for the create request modal and the workspace. However, it just looks bad and breaks from the look and feel of other dialogs/modals on the site. To use it both places requires a redesign. The modal should control the buttons and the card should contain only content.

src/react/src/components/RequestViewCard.tsx
```diff
@@ -298,25 +298,38 @@ function RequestViewCard({
...
-              },
-            ]}
+            model={
+              request &&
+              request.status &&
+              !["CREATED", "IN_PROGRESS"].includes(request.status)
+                ? [
+                    {
+                      label: "Run Again Now",
+                      // icon: <FontAwesomeIcon icon="arrow-up-right-from-square" />,
```
1maple1: Icon not used?

src/react/src/components/Navigation.tsx
```diff
@@ -276,13 +267,10 @@ function NavigationMenu({
...
 
     AddTourStep(tourStepsRef, homeLinkTourStep);
+    AddTourStep(tourStepsRef, createRequestTourStep);
     AddTourStep(tourStepsRef, requestTourStep);
     AddTourStep(tourStepsRef, schedulerTourStep);
     AddTourStep(tourStepsRef, workspaceTourStep);
-    AddTourStep(tourStepsRef, topicsTourStep);
-    AddTourStep(tourStepsRef, usersTourStep);
-    AddTourStep(tourStepsRef, rolesTourStep);
-    AddTourStep(tourStepsRef, aboutTourStep);
```
1maple1, line 285: Removed tour for these 4 but did not add the admin menu to mention these at all.

PR #2125, opened 2026-04-24
Change Theme Support

TheBurchLog:


1maple1 (commented):
Looks good for the most part. See overlay changes on top of this because change dialog may not be needed.

PR #2128, opened 2026-04-27
React: User overlay panel

1maple1:


TheBurchLog (approved):
Remove the old unit tests for navigation checks, then it is good to go

PR #2136, opened 2026-04-30
Backend changes for React UI

TheBurchLog:
Staging a few changes needed for demoing new React UI features.

### DIFF
--- a/src/app/beer_garden/scheduler.py
+++ b/src/app/beer_garden/scheduler.py
@@ -543,7 +543,7 @@ def run_job(job_id, request_template, **kwargs):
     """
     import beer_garden.router
 
-    request_template.metadata["_bg_job_id"] = job_id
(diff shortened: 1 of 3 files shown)


src/app/beer_garden/scheduler.py
```diff
@@ -543,7 +543,7 @@ def run_job(job_id, request_template, **kwargs):
     """
     import beer_garden.router
 
-    request_template.metadata["_bg_job_id"] = job_id
+    request_template.metadata["bg_job_id"] = job_id
```
1maple1, line 546: Do you need to update this reference in SchedulerViewCard? `metadata__bg_job_id`

TheBurchLog: This update was so that the reference card can query on it. The mongoengine API doesn't like sub fields that start with an underscore.

1maple1 (approved):

PR #2140, opened 2026-05-04
React rpm installer

TheBurchLog:
Updates:
- Local installer to import brewtils if doesn't exist
- Adds React preview builder
- Adds NGINX routing logic for /preview and redirects API calls through existing configurations
- Updated default OS to Rocky 9

Run bash script for testing:
```bash
# Update version if needed
make rpm-build-local VERSION=3.33.1
```


1maple1 (approved):
Built rpm and content have preview react directory

PR #2145, opened 2026-05-05
Support React UI in Docker

TheBurchLog:
Testing:

Build Unstable
```bash
make docker-build-unstable
```

Update Docker compose to target `image: bgio/ui_react:unstable`, then run it.


1maple1 (commented):
Working good, but the top-level Makefile needs to be updated to build the ui_react image, too. It only builds when in the src/react directory. 

ui-react and ui_react is inconsistent. Docker-compose has beer-garden service and the beer-garden image for an example.

Makefile
```diff
@@ -4,6 +4,7 @@ PYTHON         = python
 MODULE_NAME    = beer_garden
 APP_DIR        = src/app
 UI_DIR         = src/ui
+REACT_DIR         = src/react
```
1maple1: It would be nice if the spacing lined up

1maple1 (approved):

PR #2156, opened 2026-05-11
Skip Permission Check if disabled

TheBurchLog:


src/react/src/components/SystemCard.tsx
```diff
@@ -438,6 +438,39 @@ function SystemCard({
       });
     }
 
+    const disabledStartStopButtons = (
```
1maple1, line 441: This is disabled and the one below is disable.

1maple1 (commented):
Some duplicated code. Why not just disable the buttons by updating a state variable for the affected buttons?

src/react/src/components/SystemCard.tsx
```diff
@@ -438,6 +438,39 @@ function SystemCard({
...
+          severity="warning"
+          size="small"
+          disabled={true}
+          {...GenerateTourProps(stopInstanceTourStep)}
+        >
+          <FontAwesomeIcon icon="stop" />
+        </Button>
+      </>
+    );
+    const disableAdminMenu = (
```
1maple1, line 461: Make this disabledAdminMenu

1maple1 (approved):
HasAccess is working a lot better now.

PR #2159, opened 2026-05-12
Add view topic card button

TheBurchLog:


src/react/src/components/TopicCard.tsx
```diff
@@ -0,0 +1,342 @@
...
+    }
+  }, [listeners]);
+
+  useEffect(() => {
+    if (topic === undefined) {
+      setTopic(targetTopic);
+    } else if (requests === undefined) {
+      queryTopicRequests();
+    }
+  }, [topic]);
```
1maple1: Pagination is broken for topic requests. The lazyParams is not listed as a dependency. Also would have to remove `requests === undefined`.

src/react/src/layouts/TopicIndex.tsx
```diff
@@ -698,6 +712,28 @@ function TopicIndex({ config }: { config: Config }) {
...
+        visible={viewTopic !== undefined}
+        footer={
+          <>
+            <Button onClick={() => setViewTopic(undefined)}>Close</Button>
+          </>
+        }
+        onHide={() => {
+          setViewTopic(undefined);
+        }}
+      >
```
1maple1: These leave an empty dialog behind briefly after hitting close button. Better to hide the dialog right away and then reset whatever values.

src/react/src/layouts/TopicIndex.tsx
```diff
@@ -89,7 +89,7 @@ function TopicIndex({
       command: "",
     } as Subscriber,
   ]);
-  const [viewTopic, setViewTopic] = useState<Topic | undefined>(undefined);
+  // const [viewTopic, setViewTopic] = useState<Topic | undefined>(undefined);
```
1maple1: Not needed if not controlling dialog

1maple1 (approved):

PR #2165, opened 2026-05-14
Enable stepper header nav like back button

1maple1:
Allow back button type functionality on request wizard using the stepper header nav

### DIFF
--- a/src/react/src/components/RequestWizard.tsx
+++ b/src/react/src/components/RequestWizard.tsx
@@ -471,7 +471,7 @@ function RequestWizard({
   ];
 
   const cleanForm = () => {
-    const newRequest = { ...request, command_type: "ACTION" };
+    const newRequest = { ...request, command_type: "" };
     delete newRequest.parameters;
     delete newRequest.comment;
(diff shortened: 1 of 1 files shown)


src/react/src/components/RequestWizard.tsx
```diff
@@ -463,6 +470,31 @@ function RequestWizard({
...
 
+  const handleChangeStep = (e: StepperChangeEvent) => {
+    setActiveIndex(e.index);
+    if (e.index == 1) {
+      // Enable panel 1 & disable 2
+      setStepperPanel1Options({});
+      setStepperPanel2Options(stepperPanelOptions);
+      if (request?.parameters) {
+        const newRequest = { ...request };
+        delete newRequest.parameters;
```
TheBurchLog: Should also delete `newRequest.comment` and `newRequest.command_type` so we get a clean form.

1maple1: Great idea. Put these in reusable cleanForm function.

TheBurchLog (approved):

PR #2168, opened 2026-05-14
Sticky current requests

TheBurchLog:


src/react/src/components/CurrentRequestsTemplate.tsx
```diff
@@ -26,6 +26,7 @@ function CurrentRequestsTemplate({
     altRequests.current = requests.map((req) => ({
       id: req.id,
       status: req.status,
```
1maple1: This needs `command: req.command_display_name ?? req.command,` to display the command name correctly each time.

1maple1 (approved):

PR #2172, opened 2026-05-19
Add instance filtering for Garden Dashboard

TheBurchLog:


TheBurchLog:
doesn't update instances with events if filtered

1maple1 (commented):
The dashboard is not being updated in real time. It might have something to do with switching between filtered and selected systems. Additionally, check if adding filter to multiselect looks ok.

1maple1 (approved):

PR #2173, opened 2026-05-19
Update how widths are set for dashboard

TheBurchLog:


src/react/src/layouts/Dashboard.tsx
```diff
@@ -504,69 +504,74 @@ function GardenDashboard({
...
+              if (typeof e.value === "string") {
+                findSelectedGarden(e.value);
+              }
+            }}
+            togglerTemplate={<></>}
+          />
+        </div>
+
+        {/* MAIN WORKSPACE */}
+        <div style={{ width: "82vw", minWidth: "250px" }}>
```
1maple1: One of the things that is odd at 82vw is that the buttons don't line up closely on right with the menu bar right edge. 83vw is a bit closer to ideal. 84vw doesn't leave room for the vertical scroll bar and causes overflow so an ugly horizontal scrollbar appears.

Long story short I think percentages are better here. "16% and 84%" will work and use the full width with no overflow but I think "17% and 83%" use the space a little better. You decide.

1maple1 (approved):

PR #2176, opened 2026-05-19
Reload form values for pour again and edit job

1maple1:


src/react/src/components/RequestCreateCard.tsx
```diff
@@ -202,6 +203,27 @@ function RequestCreateCard({
...
+            ...prevReq,
+            namespace: responseJob?.request_template?.namespace,
+            system: responseJob?.request_template?.system,
+            system_version: responseJob?.request_template?.system_version,
+            instance_name: responseJob?.request_template?.instance_name,
+            command: responseJob?.request_template?.command,
+            parameters: responseJob?.request_template?.parameters,
+            command_type: responseJob?.request_template?.command_type,
+            comment: responseJob?.request_template?.comment,
+          }));
```
TheBurchLog: I am not sure why, but I keep having issues in other places with this style of setting the stateful object. Instead I have been doing:
```javascript
setRequest({
    ...request,
    namespace: "value"
});
```

I know `prevReq` is suppose to work fine, but when multiple things are updating the value, it can sometimes miss stuff.

TheBurchLog (approved):

PR #2180, opened 2026-05-21
Menu Update and 508 changes

TheBurchLog:


1maple1 (commented):
Menubar blinks icons into and out of existence on first load and hard refresh. This is only when auth is not enabled.

1maple1 (approved):

PR #2184, opened 2026-05-27
508 Dashboard

TheBurchLog:


src/react/index.html
```diff
@@ -1,5 +1,6 @@
 <!DOCTYPE html>
 <html lang="en">
+  <a href="#main-content" class="skip-link">Skip to main content</a>
```
1maple1: I understand the point of this but it is a little obtrusive seeing it on load sometimes. It needs to be visually hidden and it is once the navigation loads just not sure why it shows sometimes.

src/react/src/layouts/Dashboard.tsx
```diff
@@ -624,6 +624,17 @@ function GardenDashboard({
             placeholder="Filter by Instance Status"
             className="mb-3"
             filter
+            pt={{
+              input: {
+                "aria-label": "Filter by Instance Status",
+                "aria-controls": "instance_status_multiselect_input",
+                "autocomplete": "off",
```
1maple1: Should this be `autoComplete`? Console is showing errors.

1maple1 (approved):

PR #2191, opened 2026-05-27
508 Scheduled Index and Consolidate Functions

TheBurchLog:


1maple1 (approved):
Recommend adding a tabIndex={-1} to Link on RequestIndex so it doesn't get a tab before the open request button.

PR #2195, opened 2026-05-29
Random type problem fixes

1maple1:


TheBurchLog (approved):
That gets rid of the error messages

PR #2196, opened 2026-05-29
Fix double request generation

TheBurchLog:


1maple1 (commented):
Capturing both mousedown and click still causes double request submission. One for the mousedown and second for the click event. I  think this should just be `click`. So a user would need a mousedown and mouseup (i.e. click) on the same element to fire.

1maple1 (approved):

PR #2198, opened 2026-05-29
508 Current Requests and Links

TheBurchLog:


1maple1 (commented):
CurrentRequests needs to get the focus much like the fix for user overlay when expanded.

1maple1 (approved):

PR #2200, opened 2026-06-02
Consolidate confirm dialog and add confirm to job and request delete

1maple1:


src/react/src/layouts/Workspace.tsx
```diff
@@ -196,6 +197,7 @@ function Workspace({
   return (
     <div>
       <h1>Workspace</h1>
+      <ConfirmDialog />
```
TheBurchLog: Was this one suppose to be removed?

src/react/src/components/RequestItemCard.tsx
```diff
@@ -78,7 +82,12 @@ function RequestItemCard({
...
                 })
                 .catch((error) => {
-                  console.error("Error deleting job:", error);
+                  toast.current?.show({
+                    severity: "error",
+                    summary: "Error",
+                    detail: `Error deleting job: ${error}`,
+                    life: 3000,
+                  });
                 });
```
TheBurchLog: This code is fine, just wanted to confirm it was intended to include it within this PR.

src/react/src/layouts/Workspace.tsx
1maple1: It should not have been added. Removed it.

src/react/src/components/RequestItemCard.tsx
1maple1: Right, this is better for a toast follow-up PR so removed it for now.

TheBurchLog (approved):

PR #2204, opened 2026-06-04
Consolidate toast using toast provider

1maple1:


src/react/src/providers/ToastProvider.tsx
```diff
@@ -0,0 +1,41 @@
...
+};
+
+// Create a custom hook for consuming the context
+export const useToast = () => {
+  const context = useContext(ToastContext);
+  if (!context) {
+    throw new Error("useToast must be used within a ToastProvider");
+  }
+  return context.showToast;
+};
```
TheBurchLog, line 41: Any issue with changing this to be a pass through straight to `showToast` instead of reference? Then we only need to import the function and not have to declare it in the code:

```javascript
export const useToast = (args: ToastArgs) => {
  const context = useContext(ToastContext);
  if (!context) {
    throw new Error("useToast must be used within a ToastProvider");
  }
  return context.showToast(args);
};
```

TheBurchLog (approved):

PR #2207, opened 2026-06-04
Sort Instance dropdown

1maple1:


TheBurchLog:
All dropdowns should be sorted, not just instances

TheBurchLog (approved):

PR #2212, opened 2026-06-05
Add request deleted events

1maple1:


src/app/beer_garden/requests.py
```diff
@@ -1286,10 +1303,10 @@ def handle_event_create(event):
...
 
 def handle_event(event):
-    # TODO: Add support for Request Delete event type
-    # if event.name == Events.REQUEST_DELETED.name
-    #   and event.garden != config.get("garden.name"):
-    #     delete_requests(**event.payload)
+    if event.name == Events.REQUEST_DELETED.name and event.garden != config.get(
+        "garden.name"
+    ):
+        delete_request(**event.payload)
```
TheBurchLog: This shouldn't delete the request if a downstream cancelled it. But it could cancel the request if it is not in a completed status, and the `request.target_garden` is the same garden as the `event.garden`

TheBurchLog (approved):

PR #2213, opened 2026-06-08
Delete request should close dialog

1maple1:


src/react/src/components/RequestOptions.tsx
```diff
@@ -117,13 +119,13 @@ function RequestOptions({
...
-                })
-                .catch((error) => {
-                  console.error("Error deleting request:", error);
-                });
+              if (isCard) {
+                if (deleteRequest) {
+                  deleteRequest();
+                }
+              } else {
+                void navigate(`/requests`);
```
TheBurchLog: This doesn't delete the request when you are on the Request View page. It works only with the card version.

src/react/src/components/RequestOptions.tsx
```diff
@@ -119,7 +121,13 @@ function RequestOptions({
             const accept = () => {
               DeleteRequest(request)
                 .then(() => {
-                  void navigate(`/requests`);
+                  if (isCard) {
+                    if (closeRequest) {
+                      closeRequest();
+                    }
+                  } else {
```
TheBurchLog: Only the `if (closeRequest){` should be needed. Because only cards should be provided it, so we shouldn't create an accidental trap for ourselves if a card is used without the close request function being passed.

TheBurchLog (approved):

PR #2215, opened 2026-06-08
508 user dialogs

TheBurchLog:


src/react/src/passthrough/TreeTablePT.tsx
```diff
@@ -0,0 +1,4 @@
+export const TreeTablePT = {
+  root: { role: undefined },
+  row: { tabindex: -1, "aria-checked": undefined },
```
1maple1, line 3: This is a PrimeReact component so I think this should be "tabIndex"

TheBurchLog: There is a sub component within the row that does have `tabindex:0` that controls the toggle feature of the row.

PR #2216, opened 2026-06-09
More toast messages

1maple1:


src/react/src/components/RequestTreeChart.tsx
```diff
@@ -161,6 +161,12 @@ function RequestTreeChart(props: RequestTreeChartProps) {
             onClick={() =>
               DeleteRequest(node.data).catch((error) => {
                 console.error("Error deleting request:", error);
+                showToast({
+                  severity: "error",
+                  summary: "Error",
+                  detail: `Error deleting request: ${error}`,
+                  life: 3000,
+                });
```
TheBurchLog: Does this one need:
```js
 const showToast = useToast();
```

1maple1: Done

TheBurchLog (approved):

PR #2219, opened 2026-06-09
Handle Root Request Update Events

TheBurchLog:
Run Sleep_Say command, previously the tree would update and remove children requests. Now it adds them properly with their update events.


1maple1 (commented):
Not sure this is working correctly

TheBurchLog:
Re Verify this after split view is merged in

TheBurchLog:
Broken still in latest code

1maple1 (approved):

PR #2221, opened 2026-06-09
Pour Again dialog with Request Tree

TheBurchLog:


1maple1 (approved):
Like the dialog instead of previous redirect to workspace.

PR #2229, opened 2026-06-16
Clear Session Data with User Change

TheBurchLog:


1maple1 (commented):
Session data cleared fine. The question is should the themes be cleared on logout and if so then should the power user setting be reset, too?

1maple1 (approved):

PR #2231, opened 2026-06-23
Request view split

TheBurchLog:


src/react/src/layouts/RequestView.tsx
```diff
@@ -133,33 +87,70 @@ function RequestView({
...
-            })
-            .catch((error) => {
-              showToast({
-                severity: "error",
-                summary: "Error",
-                detail: `Error fetching parent request: ${error}`,
-                life: 3000,
-              });
-            });
+          const root_request = await GetRequest(check_request.parent.id, {});
```
1maple1: Better to use camelCase

1maple1 (commented):
The "<1s" probably should have a tooltip or something because I would not have known the delta is from created to last status update. Active tab is very hard to read in dark mode. I'm still not convinced the left side tree is the way to go because it creates a lot of whitespace and some of the information is duplicated.

1maple1 (approved):

PR #2239, opened 2026-06-26
Access Button Updates

TheBurchLog:


1maple1 (approved):
Not all these primereact props / styles will translate but code is fine.

PR #2245, opened 2026-06-26
Clear browser, fix reload UI loop, global read only requirement

TheBurchLog:


src/react/src/components/UserOverlay.tsx
```diff
@@ -170,6 +172,20 @@ function UserOverlay({
...
+        <Divider />
+        <AccessButton
+          size="small"
+          severity="warning"
+          className="mr-2"
+          onClick={onClearSession}
+          data-testid="clear-session-overlay"
+        >
+          <FontAwesomeIcon className="mr-2" icon="eraser" />
+          <span>Clear Browser Data</span>
```
1maple1: Maybe make this "Clear Session Data"? Also if logged in then this logs you out. Is that the intent?

1maple1 (approved):

PR #2253, opened 2026-07-08
Subscriber narrow options

1maple1:


src/react/src/components/SubscriberItem.tsx
```diff
@@ -43,9 +50,20 @@ function SubscriberItem({
...
 
+  const [selectedGardenName, setSelectedGardenName] = useState<
+    string | undefined
+  >(undefined);
+  const [selectedNamespaceName, setSelectedNamespaceName] = useState<
+    string | undefined
+  >(undefined);
+  const [selectedSystemName, setSelectedSystemName] = useState<
+    string | undefined
+  >(undefined);
```
TheBurchLog: These are not being used to trigger a state change on the UI or Effect, I would change them to Ref objects.

TheBurchLog (approved):

PR #2258, opened 2026-07-09
Mui dashboard buttons

1maple1:


src/react/src/components/AccessButton.tsx
```diff
@@ -24,44 +26,44 @@ const AccessButton = ({
 >) => {
   if (!Object.hasOwn(props, "tooltip")) {
```
TheBurchLog: This If statement shouldn't check props for tooltip since it is now passed through the HasAccessProps interface.

TheBurchLog (approved):

PR #2263, opened 2026-07-15
MUI Navigation

1maple1:


src/react/src/components/Navigation.tsx
```diff
@@ -409,72 +580,147 @@ function NavigationMenu({
...
+              color: "white",
+              backgroundColor: "primary.main",
+              textTransform: "none",
+              "&:hover": {
+                backgroundColor: "primary.dark",
+                opacity: [0.9, 0.8, 0.7],
+              },
+            }}
+          >
+            <MenuIcon />
```
TheBurchLog: If we stick with FontAwesome, then we don't need the second import for icons
```html
<FontAwesome icon="bars"/>
```

1maple1: I took out the mui icons so see if you like it

src/react/src/services/util_service.tsx
```diff
@@ -326,3 +329,19 @@ export const PaginatorTemplate = {
     );
   },
 };
+
+interface faProps {
```
TheBurchLog: We could use the Font Awesome Props instead:
```js
import { FontAwesomeIcon, FontAwesomeIconProps } from "@fortawesome/react-fontawesome";
```

src/react/package.json
```diff
@@ -10,6 +10,7 @@
     "@fortawesome/free-regular-svg-icons": "^6.7.2",
     "@fortawesome/free-solid-svg-icons": "^6.7.2",
     "@fortawesome/react-fontawesome": "^0.2.2",
+    "@mui/icons-material": "^9.2.0",
```
TheBurchLog: Need to do `npm uninstall @mui/icons-material` on this package to remove it

TheBurchLog (approved):

PR #2264, opened 2026-07-15
MUI Request Item Dialog

TheBurchLog:


src/react/src/App.tsx
```diff
@@ -560,18 +560,29 @@ function App() {
                         <Grid>
                           {fullScreenDialog === false && (
                             <AccessButton
+                              className="mr-2"
```
1maple1: Heads up - this will not work when primereact is removed. You will need to use `sx={{mr: 2}}`. AccessButton uses MUI ButtonProps so you should just be able to make the change now.

1maple1 (approved):

PR #2266, opened 2026-07-16
MUI Command Select

TheBurchLog:


src/react/src/components/CommandSelect.tsx
```diff
@@ -305,115 +307,194 @@ function CommandSelect({
...
+            </MenuItem>
           ))}
-        </datalist>
-        <Dropdown
+        </TextField>
+      )}
+
+      {systemNames.length === 0 && (
+        <TextField
+          sx={{ width: "100%", mr: 2, mt: 2, mb: 2 }}
```
1maple1, line 357: You can also use my: 2 instead of mt:2 and mb: 2

1maple1 (approved):

PR #2267, opened 2026-07-20
Mui scheduler form

TheBurchLog:


src/react/src/components/SchedulerForm.tsx
```diff
@@ -400,169 +469,157 @@ function IntervalForm({
...
+          >
+            Interval Number
+          </FormLabel>
+        </Box>
+        <Box sx={{ width: layoutProps.valueWidth }}>
+          <NumberField
             id="intervalNumber"
             value={intervalNumber}
-            onChange={(e) => setIntervalNumber(e.value)}
+            onValueChange={(value: number | null) => setIntervalNumber(value)}
```
1maple1, line 490: Should this be null or undefined?

1maple1 (commented):
Update NumberField to use carats

src/react/src/components/SchedulerForm.tsx
TheBurchLog: Null

TheBurchLog: https://base-ui.com/react/components/number-field#root

1maple1 (approved):

PR #2269, opened 2026-07-21
Fix Dialog Formatting

TheBurchLog:


1maple1 (commented):
Formatting looks good. Command type should have a default. There are also a number of console errors.

PR #2274, opened 2026-07-21
Fix Request Wizard buttons

1maple1:


src/react/src/components/RequestWizard.tsx
```diff
@@ -733,16 +734,20 @@ function RequestWizard({
...
-                severity="secondary"
+                color="secondary"
                 onClick={() => {
                   setSelectedInstance(undefined);
                   stepperRef.current?.prevCallback();
                 }}
-              />
-            </div>
+              >
+                Back
```
TheBurchLog, line 748: For buttons, what are your thoughts on always doing the button variant of Typography?
```html
<Typography variant="button" sx={{ display: "block" }}>
    Back
</Typography>
```

TheBurchLog (approved):

PR #2275, opened 2026-07-22
Code Examples Conversion

TheBurchLog:


1maple1 (commented):
I think it should have the divider and the X button in the top right like the parent dialog

1maple1 (approved):

PR #2280, opened 2026-07-22
Migrate Boolean Inputs

TheBurchLog:


1maple1 (commented):
Fix styling. Checkbox is pushed all the way to the right.

PR #2281, opened 2026-07-22
Migrate Choices Input

TheBurchLog:
Use `say_specific` command from dynamic to test


src/react/src/components/CommandFormField.tsx
```diff
@@ -144,117 +150,117 @@ function CommandFormField({
...
-        </datalist>
-        <Dropdown
-          id={parameter.key}
-          value={parameter.value}
-          options={parameter.options}
-          invalid={
-            (!disabled &&
+      <Box
+        key={parameter.key}
+        sx={{ display: "flex", justifyContent: "flex-end", m: 2 }}
```
1maple1, line 215: Flex-end is causing the dropdown to be pushed to the right which is not what you want

PR #2283, opened 2026-07-22
Migrate autocomplete field

TheBurchLog:
Test with dynamic say_specific_non_strict_typeahead


src/react/src/components/CommandFormField.tsx
```diff
@@ -257,37 +254,43 @@ function CommandFormField({
...
-
-        setItems(filteredItems?.map((option) => option.value));
-      }
-    };
     return (
-      <div key={parameter.key} className="p-field">
-        <AutoComplete
+      <Box
+        key={parameter.key}
+        sx={{ display: "flex", justifyContent: "flex-end", m: 2 }}
```
1maple1, line 260: This should probably be mr: 2 so it lines up with the other inputs.

1maple1 (approved):

PR #2287, opened 2026-07-23
Change instance dropdown to select

1maple1:


src/react/src/components/CommandList.tsx
```diff
@@ -84,25 +93,33 @@ function CommandList({
...
-        aria-label="Select Instance"
-      />
+      <FormControl fullWidth>
+        <InputLabel id="instance-label-select">Instance</InputLabel>
+        <Select
+          labelId="instance-label-select"
+          id="instance-select"
+          value={instance}
+          label="instance"
+          disabled={sortedInstances.length == 1}
```
TheBurchLog, line 116: When there is only one option for instances, we should make sure the instance value is displayed properly with the single instance name, instead of just showing the label and an disabled selection option

TheBurchLog (approved):

PR #2290, opened 2026-07-27
Update Request wizard

1maple1:


src/react/src/components/RequestWizard.tsx
```diff
@@ -575,356 +572,216 @@ function RequestWizard({
...
-      className="justify-content-center"
-      unstyled={isDialog}
-      header={
-        !isDialog && (
-          <Box sx={{ display: "flex" }}>
+    if (activeIndex == 1) {
+      return (
+        <>
+          <Box
+            sx={{ p: 2, border: "1px solid #e5e7eb", borderRadius: 2, mb: 2 }}
```
TheBurchLog: Instead of defining by color directly, we could use the MUI provided approach for values so it is easier to update down the road if we need to update something for contrast.

```javascript
import { grey } from '@mui/material/colors';
```
```html
<Box sx={{ p: 2, border: "1px solid", borderColor: grey[300], borderRadius: 2, mb: 2 }}>
```

TheBurchLog (approved):

PR #2295, opened 2026-07-28
Fix update request

1maple1:


TheBurchLog (approved):
Fixes the issue with Updating Job with Wizard.

PR #2297, opened 2026-07-28
Migrate Request View

TheBurchLog:


1maple1 (approved):
DialogContent in App would do wonders here... but this file looks good.

PR #2303, opened 2026-08-03
Convert Roles

1maple1:


src/react/src/layouts/RoleIndex.tsx
```diff
@@ -131,58 +126,88 @@ function RoleIndex({
...
+          gap: 1,
+        }}
+      >
+        <Typography variant="h2" component="h1">
+          Role Management
+        </Typography>
+
+        <Box sx={{ display: "flex" }}>
           <AccessButton
+            sx={buttonStyle}
```
TheBurchLog, line 175: Should we try this as a button group to save space?

src/react/src/layouts/RoleIndex.tsx
```diff
@@ -291,65 +308,100 @@ function RoleIndex({
...
+              filterable: true,
+              isString: true,
+            },
+            {
+              id: "scope_gardens",
+              field: "scope_gardens",
+              label: "Garden Scope",
+              sortable: true,
+              filterable: true,
+              isString: true,
```
TheBurchLog: String Arrays are not currently supported in the table filtering. In the current code base, a custom template is required and to disable filterable and sortable.

src/react/src/components/RoleScopeCard.tsx
```diff
@@ -113,66 +121,63 @@ function RoleScopeCard({
...
+                      index,
+                    );
+                  }}
+                  disabled={disabled}
+                  renderInput={(params) => (
+                    <TextField {...params} label="Scope" />
+                  )}
+                />
+              </CardContent>
+            </Card>
```
TheBurchLog, line 167: This creates a lot of white space due to the extra padding from the Card sections. We could get away with just a box with the border outlined, and minimum padding.

TheBurchLog:
Example to pull array of values based on path notation:

```javascript
  const findValue = (path: string, obj: any): any => {
    const keys = path
      .replace(/\[(\d+)\]/g, ".$1") // convert [0] to .0
      .split(".")
      .filter(Boolean);
    const resolve = (current: any, keyIndex: number): any => {
      if (current == null || keyIndex >= keys.length) {
        return current;
      }
      const key = keys[keyIndex];
      // 1. If the current object is an array AND the key is NOT a numeric index,
      // we branch and search all elements for that key.
      if (Array.isArray(current) && !/^\d+$/.test(key)) {
        return current
          .map((item) => resolve(item, keyIndex))
          .filter((val) => val !== undefined);
      }
      // 2. Otherwise (it's an object, or it's an array and we have a specific index),
      // we move deeper using the key.
      return resolve(current[key], keyIndex + 1);
    };
    return resolve(obj, 0);
  };
  ```

src/react/src/components/RoleScopeCard.tsx
1maple1: Agree, but let's revisit another time so the focus is on functionality and not styling/structure.

TheBurchLog (approved):

PR #2304, opened 2026-08-03
Mui topic index

TheBurchLog:


src/react/src/layouts/TopicIndex.tsx
```diff
@@ -388,315 +398,326 @@ function TopicIndex({
...
-          }}
-        />
-      );
-    };
+    const tableColumns = [
+      {
+        id: "topic",
+        label: "Topic",
+        field: "name",
+        isString: true,
```
1maple1, line 667: It would be super easy to add sortable and filterable to topic. If there are a lot of topics filtering by name would be a really nice feature to have.

src/react/src/layouts/TopicIndex.tsx
```diff
@@ -388,315 +398,326 @@ function TopicIndex({
...
-                className="ml-2"
-                aria-label={`Clear Count of ${topicSubscriber.subscriber.consumer_count} for Topic ${topicSubscriber?.topic?.name} Subscriber ${topicSubscriber.subscriber.garden ?? "*"} ${topicSubscriber.subscriber.namespace ?? "*"} ${topicSubscriber.subscriber.system ?? "*"} ${topicSubscriber.subscriber.version ?? "*"} ${topicSubscriber.subscriber.instance ?? "*"} ${topicSubscriber.subscriber.command ?? "*"}`}
-                tooltip="Clear count"
                 onClick={() =>
-                  clearCount(
-                    topicSubscriber.topic as Topic,
-                    topicSubscriber.subscriber as Subscriber,
-                  )
+                  addRequestItem({
+                    topic: { id: topic.id } as Topic,
```
1maple1: I don't think this looks up the topic using the id so the view topic card is missing the data it needs. Maybe give it `topic: { id: topic.id, name: topic.name, publisher_count: topic.publisher_count } as Topic,`. TopicCard still needs some work but that can be a follow-on.

1maple1 (commented):
Table looks great.

1maple1 (approved):

PR #2306, opened 2026-08-04
User Index Migration

TheBurchLog:


src/react/src/layouts/UserIndex.tsx
```diff
@@ -248,51 +246,35 @@ function UserIndex({
...
-      return (
-        <span {...GenerateTourProps(lastAuthenticatedUserTourStep)}>
-          {date.toUTCString()}
-        </span>
+        <Box {...GenerateTourProps(maxPermissionUserTourStep)}>OPERATOR</Box>
       );
     }
     return (
-      <span {...GenerateTourProps(lastAuthenticatedUserTourStep)}>Never</span>
+      <Box {...GenerateTourProps(maxPermissionUserTourStep)}>READ_ONLY</Box>
```
1maple1: span is equivalent to Box component="span". Not sure why this change was necessary. I thought we agreed to leave span and div unless we needed to convert to Box for MUI styling.

src/react/src/layouts/UserIndex.tsx
```diff
@@ -705,46 +660,102 @@ function UserIndex({
         />
       )}
 
-      <DataTable
+      <EnhancedTable
         data-testid="user-datatable"
```
1maple1, line 668: There is a failing test due to lack of ...props on EnhancedTable to pass forward the testid as mentioned previously.

TheBurchLog: Once Convert Roles is merged in the unit tests should pass

1maple1 (approved):

PR #2308, opened 2026-08-04
Migrate User Create

TheBurchLog:


1maple1 (commented):
Is it worth adding the button to dismiss the dialog in the upper right hand corner?

1maple1 (approved):

PR #2310, opened 2026-08-05
Migrate Topic Card

TheBurchLog:


src/react/src/components/TopicCard.tsx
```diff
@@ -208,160 +277,177 @@ function TopicCard({
...
-          >
-            <FontAwesomeIcon icon="xmark" />
-          </AccessButton>
-        )
-      }
-    >
-      <div>
-        <div className="mr-4">
+    <>
+      <DialogContent>
```
1maple1: Prefer the look with dividers but not a dealbreaker.

1maple1 (approved):

PR #2312, opened 2026-08-05
Cleanup navigation

1maple1:


src/react/src/components/Navigation.tsx
TheBurchLog: Could you also update the icon based buttons to have the color defined within the sx property instead of the color prop

```
sx={{ height: "36px", color: 'primary.contrastText' }}
```

TheBurchLog (approved):

PR #2320, opened 2026-08-07
Update Garden summary

1maple1:


src/react/src/components/GardenSummary.tsx
```diff
@@ -771,43 +786,59 @@ function GardenSummary({
...
-                    body={statusTemplate}
-                  />
-                  <Column
-                    header="Actions"
-                    body={(node: any) => connectionActions(node, "RECEIVING")}
-                  />
-                </DataTable>
+                <EnhancedTable
+                  data={receivingConnections}
+                  displayAll={false}
```
TheBurchLog: False is the default value, did you mean true?

src/react/src/components/GardenSummary.tsx
```diff
@@ -410,30 +417,38 @@ function GardenSummary({
...
-      <div className="flex ml-2 page-header">
-        <h1 className="flex-1">
+    <Box sx={{ mb: 4, width: "100%" }} key={selectedGarden?.name}>
+      <Box
+        sx={{
+          display: "flex",
+          ml: 1,
+          pb: "9px",
+          margin: "20px 0 20px",
+          borderBottom: "1px solid #eeeeee",
```
TheBurchLog: Instead of hard coding in a color, after the `<Box>` we could use the MUI `<Divider/>` to create the line

src/react/src/components/GardenSummary.tsx
```diff
@@ -771,43 +786,59 @@ function GardenSummary({
...
-                    field="status"
-                    header="Status"
-                    body={statusTemplate}
-                  />
-                  <Column
-                    header="Actions"
-                    body={(node: any) => connectionActions(node, "RECEIVING")}
-                  />
-                </DataTable>
+                <EnhancedTable
```
TheBurchLog, line 790: Not required, but we could pass the Receiving or Publishing element through the table header with a buffer around it instead of defining it externally.

TheBurchLog (approved):

PR #2321, opened 2026-08-07
Migrate Request Options

TheBurchLog:


src/react/src/components/RequestOptions.tsx
```diff
@@ -193,24 +189,142 @@ function RequestOptions({
...
+                <FAIcon icon="caret-down" />
+              </AccessButton>
+            </ButtonGroup>
+            <Popper
+              sx={{ zIndex: 1 }}
+              open={openSplitMenu}
+              anchorEl={splitButtonAnchorRef.current}
+              role={undefined}
+              transition
+              disablePortal
```
1maple1: For isCard, remove disablePortal and set the zIndex equal or higher (>=1300) than the modal so it doesn't open inside the parent container and create a scroll bar. This seems to be fine for the open request view, too.

src/react/src/components/RequestOptions.tsx
```diff
@@ -193,24 +189,142 @@ function RequestOptions({
...
+                  } else {
+                    pourAgain(request);
+                  }
+                }}
+              >
+                <Box sx={{ display: "flex" }}>
+                  {isCard ? (
+                    <FAIcon icon="up-right-from-square" className="mr-2" />
+                  ) : (
+                    <FAIcon icon="plus" className="mr-2" />
```
1maple1: FAIcon should be using the sx styling

1maple1 (approved):

PR #2325, opened 2026-08-07
Migrate Current Requests Template

TheBurchLog:


1maple1 (commented):
I haven't noticed lately but in this case it is noticeable that the counts do not span the length of the bottom row.

1maple1 (approved):

PR #2335, opened 2026-08-11
Fix Table Console Logs

TheBurchLog:


src/react/src/components/EnhancedTable/components/EnhancedTable.tsx
```diff
@@ -709,6 +711,7 @@ const EnhancedTable = ({
         </Table>
         {(displayAll === undefined || displayAll === false) && (
           <TablePagination
+            component="div"
```
1maple1, line 720: This actually fixes the weird divider issue on Current Requests. However the alignment is off and pushed to the right. I found this to push it back to the left: `sx={{
        '& .MuiTablePagination-spacer': {
          display: 'none', 
        },
        '& .MuiTablePagination-toolbar': {
          justifyContent: 'flex-start',
          paddingLeft: 2,
        },
      }}`

src/react/src/components/EnhancedTable/components/EnhancedTable.tsx
```diff
@@ -709,6 +711,7 @@ const EnhancedTable = ({
...
             rowsPerPageOptions={pageRecords ? [5, 10, 25] : []}
             colSpan={3}
```
1maple1: Colspan doesn't seem to do anything

src/react/src/components/EnhancedTable/components/EnhancedTable.tsx
1maple1: I think the footer should be moved down one line so it is outside the TableContainer. I only see it come into play with the close button for current requests.

1maple1 (approved):

PR #2344, opened 2026-08-12
Sort and filter scopes in roles

1maple1:


TheBurchLog (approved):
Exists logic needs to be updated for Arrays, but this PR works as expected

PR #2360, opened 2026-08-14
Request View 508

TheBurchLog:


src/react/src/components/RequestTreeMenu.tsx
```diff
@@ -168,10 +170,11 @@ function RequestTreeMenu({
                 icon="hourglass"
                 sx={{ mr: 2 }}
                 role="img"
+                title="Template Command"
                 aria-label="Template Command"
```
1maple1: Shouldn't this be Temp instead of Template?

1maple1 (approved):

PR #2367, opened 2026-08-18
Request wizard select version

1maple1:


src/react/src/components/VersionList.tsx
```diff
@@ -0,0 +1,66 @@
...
+        setSelectedVersion(chosenVersion);
+        if (chosenVersion) {
+          setSystems((prevSystems: System[]) =>
+            prevSystems.map((sys: System) =>
+              sys.name === system.name
+                ? { ...sys, version: chosenVersion }
+                : sys,
+            ),
+          );
+        }
```
TheBurchLog, line 60: In the UI, we can clear the input field and it appears to have nothing in it. So the `selectedVersion` is set to undefined. But it is technically still the last valid value selected. This could go one of two ways. Either change this to `<Select/>` so it forces a selection or disables the button selection for that system if a version is not selected.

TheBurchLog: Better answer, just add `disableClearable` to the `<Autocomplete` component. 

Example from docs:
```html
<Autocomplete
        {...defaultProps}
        id="disable-clearable"
        disableClearable
        renderInput={(params) => (
          <TextField {...params} label="disableClearable" variant="standard" />
        )}
      />
```

TheBurchLog (approved):

PR #2368, opened 2026-08-18
Mui theme

TheBurchLog:


1maple1 (commented):
Need to fix the wizard stepper highlighting if teal is our color. It is hard to tell what step is active.

1maple1 (approved):

PR #2378, opened 2026-08-19
Updating how selected command is displayed

TheBurchLog:


1maple1 (commented):
Needs to handle long names better. For instance, command error_string_output_type_json overflows.

1maple1 (approved):

PR #2383, opened 2026-08-20
Inline the systems for Garden Navigation

TheBurchLog:


1maple1 (approved):
Status counts are now inline with each other but could be inline with the garden name, too, but that can be another effort.

PR #2386, opened 2026-08-21
Migrate Parent to Parent_ID and update get request

TheBurchLog:
brewtils: update_request_model

### DIFF
--- a/src/app/beer_garden/db/mongo/migration.py
+++ b/src/app/beer_garden/db/mongo/migration.py
@@ -789,4 +789,5 @@ def ensure_model_migration():
             ensure_v3_29_model_migration()
             ensure_v3_30_model_migration()
 
+    # TODO: Move this over before release
     ensure_v3_35_model_migration()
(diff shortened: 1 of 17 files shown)


src/app/beer_garden/db/mongo/migration.py
```diff
@@ -235,6 +236,38 @@ def ensure_v3_29_model_migration():
...
+            if legacy_request:
+                if "parent" in legacy_request and isinstance(
+                    legacy_request["parent"], DBRef
+                ):
+                    request_updates.append(
+                        UpdateOne(
+                            {"_id": legacy_request["_id"]},
+                            {
+                                "$set": {"parent_id": legacy_request["parent"].id},
+                                "$unset": {"parent": ""},
```
1maple1: Are you sure the parent ref is no longer needed? I know the reverse delete rule is DO_NOTHING. Adding the parent_id is one thing but removing the ref is another.

1maple1 (approved):
Code looks fine. Tested as good as possible.

1maple1 (approved):

PR #2397, opened 2026-08-26
Environment variables for base url

TheBurchLog:


src/react/src/services/util_service.tsx
```diff
@@ -68,9 +68,48 @@ export const GetVersion = async (): Promise<Version> => {
 };
 
 export const GetBaseURL = (): string => {
```
1maple1, line 67: Not a fan of this approach especially the hard-coded routes. What would be nice if this could take the routes already defined in App.tsx. It looks like maybe one option is to have the routes in a separate file that could be imported.

1maple1 (approved):

PR #2399, opened 2026-08-31
Add Table Row Page Defaults

TheBurchLog:


1maple1 (commented):
If requests changes to 10-100 should we standardize other tables to have similar options? What is the motivation for keeping 5-25 and defaulting to 5 for the rest? Old UI default was 10. Topics is another one that could/should have a higher default.

1maple1 (approved):

PR #2402, opened 2026-09-01
Update table filter modal

1maple1:


src/react/src/components/EnhancedTable/components/EnhancedTableFilterOptions.tsx
```diff
@@ -435,7 +435,7 @@ export const EnhancedTableFilterOptions = ({
               onClick={() => removeFilter(id)}
               aria-label="Remove filter"
             >
-              <FontAwesomeIcon icon="x" />
+              <FAIcon icon="xmark" sx={{ color: "text.primary" }} />
```
TheBurchLog: We could instead do this globally for the icon button CSS in our theme like:
```
    MuiIconButton: {
      styleOverrides: {
        root: ({ theme }) => ({
          color: theme.palette.text.primary,
        }),
      },
    },
  },
```

This would handle any other hidden icon button we might have in the UI

src/react/src/components/EnhancedTable/components/EnhancedTableFilterSelect.tsx
```diff
@@ -77,13 +77,13 @@ export const EnhancedTableFilterSelect = ({
       <Grid container sx={{}}>
         <Grid size="grow" sx={{ ml: 2 }}>
           <Button variant="outlined" onClick={() => addFilter()}>
-            <FontAwesomeIcon icon="plus" />
+            <FAIcon icon="plus" sx={{ mr: 1 }} />
```
TheBurchLog: I'm split on this. I like the idea of having everything for enhanced table self contained. So you only need MUI and FontAwesome. This adds an external file to the component that is required. Thoughts on just wrapping the text in Typography with `ml:1`?

src/react/src/theme.ts
```diff
@@ -101,5 +101,12 @@ export const theme = createTheme({
...
       },
     },
+    MuiIconButton: {
+      styleOverrides: {
+        root: ({ theme }) => ({
+          color: theme.palette.text.primary,
+        }),
+      },
+    },
   },
```
TheBurchLog, line 136: Could you also add:
```
MuiButton: {
      styleOverrides: {
        root: ({ ownerState, theme }) => ({
          ...(ownerState.variant === 'outlined' && {
            borderColor: theme.palette.text.primary,
            color: theme.palette.text.primary,
          }),
        }),
      },
    },
```

TheBurchLog (approved):

PR #2403, opened 2026-09-01
Utilize new Request API to filter response data

TheBurchLog:


1maple1 (commented):
`Unable to find source System/Command` when clicking through the request tree. It displays correctly and then not. Nested-calls would be a good one to reproduce.

1maple1 (approved):

PR #2407, opened 2026-09-02
Add buffer around dialog for full screen scenario

TheBurchLog:


1maple1 (commented):
I'm not sure about this. The full screen on a dialog should never go past the top of the navigation bar and there has to be a way to calculate this.

1maple1 (approved):

PR #2411, opened 2026-09-02
Mui alias name overflow

TheBurchLog:


1maple1 (approved):
Much improved. Still not perfect but it will do.

PR #2414, opened 2026-09-02
Add quick links to view jobs from requests

TheBurchLog:


src/react/src/layouts/RequestIndex.tsx
```diff
@@ -115,6 +117,15 @@ function RequestIndex({
     pos: 5,
   };
 
+  const ViewJobTourStep: TourStepProps = {
+    prefix: tourPrefix,
+    uuid: tourUUID,
+    label: `View Job`,
+    content: `View assocaited Job in popup modal.`,
```
1maple1: associated

src/react/src/components/RequestOptions.tsx
```diff
@@ -170,6 +170,28 @@ function RequestOptions({
         setRequest(undefined);
       },
     });
+
+    if (
+      !isCard &&
```
1maple1: I take it you don't show this when a card because this also opens a dialog? But pour again works this way so this seems inconsistent.

src/react/src/layouts/RequestIndex.tsx
```diff
@@ -237,11 +261,22 @@ function RequestIndex({
...
+          sx={{ mr: 1 }}
           {...GenerateTourProps(ViewRequestTourStep)}
         >
           <FontAwesomeIcon icon="eye" />
         </AccessButton>
+        {showJob && (
+          <AccessButton
+            basic
+            onClick={() => PeekJobView(request)}
+            tooltip={`View Associate Job ${request?.metadata?.bg_job_id}`}
```
1maple1: Associated or simple use View Job

src/react/src/layouts/RequestIndex.tsx
```diff
@@ -237,11 +261,22 @@ function RequestIndex({
...
+            sx={{ mr: 1 }}
+            {...GenerateTourProps(ViewJobTourStep)}
```
1maple1, line 275: This step doesn't show in tour as expected. Last is still View Request.

1maple1 (approved):

PR #2416, opened 2026-09-03
Fix ports and prefix logic

TheBurchLog:


1maple1 (commented):
While you are at it, can you fix the indents? Copy icon adds indents that prevent direct paste into python console for example. JSON is the only one that doesn't add indents.

1maple1 (approved):

PR #2418, opened 2026-09-09
Convert over to using menu list

TheBurchLog:


src/react/src/components/TreeMenu.tsx
```diff
@@ -133,17 +84,31 @@ function TreeMenu<T extends ExtendedTreeItemProps, M extends boolean = false>({
...
-          {...richTreeProps}
-        />
+        <MenuList {...menuProps}>
+          {options?.map((option: TreeMenuItemProps) => (
+            <MenuItem
+              key={option.id}
+              selected={selected === option.id}
+              onClick={() => onSelected(option)}
+              sx={{
+                ml: option.depth ? option.depth * 2 : 0,
```
1maple1: If you used padding instead of margin then the whole list item would be selectable. For example `pl: option.depth ? option.depth * 2 + 2 : 2`. What do you think?

src/react/src/components/TreeMenu.tsx
```diff
@@ -133,17 +84,31 @@ function TreeMenu<T extends ExtendedTreeItemProps, M extends boolean = false>({
...
       )}
       {(isLoading === undefined || isLoading === false) && (
-        <RichTreeView
-          slots={{ item: CustomTreeItem }}
-          onItemSelectionToggle={handleItemSelectionToggle}
-          expandedItems={expandedItems}
-          onExpandedItemsChange={handleExpandedItemsChange}
-          {...richTreeProps}
-        />
+        <MenuList {...menuProps}>
```
1maple1, line 91: Need to handle overflowX. Should be hidden and add scrollbar

1maple1 (commented):
Functionality is pretty good. Don't like how the hover and selected color are nearly the same. Maybe make the text bold when selected. A nice to have would be to have the margin get the background color on hover.

src/react/src/components/TreeMenu.tsx
```diff
@@ -133,17 +90,37 @@ function TreeMenu<T extends ExtendedTreeItemProps, M extends boolean = false>({
...
+          {options?.map((option: TreeMenuItemProps) => (
+            <MenuItem
+              key={option.id}
+              selected={selected === option.id}
+              onClick={() => onSelected(option)}
+              sx={{
+                width: 'max-content',
```
1maple1: This puts a tight border on the MenuItem content. I prefer `minWidth` so it will put the border around the full width of the item and will encapsulate all content when it overflows. I think this is as good as we can get on this.

1maple1 (approved):

PR #2425, opened 2026-09-10
Change confirm dialog to use cancel

1maple1:


src/react/src/components/ConfirmDialog.tsx
TheBurchLog: Should we also update the function input `reject` to `cancel`?

1maple1: done

TheBurchLog (commented):
Need to update any references that use `<ConfirmDialog/>` like RequestOptions.tsx where it still passes reject as a parameter.

TheBurchLog (approved):

PR #2439, opened 2026-09-16
Update dashboard tree styling

1maple1:


src/react/src/layouts/Dashboard.tsx
```diff
@@ -604,7 +604,7 @@ function GardenDashboard({
               border: "1px solid",
               borderColor: grey[300],
               borderRadius: 2,
-              p: 2,
+              overflowX: "scroll"
```
TheBurchLog: So it doesn't show until it is needed
```
overflowX: "auto"
```

TheBurchLog (approved):

PR #2441, opened 2026-09-17
Make it easier for third party services to modify loaded html

TheBurchLog:


src/react/src/App.tsx
```diff
@@ -177,6 +177,10 @@ function App() {
         }
         loadRootGarden(config);
         setInvalidUrl(false);
+
+        // Allows legacy deployment services to trigger events
+        // once the configuration is loaded
+        window.dispatchEvent(new Event("load"));
```
1maple1, line 183: Assuming there is a "load" event listener per the discussion yesterday.

1maple1 (approved):

